PR #27, opened 2020-09-23
fix: add restart count metric to terminated pods

ardias:
in case a pod is in a crash loop we were not getting the restart count


paologallinaharbur (approved):
LGTM

PR #34, opened 2020-10-21
fix: containerId coming from system driver are correctly parsed

### LINKED ISSUE #33: Wrong containerId with systemD managed containers
With systemD docker container ids have a different format than docker "generic" id's and so the integration reports container identifiers that don't match with other components.

Example: 
with systemd: /docker/ae17ce6dcd2f27905cedf80609044290eccd98115b4e1ded08fcf6852cf939ae/kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-pod13118b761000f8fe2c4662d5f32d9532.slice/crio-ebccdd64bb3ef5dfa9d9b167cb5e30f9b696c2694fb7e0783af5575c28be3d1b.scope

without systemd: /docker/d44b560aba016229fd4f87a33bf81e8eaf6c81932a0623530456e8f80f9675ad/kubepods/besteffort/pod6edbcc6c66e4b5af53005f91bf0bc1fd/7588a02459ef3166ba043c5a605c9ce65e4dd250d7ee40428a28d806c4116e97
7588a02459ef3166ba043c5a605c9ce65e4dd250d7ee40428a28d806c4116e97

## Expected Behavior

Container id with systemd driver reported in K8s events should equal id's coming from docker without systemd

We need to modify the container id parsing code to account for this difference

ardias:
fixes #33


src/kubelet/metric/cadvisor.go
```diff
@@ -129,6 +136,21 @@ func createRawEntityID(m prometheus.Metric) (string, error) {
 }
 
 // /kubepods/besteffort/podba8b34d7-11a3-11e8-a084-080027352a02/a949bd136c1397b9f52905538ee11450427be33648abe38db06be2e5cfbeca49
+// /docker/ae17ce6dcd2f27905cedf80609044290eccd98115b4e1ded08fcf6852cf939ae/kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-pod13118b761000f8fe2c4662d5f32d9532.slice/crio-ebccdd64bb3ef5dfa9d9b167cb5e30f9b696c2694fb7e0783af5575c28be3d1b.scope
+// /docker/d44b560aba016229fd4f87a33bf81e8eaf6c81932a0623530456e8f80f9675ad/kubepods/besteffort/pod6edbcc6c66e4b5af53005f91bf0bc1fd/7588a02459ef3166ba043c5a605c9ce65e4dd250d7ee40428a28d806c4116e97
 func extractContainerID(v string) string {
-	return v[strings.LastIndex(v, "/")+1:]
+	containerID := v[strings.LastIndex(v, "/")+1:]
```
paologallinaharbur, line 142: can we add some comments more explaining the different cases?
In the future it can be difficult to remember :)

gsanchezgavier (dismissed):

gsanchezgavier:
could you check the linter errors?

gsanchezgavier (approved):
Looks good , thanks for the tests

PR #37, opened 2020-10-28
Add requested CPU and memory for NodeSample

adriaferre:


src/kubelet/group.go
```diff
@@ -66,12 +66,26 @@ func (r *kubelet) Group(definition.SpecGroups) (definition.RawGroups, *data.Erro
 			Errors:      []error{fmt.Errorf("error querying ApiServer: %v", err)},
 		}
 	}
+
+	podsResource, err := metric.GetMetricsPodsData(r.client)
+	var nodeRequestedCPU, nodeRequestedMemory int64 = 0, 0
+	if err != nil {
```
ahoulgrave, line 72: maybe allow to fail, but do log the error so we can debug

ahoulgrave: if we allow the request to fail, we may collect misleading data

src/kubelet/metric/metric.go
```diff
@@ -43,6 +48,28 @@ func GetMetricsData(c client.HTTPClient) (v1.Summary, error) {
 
 }
 
+// GetMetricsPodsData calls kubelet /pods endpoint and returns unmarshalled response
```
alejandrodnm, line 51: The integration at some point already queries the pods endpoint of the kubelet using this https://github.com/newrelic/nri-kubernetes/blob/main/src/kubelet/metric/pods.go#L42 I think what you should do, is hook your code at some point in the integration execution, where the integration already has the data you want.

We want to limit the amount of requests we make from the integration, to reduce as much as possible the footprint of the agent on the Kubelet and apiserver.

I think at this point https://github.com/newrelic/nri-kubernetes/blob/main/src/kubelet/group.go#L69 the `rawGroups` object already has the data you want, you'll just have to traverse the object, aggregate the data you want and add the data you want the same way the labels and allocatable resources are being added.

The team that maintains this project is super busy right now, but I can lend a hand and review your PR, ping me whenever you want.

alejandrodnm (changes requested):

src/kubelet/metric/metric.go
ahoulgrave: At this point: https://github.com/newrelic/nri-kubernetes/blob/main/src/kubelet/group.go#L69

The information i have on pods is:
```
pod: map[
    createdAt:2020-12-23 09:46:15 +0000 UTC 
    createdBy:nginx-deployment-695987c984 
    createdKind:ReplicaSet 
    deploymentName:nginx-deployment 
    isReady:True 
    isScheduled:True 
    labels:map[
        app:nginx pod-template-hash:695987c984
    ]
    namespace:default 
    nodeIP:192.168.49.2 
    nodeName:minikube 
    podName:nginx-deployment-695987c984-t229l 
    startTime:2020-12-23 09:46:15 +0000 UTC 
    status:Running
]
```

where

```
-> % k get pod nginx-deployment-695987c984-t229l -o jsonpath="{.spec.containers[0].resources}" |jq
{
  "limits": {
    "cpu": "500m",
    "memory": "128Mi"
  },
  "requests": {
    "cpu": "250m",
    "memory": "64Mi"
  }
}
```

we'll need to find a place where this data is available

ahoulgrave:
Opened a new PR : https://github.com/newrelic/nri-kubernetes/pull/61

src/kubelet/metric/metric.go
ahoulgrave: NVM: it's in the containers, not the pods :) 

```
ncontainer: map[
    containerID:33a266b2b09f33ac22aaafe5b3bedf0cbee14ee24b3cb73c4698ddb3448afee2
    containerImage:nginx:1.14.2
    containerImageID:nginx@sha256:f7988fb6c02e0ce69257d9bd9cf37ae20a60f1df7563c3a2a6abe24160306b8d
    containerName:nginx 
    container_cpu_cfs_periods_total:3 
    container_cpu_cfs_throttled_periods_total:0 
    container_cpu_cfs_throttled_seconds_total:0 
    container_memory_mapped_file:0 
    cpuLimitCores:500 
    cpuRequestedCores:250 
    deploymentName:nginx-deployment 
    isReady:true 
    labels:map[
        app:nginx 
        pod-template-hash:695987c984
    ] 
    memoryLimitBytes:134217728 
    memoryRequestedBytes:67108864 
    namespace:default 
    nodeIP:192.168.49.2 
    nodeName:minikube 
    podName:nginx-deployment-695987c984-t229l 
    restartCount:0 
    startedAt:2020-12-23 09:46:28 +0000 UTC 
    status:Running
]
```

PR #54, opened 2020-12-01
ci: check deployment ymls point to the released version

### LINKED ISSUE #53: Check the correct version is pointed in deployment ymls before releasing
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description
[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT share sensitive information, whether personal, proprietary, or otherwise! )

Release pipeline (#50) does not check that the version being released is the one used in deployment yamls. This can cause inconsistencies with the versioning in the S3 bucket.

## Expected Behavior
[NOTE]: # ( Tell us what you expected to happen. )

A deployment named `newrelic-infrastructure-k8s-1.2.3.yml` is expected to refer to deploy version `1.2.3`.

roobre:
Fixes #53

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,6 +26,12 @@ jobs:
 
       - uses: actions/checkout@v2
 
+      - name: Check consistency of deployment yamls
+        if: ${{ ! github.event.release.prerelease }} # On pre-releases, the -pre tag is added making this more complex
+        run: |
+          grep -e "$DOCKER_IMAGE_NAME:$DOCKER_IMAGE_TAG" deploy/newrelic-infra.yaml
+          grep -e "$DOCKER_IMAGE_NAME:$DOCKER_IMAGE_TAG" deploy/newrelic-infra-unprivileged.yaml
+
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -26,6 +26,12 @@ jobs:
 
       - uses: actions/checkout@v2
 
+      - name: Check consistency of deployment yamls
+        if: ${{ ! github.event.release.prerelease }} # On pre-releases, the -pre tag is added making this more complex
```
paologallinaharbur, line 30: why we are not checking this on prerelease?

paologallinaharbur: isn't better to catch the essero as soon as possible?

roobre: I tweaked it to strip the `-pre` suffix in the version check. It is not ideal, but should work without the need to change the yaml between prereleases.

paologallinaharbur (approved):

PR #57, opened 2020-12-09
feat(baseImage): updating base image

paologallinaharbur:
Bumping base image to update all the bundled integrations.

More info regarding all the integration upgraded can be found in the [release notes of the base image](https://github.com/newrelic/infrastructure-bundle/releases/tag/2.0.0).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+## 2.0.0
+
+### Changed
+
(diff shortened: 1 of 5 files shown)


roobre (changes requested):
It think we're missing https://github.com/newrelic/nri-kubernetes/blob/218825f61f05cbf32eb98afee9ebac24a4bcacf3/deploy/newrelic-infra.yaml#L63

paologallinaharbur:
Nice catch, I had the file modified locally but it was not saved

roobre (approved):
lgtm

Issue #56, opened 2020-12-03
Add missing metrics


jlegoff:
A number of metrics are being computed on the fly by the backend, as explained in this document: https://docs.google.com/document/d/1C45zYn7StoJfbjqxe70rRB-bZREbg8bqwIJRooXzkX8

These computed metrics should be added by the integration.

arvdias:
A bit more context for this Issue:

- Some metrics for this is integration are being calculated in the backend from the data we're sending. 
- These metrics are being used in "special" alerts we provide for customers. 
- While calculating them in the backend works, we should add these metrics to the integration as it makes them easier to manage in the backend and reduces the workaround currently in place.

Event type: 
- K8sReplicasetSample, 
- K8sStatefulSample, 
- K8sDeploymentSample, 
- K8sDaemonsetSample
Calculation: **podsMissing** = podsDesired - podsReady

Event type: K8sContainerSample
Calculation: **cpuCoresUtilization** = (cpuUsedCores / cpuLimitCores) * 100

Event type: K8sContainerSample
Calculation: **requestedCpuCoresUtilization** = (cpuUsedCores / cpuRequestedCores) * 100

Event type: K8sContainerSample
Calculation: **memoryUtilization** = (memoryUsedBytes / memoryLimitBytes) * 100

Event type: K8sContainerSample
Calculation: **requestedMemoryUtilization** = (memoryUsedBytes / memoryRequestedBytes) * 100

Event type: K8sEtcdSample
Calculation: **processFdsUtilization** = (processOpenFds / processMaxFds) * 100

Event type: K8sNodeSample
Calculation: **allocatableCpuCoresUtilization** = (cpuUsedCores / allocatableCpuCores) * 100

Event type: K8sNodeSample
Calculation: **allocatableMemoryUtilization** = (memoryWorkingSetBytes / allocatableMemoryBytes) * 100

Event type: K8sNodeSample
Calculation: **fsCapacityUtilization** = (fsUsedBytes / fsCapacityBytes) * 100

arvdias:
Another set of metrics we will add as part of this issue are Horizontal Pod Autoscaler Metrics: https://github.com/kubernetes/kube-state-metrics/blob/master/docs/horizontalpodautoscaler-metrics.md

PR #67, opened 2021-02-04
arvdias/add new metrics

arvdias:


src/metric/populate_test.go
```diff
@@ -147,13 +147,18 @@ func TestPopulateK8s(t *testing.T) {
...
+	// TODO not good to compare error strings...
 	expectedErrs := []error{
 		errors.New("error populating metric for entity ID kube-system_newrelic-infra-rz225: cannot fetch value for metric deploymentName, metric not found"),
 		errors.New("error populating metric for entity ID kube-system_newrelic-infra-rz225: cannot fetch value for metric reason, metric not found"),
 		errors.New("error populating metric for entity ID kube-system_newrelic-infra-rz225: cannot fetch value for metric message, metric not found"),
 		errors.New("error populating metric for entity ID kube-system_newrelic-infra-rz225_newrelic-infra: cannot fetch value for metric deploymentName, metric not found"),
 		errors.New("error populating metric for entity ID kube-system_newrelic-infra-rz225_newrelic-infra: cannot fetch value for metric cpuLimitCores, metric not found"),
 		errors.New("error populating metric for entity ID kube-system_newrelic-infra-rz225_newrelic-infra: cannot fetch value for metric reason, metric not found"),
+		errors.New("error populating metric for entity ID kube-system_newrelic-infra-rz225_newrelic-infra: cannot fetch value for metric cpuCoresUtilization, 'cpuUsedCores' is nil"),
```
paologallinaharbur, line 158: Should we at least create a constant for these expected values?

paologallinaharbur: I guess it is not ideal since they are kind of generated 😕

arvdias: yes, checking strings is not the best option. We should create some kind of error type for these. I just didn't have time in this pr

src/metric/definition.go
```diff
@@ -974,3 +1068,31 @@ func toCores(value definition.FetchedValue) (definition.FetchedValue, error) {
...
+		return float64(v), nil
+	case prometheus.CounterValue:
+		return float64(v), nil
+	}
+
+	return nil, fmt.Errorf("invalid type value '%v'. Expected 'gauge' or 'counter', got '%T'", value, value)
+}
+
+// Subtract returns a new FetchFunc that subtracts 2 values. It expects that the values are float64
+func Subtract(left definition.FetchFunc, right definition.FetchFunc) definition.FetchFunc {
```
paologallinaharbur, line 1084: Now we understand when you said finding where to add these function was not straightforward 😅

src/metric/definition.go
```diff
@@ -405,6 +405,12 @@ var EtcdSpecs = definition.SpecGroups{
 				ValueFunc: prometheus.FromValueWithOverriddenName("go_goroutines", "goGoroutines"),
 				Type:      sdkMetric.GAUGE,
 			},
+			// computed
```
paologallinaharbur, line 408: is it any difference between `computed` and `calculated` metrics?

src/metric/definition.go
arvdias: yeah, it's all a bit complex

src/metric/definition.go
arvdias: no, it's probably I just that I did it at different times

src/metric/populate_test.go
arvdias: Let's do a release a first and then  I'll do another PR to change this. It's not a critical issue at the moment

src/metric/definition.go
```diff
@@ -495,6 +506,11 @@ var KSMSpecs = definition.SpecGroups{
 			{Name: "statefulsetName", ValueFunc: prometheus.FromLabelValue("kube_statefulset_created", "statefulset"), Type: sdkMetric.ATTRIBUTE},
 			{Name: "namespaceName", ValueFunc: prometheus.FromLabelValue("kube_statefulset_created", "namespace"), Type: sdkMetric.ATTRIBUTE},
 			{Name: "label.*", ValueFunc: prometheus.InheritAllLabelsFrom("statefulset", "kube_statefulset_labels"), Type: sdkMetric.ATTRIBUTE},
+			//computed
+			{Name: "podsMissing", ValueFunc: Subtract(
```
paologallinaharbur, line 510: Is it normal that the name is always podsMissing?

paologallinaharbur: Is it the same attribute saved in different sample?

paologallinaharbur: Yes, I believe it is that, than it makes sense

paologallinaharbur (approved):

arvdias:
closes #56

PR #70, opened 2021-02-16
Multiarch support

roobre:
This PR adds multiarch support for nri-kubernetes.

- Build pipeline now runs for several GOOS/GOARCH pairs
- `docker buildx` is used to produce multiarch images


.github/workflows/release.yml
```diff
@@ -5,12 +5,34 @@ on:
...
+        with:
+          name: nri-kubernetes-${{ matrix.goos }}-${{ matrix.goarch }}
+          path: bin/nri-kubernetes-${{ matrix.goos }}-${{ matrix.goarch }}
+
   release:
     name: Release docker images and config
     runs-on: ubuntu-latest
     env:
-      DOCKER_IMAGE_NAME: newrelic/infrastructure-k8s
+      DOCKER_IMAGE_NAME: roobre/nri-kubernetes # TODO: Change this back
```
paologallinaharbur, line 33: is this intentional?

roobre: Yes, just to avoid pushing `-pre` images to prod. It will be reverted once we test it working.

.github/workflows/release.yml
```diff
@@ -5,12 +5,35 @@ on:
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build integration
+        env:
+          GOOS: ${{ matrix.goos }}
+          GOARCH: ${{ matrix.goarch }}
+        run: |
+          make compile
+      - name: Upload artifact for docker build step
+        uses: actions/upload-artifact@v2
```
arvdias, line 24: what does this action do? where does it upload to?

roobre: To a very comfy place where the artifacts can live a peaceful and fulfilling life :)
https://docs.github.com/en/actions/guides/storing-workflow-data-as-artifacts

paologallinaharbur (approved):
Just keep in mind to change everything back

roobre:
rebased main

PR #73, opened 2021-02-23
Retrieve load balancer IP from kube_service_status_load_balancer_ingress

### LINKED ISSUE #74: Load Balancer IP missing from K8sServiceSample
## Description

Our [docs](https://docs.newrelic.com/docs/integrations/kubernetes-integration/understand-use-data/find-use-your-kubernetes-data#service-data) show that we collect the `loadBalancerIP` for Services, however, I do not see this data present in the `K8sServiceSample`.

image

## Expected Behavior

I'm expecting to see a Load Balancer IP for my NGINX Ingress controller service.

## Steps to Reproduce
N/A

## Your Environment

EKS v1.17.12-eks-7684af
New Relic Infra: `newrelic/infrastructure-k8s:2.2.0`

roobre:
Previously, the `LoadBalancerIP` field was retrieved from `kube_service_info.load_balancer_ip`. This value, however, seems to be blank for the current ksm version.

This PR changes it to be retrieved from `kube_service_status_load_balancer_ingress.ip`, which seems to be working with the current version.

Temporary images images for testing are available:
- `roobre/infrastructure-k8s:load-balancer`
- `roobre/infrastructure-k8s:load-balancer-unprivileged`

### Tasks
- [ ] Investigate whether this is still needed/recommended in newer kube-state-metrics versions
- [ ] Add a test case for this

---

Fixes #74

### DIFF
--- a/src/metric/definition.go
+++ b/src/metric/definition.go
@@ -776,6 +776,7 @@ var KSMQueries = []prometheus.Query{
 	{MetricName: "kube_service_created"},
 	{MetricName: "kube_service_labels"},
 	{MetricName: "kube_service_info"},
+	{MetricName: "kube_service_status_load_balancer_ingress"},
 	{MetricName: "kube_service_spec_type", Value: prometheus.QueryValue{
 		Value: prometheus.GaugeValue(1),
 	}},
--- a/src/metric/definition.go
+++ b/src/metric/definition.go
@@ -591,7 +591,7 @@ var KSMSpecs = definition.SpecGroups{
 			},
 			{
 				Name:      "loadBalancerIP",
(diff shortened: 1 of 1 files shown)


roobre:
Made a few tests, and I can confirm the behavior of `kube-state-metrics` is the same on 1.9.7 and 1.9.8 (latest at the time of writing).

arvdias (approved):

roobre:
As reported upstream, `kube_service_info.load_balancer_ip` only reflects an IP if it is defined in the service spec directly. `kube_service_status_load_balancer_ingress.ip` should be used when this IP is not known at deployment-time.

PR #79, opened 2021-02-25
fix(permissions): dropping some permissions and deleted old manifest

paologallinaharbur:
Following the change in the helm chart we need to update as well the static manifests


roobre (approved):
lgtm

PR #80, opened 2021-03-10
Fix docs for static pod status featureflag

alejandrodnm:


roobre:
Thanks for catching this!

PR #91, opened 2021-04-19
ksm/discovery: try several well-known port names

roobre:
Previously, when performing DNS discovery, we were looking for a port named `http-metrics`.

It would seem that this port is no longer used by `ksm` by default, and now just `http` is used. This PR adds the ability to search for both.

### DIFF
--- a/src/ksm/client/discovery.go
+++ b/src/ksm/client/discovery.go
@@ -24,13 +24,15 @@ var errNoKSMPodsFound = errors.New("no KSM pods found")
 
 const (
 	ksmAppLabelValue         = "kube-state-metrics"
-	ksmPortName              = "http-metrics"
 	k8sTCP                   = "TCP"
 	ksmQualifiedName         = "kube-state-metrics.kube-system.svc.cluster.local"
 	ksmDNSService            = "http-metrics"
 	ksmDNSProto              = "tcp"
 	headlessServiceClusterIP = "None"
 )
+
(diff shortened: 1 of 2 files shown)


roobre:
I believe this mechanism could be improved to also retrieve the namespace where the integration is deployed and the helm release name (from an annotation), and attempt to resolve that as well. That would add to the fast-path of DNS discovery the case where ksm was installed through `nri-bundle`.

src/ksm/client/discovery.go
```diff
@@ -160,7 +172,7 @@ func (sd *discoverer) apiDiscover() (url.URL, error) {
 		if service.Spec.ClusterIP != "" && len(service.Spec.Ports) > 0 {
 			// Look for a port called "http-metrics"
 			for _, port := range service.Spec.Ports {
-				if port.Name == ksmPortName {
+				if port.Name == ksmPortNames[0] {
```
gsanchezgavier, line 175: shouldn't also look for the others posible names in `ksmPortNames` ?

roobre: I think that would be ideal, but since we also have:

```
// If not found, return the first TCP port
```

I thought it was not needed to complicate the logic with a for there. Can still be added though.

gsanchezgavier (approved):
lgtm

roobre:
Closing for now since soon a new discovery mechanism will be implemented soon.

PR #110, opened 2021-05-26
helm3 support in e2e

paologallinaharbur:


paologallinaharbur:
https://github.com/newrelic/nri-kubernetes/issues/102

e2e/README.md
```diff
@@ -1,34 +1,20 @@
...
+eval $(minikube -p minikube docker-env)
 ```
-```
-helm init --service-account tiller
 
-# or if you installed Helm2 using Brew
-/usr/local/opt/helm@2/bin/helm init --service-account tiller
+Then you need to build the binary and the image. Notice that the Dockerfile is currently multiarch.
+Depending on your environment the easiest thing to do can be to manually modify the Dockerfile adding
+manually the values for TARGETARCH and TARGETOS
```
roobre, line 10:
```suggestion
Then you need to build the binary and the image. Notice that  since the Dockerfile includes multiarch
support, you may need to set `DOCKER_BUILDKIT=1` when running `docker build` for the `TARGETARCH`
and `TARGETOS` args to be populated.
```

roobre: (I guess you tried to warn against those not being populated? Setting `DOCKER_BUILDKIT=1` should fix that)

e2e/cmd/e2e.go
```diff
@@ -573,18 +531,25 @@ func installRelease(_ context.Context, s scenario.Scenario, logger *logrus.Logge
 		fmt.Sprintf("daemonset.clusterVersion=%s.%s.x", versionSplitted[0], versionSplitted[1]),
 	)
 
-	o, err := helm.InstallRelease(filepath.Join(dir, cliArgs.NrChartPath), cliArgs.Context, logger, options...)
+	releaseName := randString(5)
```
roobre, line 534: You can probably use https://pkg.go.dev/k8s.io/apimachinery/pkg/util/rand#String

e2e/cmd/e2e.go
```diff
@@ -573,18 +531,25 @@ func installRelease(_ context.Context, s scenario.Scenario, logger *logrus.Logge
...
 
-	return string(releaseName), nil
+var letterRunes = []rune("abcdefghijklmnopqrstuvwxyz")
+
+func randString(n int) string {
+	b := make([]rune, n)
+	for i := range b {
+		b[i] = letterRunes[rand.Intn(len(letterRunes))]
+	}
+	return string(b)
```
roobre, line 554: Would not be necessary if we use Kubernetes' rand, see above.

e2e/helm/helm.go
```diff
@@ -38,10 +35,9 @@ func InstallRelease(path, context string, logger *logrus.Logger, config ...strin
 	c := exec.Command(_helmBinary, args...)
 	o, err := c.CombinedOutput()
 	if err != nil {
-		return nil, fmt.Errorf("%s - %s", err, string(o))
+		return fmt.Errorf("%s - %s", err, string(o))
```
roobre, line 38:
```suggestion
		return fmt.Errorf("%w - %s", err, string(o))
```

paologallinaharbur: I updated also other occurrences in the code

roobre (approved):
Looking great!

PR #107, opened 2021-05-21
Update Kubernetes Go dependencies to latest versions

invidian:
Part of #90.

Signed-off-by: Mateusz Gozdek <mateusz@kinvolk.io>

PR #111, opened 2021-05-27
Update modules

paologallinaharbur:
- Update dependencies
 - Update to /v2

[done] To be rebase and merged after https://github.com/newrelic/nri-kubernetes/pull/110

Take over https://github.com/newrelic/nri-kubernetes/pull/107


src/kubelet/metric/resource_test.go
```diff
@@ -4,11 +4,19 @@ import (
...
 	"k8s.io/apimachinery/pkg/api/resource"
+
+	"github.com/newrelic/nri-kubernetes/v2/src/definition"
+)
+
+const (
+	// Copied from old version of k8s.io/api/core/v1.
+	//
+	// NVIDIA GPU, in devices. Alpha, might change: although fractional and allowing values >1, only one whole device per node is assigned.
+	resourceNvidiaGPU v1.ResourceName = "alpha.kubernetes.io/nvidia-gpu"
```
roobre, line 19:
```suggestion
	resourceNvidiaGPUName v1.ResourceName = "alpha.kubernetes.io/nvidia-gpu"
```

roobre: I'd suggest to add a `// TODO` here since we should revisit this later, to consider whether we should deprecate it or not.

roobre (approved):
Looking great, just a the minor comment on the nvidia resource

assets/licence/rules.json
```diff
@@ -17,6 +17,8 @@
       "Ruby",
       "ISC",
       "CC0-1.0",
-      "Zlib"
+      "Zlib",
+      "BSD-3-Clause License",
+      "MPL-2.0 License"
     ]
   }
```
roobre, line 24: EOF

src/kubelet/metric/resource_test.go
```diff
@@ -4,11 +4,17 @@ import (
...
 	"github.com/stretchr/testify/require"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+
+	"github.com/newrelic/nri-kubernetes/v2/src/definition"
+)
+
+const (
+	resourceTestBeta v1.ResourceName = "beta.newrelic.com/test-name"
+	resourceTestAlfa v1.ResourceName = "alfa.newrelic.com/test-name"
```
roobre, line 17:
```suggestion
	resourceTestAlpha v1.ResourceName = "alpha.newrelic.com/test-name"
```

src/kubelet/metric/resource_test.go
paologallinaharbur: I get rid of that part, adding the deprecation note was quite confusing giving that we were not using that string at all

PR #132, opened 2021-06-25
Bump integration version

alvarocabanas:
- Bump integration version so it appears updated in the samples.


paologallinaharbur (approved):
thanks for taking care of it!

PR #134, opened 2021-06-29
Add nightly builds

### LINKED ISSUE #125: k8s-stg/night: Automate nightly builds for nri-kubernetes
We should automate the generation of nightly builds for nri-kubernetes.
These builds should include:
* The latest pre-release of nri-kubernetes
* The latest pre-release of all of the integrations that are present in the bundle

In order to achieve this, we can follow two approaches:
1. Implement nightly builds on nri-kubernetes repo, using `newrelic/infrastructure-bundle-nightly` as a base
2. Deprecate `infrastructure-k8s` and bundle nri-kubernetes in the infrastructure bundle, thus reusing the logic already present there

invidian:
This PR adds new GitHub Actions workflow, which will build and push
integration Docker images daily, so one can use those images in bleeding
edge environments to spot any potential issues.

Based on https://github.com/newrelic/infrastructure-bundle/pull/67

Closes #125

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,100 @@
+name: Nightly build
+
+on:
+  schedule:
+    - cron: "0 3 * * *"
+  push:
+    branches:
+      - master
+      - main
+
+jobs:
+  build:
+    name: Build integration for
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        goos: [ linux ]
+        goarch: [ amd64, arm64, arm ]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.16'
+      - name: Build integration
+        env:
+          GOOS: ${{ matrix.goos }}
+          GOARCH: ${{ matrix.goarch }}
+        run: |
+          make compile
+      - name: Upload artifact for docker build step
+        uses: actions/upload-artifact@v2
+        with:
+          retention-days: 1
+          name: nri-kubernetes-${{ matrix.goos }}-${{ matrix.goarch }}
+          path: bin/nri-kubernetes-${{ matrix.goos }}-${{ matrix.goarch }}
+
+  docker:
+    name: Build and push images
+    needs: [ build ]
+    runs-on: ubuntu-latest
+    env:
+      DOCKER_IMAGE_NAME: newrelic/infrastructure-k8s
+      DOCKER_IMAGE_TAG: nightly
+      DOCKER_PLATFORMS: "linux/amd64,linux/arm64,linux/arm" # Must be consistent with the matrix from the job above
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Download all artifacts from build job
+        uses: actions/download-artifact@v2
+        with:
+          path: bin
+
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,98 @@
...
+          GOOS: ${{ matrix.goos }}
+          GOARCH: ${{ matrix.goarch }}
+        run: |
+          make compile
+      - name: Upload artifact for docker build step
+        uses: actions/upload-artifact@v2
+        with:
+          retention-days: 1
+          name: nri-kubernetes-${{ matrix.goos }}-${{ matrix.goarch }}
+          path: bin/nri-kubernetes-${{ matrix.goos }}-${{ matrix.goarch }}
```
invidian, line 35: Won't this collide with release workflow?

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,98 @@
...
+
+  docker:
+    name: Build and push images
+    needs: [ build ]
+    runs-on: ubuntu-latest
+    env:
+      DOCKER_IMAGE_NAME: newrelic/infrastructure-k8s
+      DOCKER_IMAGE_TAG: latest-nightly
```
invidian, line 43: Perhaps new repository should be created?

invidian (commented):
I did not test the pipeline yet, as it depends on #67 on which I either don't have access or it does not exist yet. I also have little idea how this can actually be tested.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,100 @@
+name: Nightly build
+
+on:
+  schedule:
+    - cron: "0 3 * * *"
+  push:
+    branches:
+      - master
```
paologallinaharbur, line 8: does it make sense to keep master?

.github/workflows/nightly.yml
paologallinaharbur: @roobre

.github/workflows/nightly.yml
invidian: Not sure, probably not. Copy-pasta from https://github.com/newrelic/infrastructure-bundle/pull/67.

.github/workflows/nightly.yml
paologallinaharbur: If I am not wrong each artifact is uploaded to the worflow itself, therefore they should not collide

.github/workflows/nightly.yml
paologallinaharbur: personally I would go for a no on this

.github/workflows/nightly.yml
paologallinaharbur: I checked and even if it is not useful we are following with the same approach in all the repos

paologallinaharbur (approved):

.github/workflows/nightly.yml
invidian: Note, that this repository already has `latest` tag, similar to `infrastructure-bundle` and `infrastructure-bundle` images will be pushed to `infrastructure-bundle-nightly` repository, hence the question.

.github/workflows/nightly.yml
invidian: I'll mark it as resolved then.

.github/workflows/nightly.yml
invidian: `newrelic/infrastructure-k8s-nightly`

invidian: We decided on call to push it to the same repository with `nightly` tag.

paologallinaharbur (approved):
everything looks good!

PR #148, opened 2021-07-09
cmd/kubernetes-static: fixes and improvements

invidian:


src/definition/fetch_test.go
```diff
@@ -44,15 +43,15 @@ func TestFromRawErrorsOnNotFound(t *testing.T) {
...
 
 	v, err = FromRaw("metric_name_3")("group1", "nonExistingEntity", raw)
-	assert.EqualError(t, err, "entity not found")
+	assert.EqualError(t, err, "entity \"nonExistingEntity\" not found")
 	assert.Nil(t, v)
 
 	v, err = FromRaw("non_existing_metric")("group1", "entity2", raw)
-	assert.EqualError(t, err, "metric not found")
+	assert.EqualError(t, err, "metric \"non_existing_metric\" not found")
 	assert.Nil(t, v)
```
roobre, line 55: Could we table-test this? Seems like a quick change.

src/definition/fetch_test.go
```diff
@@ -99,14 +98,14 @@ func TestTransformBypassesError(t *testing.T) {
...
 	v, err = Transform(FromRaw("metric_name_3"), transformFunc)("group1", "nonExistingEntity", raw)
-	assert.EqualError(t, err, "entity not found")
+	assert.EqualError(t, err, "entity \"nonExistingEntity\" not found")
 	assert.Nil(t, v)
 
 	v, err = Transform(FromRaw("non_existing_metric"), transformFunc)("group1", "entity2", raw)
-	assert.EqualError(t, err, "metric not found")
+	assert.EqualError(t, err, "metric \"non_existing_metric\" not found")
 	assert.Nil(t, v)
 }
```
roobre, line 111: As above, seems easy to build a table for this

src/kubelet/metric/metric.go
```diff
@@ -33,14 +33,12 @@ func GetMetricsData(c client.HTTPClient) (v1.Summary, error) {
...
 
-	var summary = new(v1.Summary)
-	err = json.Unmarshal(body, summary)
-	if err != nil {
+	summary := &v1.Summary{}
+	if err := json.Unmarshal(body, summary); err != nil {
 		return v1.Summary{}, fmt.Errorf("error unmarshaling the response body. Got error: %v", err.Error())
 	}
 
 	return *summary, nil
```
roobre, line 41:
```suggestion
	summary := v1.Summary{}
	if err := json.Unmarshal(body, &summary); err != nil {
		return v1.Summary{}, fmt.Errorf("error unmarshaling response into v1.Summary: %w", err)
	}

	return summary, nil
```

roobre: Not a huge fan of returning empty objects rather than nil pointers though.

src/kubelet/metric/metric.go
```diff
@@ -214,38 +211,31 @@ func GroupStatsSummary(statsSummary v1.Summary) (definition.RawGroups, []error)
...
 		}
+
 		g["pod"][rawEntityID] = rawPodMetrics
-	VolumeListLoop:
 		for _, volume := range pod.VolumeStats {
 			rawVolumeMetrics, err := fetchVolumeStats(volume)
 			if err != nil {
 				errs = append(errs, err)
-				continue VolumeListLoop
+				continue
```
roobre, line 226: Does this change preserve the old behavior? Seems like an error fetching volume stats will cause the whole pod to be skipped without the labels.

src/kubelet/metric/metric.go
```diff
@@ -214,38 +211,31 @@ func GroupStatsSummary(statsSummary v1.Summary) (definition.RawGroups, []error)
...
-			continue PodListLoop
-		}
-
-	ContainerListLoop:
 		for _, container := range pod.Containers {
 			rawContainerMetrics, err := fetchContainerStats(container)
 			if err != nil {
 				errs = append(errs, err)
-				continue ContainerListLoop
+				continue
```
roobre, line 238: Same as above, an error in a container will skip metrics for the whole pod.

src/metric/populate_test.go
```diff
@@ -144,24 +143,8 @@ func TestPopulateK8s(t *testing.T) {
 
 	k8sVersion := &version.Info{GitVersion: "v1.15.42"}
 	err = p.Populate(foo, kubeletSpecs, i, "test-cluster", k8sVersion)
-	assert.Error(t, err)
+	assert.Empty(t, err.(data.PopulateResult).Errors)
```
roobre, line 146: Wouldn't this assert pass if the type assertion fails?

src/prometheus/definition.go
```diff
@@ -78,16 +77,15 @@ func getLabels(groupLabel, rawEntityID, key string, groups definition.RawGroups,
...
 		if v == nil {
-			return s, fmt.Errorf("label %s not found for metric %s", label, key)
-
+			return s, fmt.Errorf("label %q not found for metric %q", label, key)
 		}
 
 		val, ok := v.(string)
 		if !ok {
-			return s, fmt.Errorf("incorrect type of label %s for metric %s", label, key)
+			return s, fmt.Errorf("incorrect type of label %q for metric %q", label, key)
```
roobre, line 88:
```suggestion
			return s, fmt.Errorf("unexpected type %T of label %q for metric %q", label, label, key)
```

src/prometheus/definition.go
```diff
@@ -100,17 +98,16 @@ func FromLabelValueEntityIDGenerator(key, label string) definition.EntityIDGener
...
 
 		if v == nil {
-			return "", fmt.Errorf("incorrect value of fetched data for metric %s", key)
+			return "", fmt.Errorf("incorrect value of fetched data for metric %q", key)
 		}
 
 		val, ok := v.(string)
 		if !ok {
-			return "", fmt.Errorf("incorrect type of fetched data for metric %s", key)
+			return "", fmt.Errorf("incorrect type of fetched data for metric %q", key)
```
roobre, line 110:
```suggestion
			return "", fmt.Errorf("unexpected type %T of fetched data for metric %q", val, key)
```

src/prometheus/definition.go
```diff
@@ -100,17 +98,16 @@ func FromLabelValueEntityIDGenerator(key, label string) definition.EntityIDGener
...
 		v, err := FromLabelValue(key, label)(groupLabel, rawEntityID, g)
 		if err != nil {
-
-			return "", fmt.Errorf("cannot fetch label %s for metric %s, %s", label, key, err)
+			return "", fmt.Errorf("cannot fetch label %q for metric %q: %w", label, key, err)
 		}
 
 		if v == nil {
-			return "", fmt.Errorf("incorrect value of fetched data for metric %s", key)
+			return "", fmt.Errorf("incorrect value of fetched data for metric %q", key)
```
roobre, line 105:
```suggestion
			return "", fmt.Errorf("uniexpected nil value of fetched data for metric %q", key)
```

src/metric/populate_test.go
invidian: It will panic if type assertion fails. Perhaps we should assert the type explicitly above, so the test suite does not panic.

src/definition/fetch_test.go
invidian: I think creating a struct with 5 fields here brings a little benefit here, right now test cases are very clear and readable. Little abstraction is worse than little duplication. We could wrap each case in sub-test, but it seems out of scope for this already big PR. I only touched areas of the code I was debugging. Perhaps there shouldn't be tests on `FromRaw` level at all if we take a look more into it.

src/definition/fetch_test.go
invidian: I'll mark this comment as duplicate to the other one.

src/kubelet/metric/metric.go
invidian: >Not a huge fan of returning empty objects rather than nil pointers though.

Agreed. I've made a bit more improvements along the way, please have a look.

src/kubelet/metric/metric.go
invidian: No, `continue` will skip the closest loop, in this case `for _, volume := range pod.VolumeStats {`, so yeah, it should be fine.

I can add tests for that.

src/kubelet/metric/metric.go
invidian: Will mark as duplicate.

src/kubelet/metric/metric.go
invidian: https://play.golang.org/p/_ezHOB-4wW3

src/metric/populate_test.go
invidian: I added `require.IsType` before we assert, so we don't panic.

src/kubelet/metric/metric.go
roobre: ... You are entirely right, I don't know why I thought otherwise 😕

cmd/kubernetes-static/main.go
```diff
@@ -145,31 +212,56 @@ func main() {
...
 
 	mux := http.NewServeMux()
 
-	dataDir := fmt.Sprintf("./cmd/kubernetes-static/data/%s", k8sMetricsVersion)
-
-	path, err := filepath.Abs(dataDir)
+	k8sContent, err := fs.Sub(content, filepath.Join("data", k8sMetricsVersion))
 	if err != nil {
-		log.Fatal(errors.New("cannot start server"))
+		panic(err)
```
roobre, line 233: Why prefer panic to `log.Fatal(err)`? Just curious

cmd/kubernetes-static/readme.md
```diff
@@ -4,23 +4,14 @@ Kubernetes Static is a project used to run the Kubernetes Integration locally on
 
 ## How it works
 
-The files in the `./data` folder are saved outputs from KSM and various kubelet endpoints, all required to run the full integration. 
+The files in the `./data` folder are saved outputs from KSM and various kubelet endpoints, all required to run the full integration.
 The program wil start a temporary HTTP server and serve these files. The groupers are configured to use these endpoints instead of discovering them.
```
roobre, line 8:
```suggestion
The files in the `./data` folder are saved outputs from KSM and various kubelet endpoints, which will be embedded at build-time and then served from a temporary HTTP. The groupers are configured to use these endpoints instead of discovering them.
```

src/ksm/group.go
```diff
@@ -48,20 +51,21 @@ func (r *ksmGrouper) Group(specGroups definition.SpecGroups) (definition.RawGrou
 	if err != nil {
 		return nil, &data.ErrorGroup{
 			Recoverable: false,
-			Errors:      []error{fmt.Errorf("error querying KSM. %s", err)},
+			Errors:      []error{fmt.Errorf("error querying KSM: %w", err)},
```
roobre, line 54:
```suggestion
			Errors:      []error{fmt.Errorf("querying KSM: %w", err)},
```

src/kubelet/metric/metric.go
```diff
@@ -18,29 +18,29 @@ import (
 const StatsSummaryPath = "/stats/summary"
 
 // GetMetricsData calls kubelet /stats/summary endpoint and returns unmarshalled response
-func GetMetricsData(c client.HTTPClient) (v1.Summary, error) {
+func GetMetricsData(c client.HTTPClient) (*v1.Summary, error) {
 	resp, err := c.Do(http.MethodGet, StatsSummaryPath)
 	if err != nil {
-		return v1.Summary{}, err
+		return nil, fmt.Errorf("performing GET request to kubelet endpoint: %w", err)
```
roobre, line 24:
```suggestion
		return nil, fmt.Errorf("performing GET request to kubelet endpoint %s: %w", StatsSummaryPath, err)
```

src/kubelet/metric/metric.go
```diff
@@ -18,29 +18,29 @@ import (
...
-		return v1.Summary{}, fmt.Errorf("error reading the response body of kubelet endpoint. Got error: %v", err.Error())
+		return nil, fmt.Errorf("reading the response body of kubelet endpoint (response code %d): %w", resp.StatusCode, err)
 	}
 
-	var summary = new(v1.Summary)
-	err = json.Unmarshal(body, summary)
-	if err != nil {
-		return v1.Summary{}, fmt.Errorf("error unmarshaling the response body. Got error: %v", err.Error())
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("received non-OK response code from kubelet: %d. Response body: %v", resp.StatusCode, body)
```
roobre, line 34: Why do this check after attempting to read the body? It should be fine to call `Close()` on it without consuming it first.

src/kubelet/metric/metric.go
```diff
@@ -18,29 +18,29 @@ import (
...
 	}
 
-	return *summary, nil
+	summary := &v1.Summary{}
 
+	if err := json.Unmarshal(body, summary); err != nil {
```
roobre, line 39: Could we consider using `json.NewDecoder(resp.Body).Decode(summary)`? I think it's easier to read if we are not doing weird stuff with the body.

src/kubelet/metric/metric.go
roobre: It might additionally be easier on memory for large payloads.

cmd/kubernetes-static/main.go
invidian: Ah, I'm not sure. Perhaps we can still use `log.Fatal` + fmt.Sprintf, so we can format the received error as well. Good point.

src/kubelet/metric/metric.go
invidian: Body may include the error message, so I guess it's useful to read it it for possible logging, to make debugging easier.

src/kubelet/metric/metric.go
roobre: > It might additionally be easier on memory for large payloads.
This comment was intended to go here.

cmd/kubernetes-static/main.go
invidian: Well, also the code above this use `panic()`. I changed both to use `log.Fatal` (which shouldn't be used at all BTW, generally one should favor returning `error` and doing `os.Exit()` in one place explicitly.

src/kubelet/metric/metric.go
invidian: We could, though then we lose the ability of reading the error message from the request, which might be for example:
```
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* Connection state changed (MAX_CONCURRENT_STREAMS == 250)!
< HTTP/2 403
< content-type: text/plain; charset=utf-8
< content-length: 92
< date: Fri, 16 Jul 2021 06:15:09 GMT
<
* Connection #0 to host 10.0.0.2 left intact
Forbidden (user=system:kube-controller-manager, verb=get, resource=nodes, subresource=stats)
```

src/kubelet/metric/metric.go
```diff
@@ -192,7 +191,11 @@ func fetchVolumeStats(v v1.VolumeStats) (definition.RawMetrics, error) {
 }
 
 // GroupStatsSummary groups specific data for pods, containers and node
-func GroupStatsSummary(statsSummary v1.Summary) (definition.RawGroups, []error) {
+func GroupStatsSummary(statsSummary *v1.Summary) (definition.RawGroups, []error) {
```
gsanchezgavier, line 194: this makes me think that statsSummary is going to be modified inside the function.

gsanchezgavier (approved):
LGTM thanks!

src/kubelet/metric/metric.go
roobre: We could check the status code, and if it is non-200 then print the body along the error. It seems reasonable to assume that a 200 status will come with a valid JSON payload.

src/kubelet/metric/metric.go
roobre: @ Rob Pike pls add `const` to go asap thanks

src/kubelet/metric/metric.go
invidian: That creates more cases, as error handling will have to read the body in a different way than the regular path. It seems more complicated to me without clear benefit.

src/kubelet/metric/metric.go
invidian: >@ Rob Pike pls add const to go asap thanks

I subscribed to https://github.com/golang/go/issues/22876 just yesterday :smile: I think Rust does a good job with mutable ref here.

>this makes me think that statsSummary is going to be modified inside the function.


@gsanchezgavier I understand your concern. However, as far as I know, most of the time you should expect that function will not modify it's parameters, but rather return a mutated copy of the data. Modifying in place is a performance optimization IMO.

Given that this function returns a different format, I think it should be obvious to assume the summary won't be modified.

roobre (approved):
LGTM, thanks and killer work @invidian !

PR #155, opened 2021-07-13
Add node status conditions metrics

invidian:
_Includes commits from #148_. Proper diff to look at is here: https://github.com/newrelic/nri-kubernetes/compare/invidian/improvements...invidian/add-node-status-metrics

Draft as the following things are missing:
- [x] Support for unknown value for conditions
- [x] Tests

### DIFF
--- a/src/metric/definition.go
+++ b/src/metric/definition.go
@@ -1100,7 +1100,7 @@ func toNumericBoolean(value definition.FetchedValue) (definition.FetchedValue, e
 	case "unknown":
 		return -1, nil
 	default:
-		return nil, errors.New("value can not be converted to numeric boolean")
+		return nil, fmt.Errorf("value '%v' can not be converted to numeric boolean", value)
 	}
 }
 
--- a/src/metric/definition.go
+++ b/src/metric/definition.go
@@ -1069,6 +1069,8 @@ func toNumericBoolean(value definition.FetchedValue) (definition.FetchedValue, e
 		return 1, nil
 	case "false", "False", false, 0:
 		return 0, nil
+	case "unknown":
+		return -1, nil
 	default:
 		return nil, errors.New("value can not be converted to numeric boolean")
 	}
--- a/src/metric/definition.go
+++ b/src/metric/definition.go
@@ -715,6 +715,22 @@ var KSMSpecs = definition.SpecGroups{
 			{Name: "isLimited", ValueFunc: prometheus.FromValue("kube_hpa_status_condition_limited")},
 		},
 	},
+	"node": {
+		TypeGenerator: kubeletMetric.FromRawGroupsEntityTypeGenerator,
+		Specs: []definition.Spec{
+			{
+				Name: "condition.*",
+				ValueFunc: definition.Transform(
+					// Will produce `condition.whatever = true` metrics from KSM output like
+					// kube_node_status_condition{condition="whatever",status="true"} 1.
+					prometheus.FromLabelValueWithLabelInName("kube_node_status_condition", "condition", "condition.%s", "status"),
+					// Will transform `condition.whatever = true` from previous step to `condition.whatever = 1`.
+					forEachFetchedValue(toNumericBoolean),
+				),
+				Type: sdkMetric.GAUGE,
+			},
+		},
+	},
 }
 
 // KSMQueries are the queries we will do to KSM in order to fetch all the raw metrics.
@@ -822,6 +838,18 @@ var KSMQueries = []prometheus.Query{
 	},
 	{MetricName: "kube_hpa_status_current_replicas"},
 	{MetricName: "kube_hpa_status_desired_replicas"},
+	// Node status condition
+	{MetricName: "kube_node_status_condition", Value: prometheus.QueryValue{
+		// Since we aggregate metrics which look like the following:
+		//
+		// kube_node_status_condition{node="minikube",condition="MemoryPressure",status="true"} 0
+		// kube_node_status_condition{node="minikube",condition="MemoryPressure",status="false"} 1
+		// kube_node_status_condition{node="minikube",condition="MemoryPressure",status="unknown"} 0
+		//
+		// KSM should never produce a positive value for more than one status, so we can simply fetch
+		// only values which has value 1 for processing.
+		Value: prometheus.GaugeValue(1),
+	}},
 }
 
 // CadvisorQueries are the queries we will do to the kubelet metrics cadvisor endpoint in order to fetch all the raw metrics.
@@ -1096,6 +1124,26 @@ func fromPrometheusNumeric(value definition.FetchedValue) (definition.FetchedVal
 	}
 
 	return nil, fmt.Errorf("invalid type value '%v'. Expected 'gauge' or 'counter', got '%T'", value, value)
+}
+
+func forEachFetchedValue(tf definition.TransformFunc) definition.TransformFunc {
+	return func(value definition.FetchedValue) (definition.FetchedValue, error) {
+		values, ok := value.(definition.FetchedValues)
+		if !ok {
+			return nil, fmt.Errorf("value of type %T is not a definition.FetchedValues", value)
+		}
+
+		for k, value := range values {
+			newValue, err := tf(value)
+			if err != nil {
+				return nil, fmt.Errorf("converting value %q: %w", k, err)
+			}
+
+			values[k] = newValue
+		}
+
+		return values, nil
+	}
 }
 
 // Subtract returns a new FetchFunc that subtracts 2 values. It expects that the values are float64
--- a/src/prometheus/definition.go
+++ b/src/prometheus/definition.go
@@ -726,3 +726,62 @@ func getRawEntityID(parentGroupLabel, groupLabel, entityID string, groups defini
 	}
 	return rawEntityID, nil
 }
+
+// FromLabelValueWithLabelInName returns a FetchFunc that returns metrics with a templatizable name based on a label,
+// and whose value is gathered from another label.
+//
+// `format` will be evaluated using the value of the label `label` and result will be used as metric name.
+//
+// The value of the metric will be set to the value of `valueLabel` label.
+//
+// For example, for the following metric:
+//
+// rawMetricName{foo="whatever",bar="false"} 1
+//
+// called as:
+//
+// FromLabelValueWithLabelInName("rawMetricName", "foo", "myMetric.%s", "bar")
+//
+// the function will produce the following metric:
+//
+// myMetric.whatever = "false"
+//
+func FromLabelValueWithLabelInName(rawMetricName, label, format, valueLabel string) definition.FetchFunc {
+	return func(groupLabel, entityID string, rawGroups definition.RawGroups) (definition.FetchedValue, error) {
+		fetchedValues := definition.FetchedValues{}
+
+		metricsRaw, err := definition.FromRaw(rawMetricName)(groupLabel, entityID, rawGroups)
+		if err != nil {
+			return nil, fmt.Errorf("fetching metric %q for entity %q in group %q: %w", rawMetricName, entityID, groupLabel, err)
+		}
+
+		metrics := []Metric{}
+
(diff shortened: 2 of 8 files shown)


src/prometheus/definition.go
```diff
@@ -701,3 +724,41 @@ func getRawEntityID(parentGroupLabel, groupLabel, entityID string, groups defini
...
+		fetchedValues := definition.FetchedValues{}
+		metricsRaw, err := definition.FromRaw(rawMetricName)(groupLabel, entityID, rawGroups)
+		if err != nil {
+			return nil, fmt.Errorf("fetching metric %q for entity %q in group %q: %w", rawMetricName, entityID, groupLabel, err)
+		}
+
+		metrics, ok := metricsRaw.([]Metric)
+		if !ok {
+			return nil, fmt.Errorf("incompatible metric type for %q. Expected: []Metric. Got: %T", rawMetricName, metricsRaw)
+		}
```
roobre, line 739: Will this work if KSM only reports one condition?

src/prometheus/definition.go
```diff
@@ -701,3 +724,41 @@ func getRawEntityID(parentGroupLabel, groupLabel, entityID string, groups defini
 	}
 	return rawEntityID, nil
 }
+
+func WithLabelValueInMetricNameAndWithLabelValueMappedToMetricValue(rawMetricName, label, format, valueLabel string, valueMap map[string]int) definition.FetchFunc {
```
roobre, line 728: I would at least decouple the "valueMap" logic from here, if possible. There is already a TransformFunc concept somewhere we could reuse.

roobre: Since what we are doing is flattening a label to the metric name, perhaps we could call it `FromValueWithLabelInName`?

src/prometheus/definition.go
```diff
@@ -701,3 +724,41 @@ func getRawEntityID(parentGroupLabel, groupLabel, entityID string, groups defini
...
+
+		metrics, ok := metricsRaw.([]Metric)
+		if !ok {
+			return nil, fmt.Errorf("incompatible metric type for %q. Expected: []Metric. Got: %T", rawMetricName, metricsRaw)
+		}
+
+		for _, metric := range metrics {
+			nameLabelValue, ok := metric.Labels[label]
+			if !ok {
+				return nil, fmt.Errorf("label %q for metric name not found in metric %v", label, metric)
```
roobre, line 744: Would it be better to error here, or just `continue` in case a particular metric does not contain the label?

src/prometheus/definition.go
```diff
@@ -701,3 +724,41 @@ func getRawEntityID(parentGroupLabel, groupLabel, entityID string, groups defini
...
+			}
+
+			fetchedValueName := fmt.Sprintf(format, nameLabelValue)
+
+			valueLabelValue, ok := metric.Labels[valueLabel]
```
roobre, line 749: Not very happy about the naming. If this is the intermediate value, what about
```suggestion
			originalMetricValue, ok := metric.Labels[valueLabel]
```

src/prometheus/definition.go
```diff
@@ -701,3 +724,41 @@ func getRawEntityID(parentGroupLabel, groupLabel, entityID string, groups defini
...
+			if !ok {
+				return nil, fmt.Errorf("label %q for metric value not found in metric %v", valueLabel, metric)
+			}
+
+			metricValue, ok := valueMap[valueLabelValue]
+			if !ok {
+				return nil, fmt.Errorf("key %q not found in value map %v", valueLabelValue, valueMap)
```
roobre, line 756: Might want to log+continue on this one, so a malformed label does not break the rest of the metrics.

src/prometheus/definition.go
```diff
@@ -701,3 +724,41 @@ func getRawEntityID(parentGroupLabel, groupLabel, entityID string, groups defini
...
+			}
+
+			fetchedValues[fetchedValueName] = metricValue
+		}
+
+		return fetchedValues, nil
```
roobre, line 762: If we switch to a more lenient parsing as I suggested, we would need to add a non-zero length here in case we were not able to parse any metric and return an error.

roobre:
I would also update the JsonSchema files for e2e test to capture this, with the basic kubelet conditions.

src/metric/definition.go
```diff
@@ -715,6 +715,16 @@ var KSMSpecs = definition.SpecGroups{
 			{Name: "isLimited", ValueFunc: prometheus.FromValue("kube_hpa_status_condition_limited")},
 		},
 	},
+	"node": {
+		TypeGenerator: kubeletMetric.FromRawGroupsEntityTypeGenerator,
```
gsanchezgavier, line 719: Isn't the `IDGenerator` needed here to identify each node? something like this 
```
IDGenerator:   prometheus.FromLabelValueEntityIDGenerator("kube_node_status_condition", "node"),
```

roobre: Good point, I didn't notice that. However I just checked and the `node` section for the kubelet metrics do not have an ID generator either. I think this might be related to nodes not being complete "entities", i.e. they are not registered in the backend as such.

But I'd leave this open so we don't forget to doublecheck this.

gsanchezgavier: mmm interesting i made some test and is not needed, i will try to find where this magic is happen

gsanchezgavier: I think i [found it](https://github.com/newrelic/nri-kubernetes/blob/main/src/prometheus/definition.go#L211).

src/prometheus/definition.go
invidian: Good point, it will not. Though it is guaranteed that multiple will be returned as kubelet itself reports 4 standard ones.

invidian: I added suport for both single metric and multiple metrics, though it's a bit annoying. I think eventually all metrics should be slices.

src/prometheus/definition.go
invidian: Sounds good, though that would be `FromLabelValueWithLabelInName` :smile:

invidian: Changed, though I'm not sure if it's better, as Transform function require bunch of type assertions again.

src/prometheus/definition.go
invidian: It would be recommended to log this error and I don't have access to the logger here, so I'm not sure. I mean, I don't see a best-effort way forward here at the moment and I wouldn't swallow the errors.

src/prometheus/definition.go
invidian: Yeah, name is not great. Maybe `labelValue`? Or `requestedLabelValue`? I wanted to indicate it's a label value, which will become a metric value. Or `metricValueFromLabel`? Or `fetchedValuefromLabel`?

src/metric/definition.go
invidian: Yeah, this is simply copied from https://github.com/newrelic/nri-kubernetes/blob/e34287a6a9e8f06cf883b57a58ca501bd862ab6b/src/metric/definition.go#L915-L916

invidian:
>I would also update the JsonSchema files for e2e test to capture this, with the basic kubelet conditions.

Done, I think :smile:

src/prometheus/definition.go
```diff
@@ -701,3 +724,42 @@ func getRawEntityID(parentGroupLabel, groupLabel, entityID string, groups defini
 	}
 	return rawEntityID, nil
 }
+
+func FromLabelValueWithLabelInName(rawMetricName, label, format, valueLabel string) definition.FetchFunc {
```
gsanchezgavier, line 728: I think it would be nice some description here perhaps with an example :)

invidian: Right. My idea was that when we roughly agree on the implementation, I'll add proper unit tests :)

gsanchezgavier: I would rename `label` parameter to something that express that the value of that label will be used in the metric name resulted.
Perhaps `nameLabel` ? but perhaps is not needed with the descriptions and example

src/prometheus/definition.go
gsanchezgavier: `fetchedValuefromLabel` +1

src/prometheus/definition.go
invidian: I agree, that's a good idea.

invidian: I hate so much fact that this function is doing 2 things and I cannot really easily compose those 2 things together instead of hardcoding them in a single function :smile:

roobre:
E2E tests seem great to me, thanks!

src/metric/definition.go
```diff
@@ -822,6 +835,10 @@ var KSMQueries = []prometheus.Query{
 	},
 	{MetricName: "kube_hpa_status_current_replicas"},
 	{MetricName: "kube_hpa_status_desired_replicas"},
+	// Node status condition
+	{MetricName: "kube_node_status_condition", Value: prometheus.QueryValue{
+		Value: prometheus.GaugeValue(1),
```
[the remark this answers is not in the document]
invidian: I'll add a comment (though other metrics do not have them :)). It simply tells to fetch only metrics which value is `1`, which makes it fetch only active conditions, which we care about.

src/prometheus/definition.go
invidian: I think the following will be more accurate:
```suggestion
// FromLabelValueWithLabelInName returns a FetchFunc that returns metrics with a templatizable name based on a label,
// and whose value is gathered from another label.
//
// `format` will be evaluated using the value of the label `label` and result will be used as metric name.
//
// The value of the metric will be set to the value of `valueLabel` label.
//
// For example, for the following metric:
//
// rawMetricName{foo="whatever",bar="false"} 1
//
// called as:
//
// FromLabelValueWithLabelInName("rawMetricName", "foo", "myMetric.%s", "bar")
//
// the function will produce the following metric:
//
// myMetric.whatever = "false"
//
func FromLabelValueWithLabelInName(rawMetricName, label, format, valueLabel string) definition.FetchFunc {
```

src/metric/definition.go
```diff
@@ -1035,6 +1052,35 @@ func toUtilization(dividendMetric, divisorMetric string) definition.FetchFunc {
 	}
 }
 
+func conditionStatusesToGauge(value definition.FetchedValue) (definition.FetchedValue, error) {
```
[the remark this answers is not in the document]
invidian: Can do, but it may have unexpected behavior.

src/prometheus/definition.go
invidian: Changed.

src/prometheus/definition.go
roobre: I think that together they are longer but also makes them slightly easier to read, in my view. I'd settle for the current approach.

src/prometheus/definition.go
roobre: Fucking non-global logger... Let's error out then,

src/prometheus/definition.go
roobre: Looks great!

src/metric/definition.go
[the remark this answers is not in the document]
roobre: Let's settle with the custom one then, I'm happy provided there's a comment explaining what it does :)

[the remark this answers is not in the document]
invidian: I already pushed merged version, I guess it should be fine :tm:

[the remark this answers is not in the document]
invidian: Ah, actually not. `toNumericBoolean` only supports single value, while we need to support a collection.

roobre (approved):

roobre:
Looking great, thanks @invidian!

invidian:
>Looking great, thanks @invidian!

@roobre thanks. I just pushed the change to e2e tests instead of how MetricSets are composed as we agreed. I think this is now ready to review.

gsanchezgavier (approved):
Thanks!

PR #156, opened 2021-07-14
build: pick up integration version from TAG env var

### LINKED ISSUE #131: Inject Integration Version
Right now Integration Version comes from a constant, we should have it injected from the pipeline

roobre:
SHA1 of HEAD will also be picked up automatically by Makefile

Fixes #131

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,6 +21,7 @@ jobs:
         env:
           GOOS: ${{ matrix.goos }}
           GOARCH: ${{ matrix.goarch }}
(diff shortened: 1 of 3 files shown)


gsanchezgavier (approved):
thanks looks good! no more replace!

paologallinaharbur:
the version is still hardcoded in deploy manifests. Can we fix that as well?

roobre:
@paologallinaharbur I always find tricky to automate text replacements in the CI, it's very easy to make a change in the manifest and cause the script to fail unexpecedly.

Moreover we will be getting rid of those manifests soon, hopefullly when we make the k8s version cutoff as well :)

PR #159, opened 2021-07-16
feat: add unschedulable node metric

### LINKED ISSUE #139: Add support for kube_node_spec_unschedulable metric
Apart from node condition we can leverage as well kube_node_spec_unschedulable.

Check if it supported across vendors and its source and if so add it to the nodeSample

gsanchezgavier:
closes #139

### DIFF
--- a/e2e/schema/node-no-network.json
+++ b/e2e/schema/node-no-network.json
@@ -312,6 +312,13 @@
       "examples": [
         1449194914
       ]
+    },
+    "unschedulable": {
+      "$id": "#/properties/unschedulable",
+      "type": "integer",
+      "examples": [
+        0
+      ]
     }
   },
 
@@ -363,7 +370,8 @@
       "runtimeInodes": 3145728,
       "runtimeInodesFree": 2875132,
       "runtimeInodesUsed": 270596,
-      "runtimeUsedBytes": 1449194914
+      "runtimeUsedBytes": 1449194914,
+      "unschedulable": 0
     }
   ],
   "required": [
@@ -410,6 +418,7 @@
     "runtimeInodes",
     "runtimeInodesFree",
     "runtimeInodesUsed",
-    "runtimeUsedBytes"
+    "runtimeUsedBytes",
+    "unschedulable"
   ]
 }
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -18,3 +18,5 @@ Then you can run manually any scenario you are interested into
 helm dependencies update ./e2e/charts/newrelic-infrastructure-k8s-e2e
 go run e2e/cmd/e2e.go --verbose --cluster_name=e2e --nr_license_key="fakeLicense" --rbac=true --integration_image_tag=test --integration_image_repository=test_image_normal  --k8s_version=v1.19.4
 ```
+
+Note: On macOS you might have to do run `minikube start --vm` to get a proper IP that will be used by the e2e test to check the cluster flavor. 
--- a/e2e/charts/newrelic-infrastructure-k8s-e2e/values.yaml
+++ b/e2e/charts/newrelic-infrastructure-k8s-e2e/values.yaml
@@ -48,9 +48,9 @@ ksm-instance-one:
     namespaces: true
     replicasets: true
     pods: true
+    nodes: true
 
     certificatesigningrequests: false
-    nodes: true
     replicationcontrollers: false
     resourcequotas: false
     cronjobs: false
@@ -73,9 +73,9 @@ ksm-instance-two:
     namespaces: true
     replicasets: true
     pods: true
+    nodes: true
 
     certificatesigningrequests: false
-    nodes: true
     replicationcontrollers: false
     resourcequotas: false
     cronjobs: false
--- a/e2e/schema/node.json
+++ b/e2e/schema/node.json
@@ -333,6 +333,13 @@
       "examples": [
         1449194914
       ]
+    },
+    "unschedulable": {
+      "$id": "#/properties/unschedulable",
+      "type": "integer",
+      "examples": [
+        0
+      ]
     }
   },
 
@@ -384,7 +391,8 @@
       "condition.DiskPressure": 0,
       "condition.MemoryPressure": 0,
       "condition.PIDPressure": 0,
-      "condition.Ready": 1
+      "condition.Ready": 1,
+      "unschedulable": 0
     }
   ],
   "required": [
@@ -434,6 +442,7 @@
     "runtimeInodes",
     "runtimeInodesFree",
     "runtimeInodesUsed",
-    "runtimeUsedBytes"
+    "runtimeUsedBytes",
+    "unschedulable"
   ]
 }
--- a/src/metric/definition.go
+++ b/src/metric/definition.go
@@ -729,6 +729,7 @@ var KSMSpecs = definition.SpecGroups{
 				),
 				Type: sdkMetric.GAUGE,
 			},
+			{Name: "unschedulable", ValueFunc: prometheus.FromValue("kube_node_spec_unschedulable"), Type: sdkMetric.GAUGE},
 		},
 	},
 }
@@ -850,6 +851,7 @@ var KSMQueries = []prometheus.Query{
 		// only values which has value 1 for processing.
 		Value: prometheus.GaugeValue(1),
 	}},
+	{MetricName: "kube_node_spec_unschedulable"},
 }
 
 // CadvisorQueries are the queries we will do to the kubelet metrics cadvisor endpoint in order to fetch all the raw metrics.


.gitignore
```diff
@@ -4,3 +4,4 @@ deploy/local-unprivileged.yaml
 deploy/local-openshift.yaml
 bin
 .idea
+.vscode
```
invidian, line 7: https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer maybe?

src/definition/populate.go
```diff
@@ -85,7 +85,14 @@ func IntegrationProtocol2PopulateFunc(
 					continue
 				}
 
-				ms := e.NewMetricSet(msType)
```
invidian, line 88: ~This won't help with our issue because different jobs calls populate and `IntegrationProtocol2PopulateFunc` multiple times anyway. We would have to globally share the metric sets map by type, which doesn't seem like a good idea.~

e2e/schema/node-no-network.json
```diff
@@ -345,13 +345,11 @@
   "required": [
     "allocatableCpuCores",
     "allocatableEphemeralStorageBytes",
-    "allocatableHugepages1Gi",
```
invidian, line 340: Why do we need to remove those? Some explanation would be nice.

src/definition/populate.go
```diff
@@ -85,7 +85,14 @@ func IntegrationProtocol2PopulateFunc(
...
+				var ms metric.MetricSet
+				if len(e.Metrics) > 0 {
+					// This avoid generating a new ms for metrics collected from different sources like kubelet and ksm.
+					ms = e.Metrics[0]
```
invidian, line 91: Oooh, so entities are shared globally. That's horrible and good at the same time :smile: This seems better than my merging code in the e2e tests (though mine has less production impact I guess).

src/definition/populate.go
```diff
@@ -85,7 +85,14 @@ func IntegrationProtocol2PopulateFunc(
...
+				} else {
+					ms = e.NewMetricSet(msType)
+				}
```
invidian, line 94: How about:
```suggestion
				ms := e.NewMetricSet(msType)
				if len(e.Metrics) > 0 {
					// This avoid generating a new ms for metrics collected from different sources like kubelet and ksm.
					ms = e.Metrics[0]
				}
```

So we can avoid using `var`, explicit type declaration and `else` :smile:

src/definition/populate_test.go
```diff
@@ -80,6 +80,76 @@ func metricsNamingManipulator(ms metric.MetricSet, entity sdk.Entity, clusterNam
 	return ms.SetMetric("displayName", entity.Name, metric.ATTRIBUTE)
 }
 
+func TestIntegrationProtocol2PopulateFunc_AddingNewMetricSetToEntity(t *testing.T) {
```
invidian, line 83: A bit out of local conventions, but perhaps test name could be more descriptive, e.g.
```suggestion
func Test_IntegrationProtocol2Populate_generates_single_metric_set_for_single_entity_for_metrics_from_multiple_groups(t *testing.T) {
```

src/definition/populate_test.go
```diff
@@ -80,6 +80,76 @@ func metricsNamingManipulator(ms metric.MetricSet, entity sdk.Entity, clusterNam
...
+			Specs: []Spec{
+				{"metric_1", FromRaw("raw_metric_name_1"), metric.GAUGE, true},
+				{"metric_2", FromRaw("raw_metric_name_2"), metric.ATTRIBUTE, true},
+			},
+		},
+	}
+	integration, err := sdk.NewIntegrationProtocol2("nr.test", "1.0.0", new(struct{}))
+	if err != nil {
+		t.Fatal()
+	}
```
invidian, line 110: Perhaps `require.Nil()` can be used here? It seems odd to mix `assert` and regular error handling :thinking:

src/definition/populate_test.go
```diff
@@ -80,6 +80,76 @@ func metricsNamingManipulator(ms metric.MetricSet, entity sdk.Entity, clusterNam
...
+		integration,
+		defaultNS,
+		&version.Info{GitVersion: "v1.15.42"},
+		fromGroupMetricSetTypeGuessFunc,
+		metricsNamingManipulator,
+		clusterMetricsManipulator,
+	)(rawGroupsSample1, specs)
+	assert.True(t, populated)
+	assert.Empty(t, errs)
+	assert.Contains(t, integration.Data, &expectedEntityData)
```
invidian, line 136: It seems to me those asserts should be part of other test. We should only test the behavior of multiple metricsets here I think.

So those asserts are perhaps duplicated in another test.

This will also make test easier, as we won't need to further mutate `expectedMetricSet`.

src/definition/populate_test.go
```diff
@@ -80,6 +80,76 @@ func metricsNamingManipulator(ms metric.MetricSet, entity sdk.Entity, clusterNam
...
+	specs := SpecGroups{
+		"test": SpecGroup{
+			TypeGenerator: fromGroupEntityTypeGuessFunc,
+			Specs: []Spec{
+				{"metric_1", FromRaw("raw_metric_name_1"), metric.GAUGE, true},
+				{"metric_2", FromRaw("raw_metric_name_2"), metric.ATTRIBUTE, true},
+			},
+		},
+	}
+	integration, err := sdk.NewIntegrationProtocol2("nr.test", "1.0.0", new(struct{}))
```
invidian, line 107: Why using `new` here?
```suggestion
	integration, err := sdk.NewIntegrationProtocol2("nr.test", "1.0.0", struct{}{})
```

invidian (changes requested):
Some suggestions, but overall LGTM

src/definition/populate.go
gsanchezgavier: I agree , at the end i wanted to discuss this also with @roobre  since all these related changes will not be needed if moving to v3 happen , but not sure if we want that or not ( and also that has even more production impact)

src/definition/populate.go
gsanchezgavier: this was my first choise but then realized that when `e.NewMetricSet(msType)` is called , is being added to the entity already and that is exactly what i want it to avoid with this ~hack~ fix

src/definition/populate_test.go
gsanchezgavier: nice, I will use that name. but i think the rest of the tests has the convention `TestNameOfFunc_NameOfTest` I will follow that to be consistent with the rest.

src/definition/populate_test.go
gsanchezgavier: definitely this was purely c&p :shame:

src/definition/populate_test.go
gsanchezgavier: you're right i will remove this , there is no point to double test this.

e2e/schema/node-no-network.json
gsanchezgavier: This metric was not present on my local minikube environment, i didn't spent any time looking at the root cause , I will check it.

src/definition/populate_test.go
gsanchezgavier: I was missing old style C++ 🤣 , same c&p :shame:

e2e/schema/node-no-network.json
invidian: I mean, perhaps we should have some alternative test for this metric then?

src/definition/populate.go
invidian: Make sense, then maybe:

```suggestion
                              if len(e.Metrics) == 0 {
				        e.Metrics = append(e.Metrics, e.NewMetricSet(msType))
					// This avoid generating a new ms for metrics collected from different sources like kubelet and ksm.
				}
				
				ms := e.Metrics[0]
```

It seems even better to the previous suggestion :smile:

gsanchezgavier: I like that! , I was not convinced about my if condition `len(e.Metrics) > 0` , but is liked the use of the else for readability, I almost never had the good chance to use it 😄

src/definition/populate_test.go
gsanchezgavier: mmm interesting ,panics when initialize without new, looks like reflection is not being correctly handled in the sdk
```
--- FAIL: TestIntegrationProtocol2PopulateFunc_GeneratesSingleMetricSetForSingleEntityForMetricsFromMultipleGroups (0.00s)
panic: reflect: call of reflect.Value.Elem on struct Value [recovered]
	panic: reflect: call of reflect.Value.Elem on struct Value
```

src/definition/populate.go
gsanchezgavier: had to modify your suggestion a bit since the append is already happening inside the New function

src/definition/populate_test.go
invidian: Maybe it must be a pointer type?

invidian:
```diff
diff --git src/definition/populate_test.go src/definition/populate_test.go
index 440d170..0d0c248 100644
--- src/definition/populate_test.go
+++ src/definition/populate_test.go
@@ -105,7 +105,7 @@ func TestIntegrationProtocol2PopulateFunc_GeneratesSingleMetricSetForSingleEntit
                        },
                },
        }
-       integration, err := sdk.NewIntegrationProtocol2("nr.test", "1.0.0", new(struct{}))
+       integration, err := sdk.NewIntegrationProtocol2("nr.test", "1.0.0", &struct{}{})
        if err != nil {
                t.Fatal()
        }
```

Seems to work, but perhaps we can just stay consistent with `new()` in other test cases.

e2e/schema/node-no-network.json
gsanchezgavier: i was not able to launch minikube locally having this resource 😞 to document it somewhere , do you have any suggestions? I fear if that we do not remove from required others might suffer the same as I locally.

.gitignore
roobre: TIL this is a thing

src/definition/populate.go
roobre: Local DEVELOPER will do ANYTHING to avoid `else`

e2e/schema/node-no-network.json
gsanchezgavier: I removed this commit , i will remove this metrics locally

src/definition/populate.go
gsanchezgavier: we agree on remove this approach and follow the one merging the metric sets in the e2e test.

invidian (approved):
LGTM

PR #163, opened 2021-07-20
main: refactor main package

roobre:
Split from #160 

- Make argumentList a smarter object
- Wrap most of the main logic in a run function which returns an error
- Make logger global (for main package only)
- Minor improvements to error messages and formatting

PR #165, opened 2021-07-20
Add KSM namespace selector flag support

### LINKED ISSUE #105: Allow to namespace discovery operations

The performance impact of non-namespaced resource listings scales linearly with the size of a K8s cluster, being particularly heavy in the etcd component of the control plane.

Currently, we need to perform several listings in order to discover the location of some deployments, notably KSM. We could make this discovery lighter in the API server by limiting resource listing to a single namespace.

## Feature Description

A new config option, `ksmNamespace` could be added. We could prepopulate this variable in the helm chart from the namespace where the solution is being deployed (when ksm dependency is enabled), and also let the user override it as well.

This would require changes in the `Kubernetes` interface (src/client/client.go:22) and all its usages.

invidian:
Closes #105

Draft as tests are missing.

### DIFF
--- a/e2e/cmd/e2e.go
+++ b/e2e/cmd/e2e.go
@@ -192,6 +192,9 @@ func main() {
 			// not the logic of KSM. This might change if KSM sharding becomes enabled by default.
 			s.TwoKSMInstances = true
 		},
+		"with_namespace_filtering": func(s *scenario.Scenario) {
+			s.ExtraArgs = []string{"-kube_state_metrics_namespace=default"}
+		},
 	}
 
 	for scenarioName, mutateTestScenario := range scenarios {
--- a/e2e/cmd/e2e.go
+++ b/e2e/cmd/e2e.go
@@ -86,14 +86,14 @@ const (
 
 var allJobs = [...]job{jobKSM, jobKubelet, jobScheduler, jobEtcd, jobControllerManager, jobAPIServer}
 
-func (e *testEnv) execIntegration(pod v1.Pod, ksmPod *v1.Pod) (*integrationData, error) {
-	defer timer.Track(time.Now(), fmt.Sprintf("execIntegration func for pod %s", pod.Name), e.logger)
+func (se *scenarioEnv) execIntegration(pod v1.Pod, ksmPod *v1.Pod) (*integrationData, error) {
+	defer timer.Track(time.Now(), fmt.Sprintf("execIntegration func for pod %s", pod.Name), se.logger)
 
 	d := &integrationData{
 		podName: pod.Name,
 	}
 
-	output, err := e.k8sClient.PodExec(namespace, pod.Name, nrContainer, "/var/db/newrelic-infra/newrelic-integrations/bin/nri-kubernetes", "-timeout=30000", "-verbose")
+	output, err := se.k8sClient.PodExec(namespace, pod.Name, nrContainer, "/var/db/newrelic-infra/newrelic-integrations/bin/nri-kubernetes", "-timeout=30000", "-verbose")
 	if err != nil {
 		return nil, fmt.Errorf("executing command inside pod: %w", err)
 	}
@@ -108,7 +108,7 @@ func (e *testEnv) execIntegration(pod v1.Pod, ksmPod *v1.Pod) (*integrationData,
 		}
 	}
 
-	e.logger.Printf("Pod: %s, hostIP %s, expected jobs: %#v", pod.Name, pod.Status.HostIP, d.expectedJobs)
+	se.logger.Printf("Pod: %s, hostIP %s, expected jobs: %#v", pod.Name, pod.Status.HostIP, d.expectedJobs)
 
 	return d, nil
 }
@@ -386,13 +386,13 @@ func (se *scenarioEnv) executeTests(ksmPod *v1.Pod, releaseName string) error {
 	return nil
 }
 
-func (e *testEnv) executeIntegrationForAllPods(ksmPod *v1.Pod, nrPods *v1.PodList) (map[string]*integrationData, error) {
+func (se *scenarioEnv) executeIntegrationForAllPods(ksmPod *v1.Pod, nrPods *v1.PodList) (map[string]*integrationData, error) {
 	output := map[string]*integrationData{}
 
 	for _, p := range nrPods.Items {
-		e.logger.Debugf("Executing integration inside pod: %s", p.Name)
+		se.logger.Debugf("Executing integration inside pod: %s", p.Name)
 
-		integrationData, err := e.execIntegration(p, ksmPod)
+		integrationData, err := se.execIntegration(p, ksmPod)
 		if err != nil {
 			return output, fmt.Errorf("pod %q: %w", p.Name, err)
 		}
--- a/e2e/cmd/e2e.go
+++ b/e2e/cmd/e2e.go
@@ -93,7 +93,15 @@ func (se *scenarioEnv) execIntegration(pod v1.Pod, ksmPod *v1.Pod) (*integration
 		podName: pod.Name,
 	}
 
-	output, err := se.k8sClient.PodExec(namespace, pod.Name, nrContainer, "/var/db/newrelic-infra/newrelic-integrations/bin/nri-kubernetes", "-timeout=30000", "-verbose")
+	args := []string{
+		"/var/db/newrelic-infra/newrelic-integrations/bin/nri-kubernetes",
+		"-timeout=30000",
+		"-verbose",
+	}
+
+	args = append(args, se.scenario.ExtraArgs...)
+
+	output, err := se.k8sClient.PodExec(namespace, pod.Name, nrContainer, args...)
 	if err != nil {
 		return nil, fmt.Errorf("executing command inside pod: %w", err)
 	}
--- a/src/client/client.go
+++ b/src/client/client.go
@@ -26,13 +26,13 @@ type Kubernetes interface {
 	FindNode(name string) (*v1.Node, error)
 
 	// FindPodsByLabel returns a PodList reference containing the pods matching the provided label selector.
-	FindPodsByLabel(labelSelector metav1.LabelSelector) (*v1.PodList, error)
+	FindPodsByLabel(namespace string, labelSelector metav1.LabelSelector) (*v1.PodList, error)
 
 	// FindServicesByLabel returns a ServiceList containing the services matching the provided label selector.
-	FindServicesByLabel(labelSelector metav1.LabelSelector) (*v1.ServiceList, error)
+	FindServicesByLabel(namespace string, labelSelector metav1.LabelSelector) (*v1.ServiceList, error)
 
 	// ListServices returns a ServiceList containing all the services.
-	ListServices() (*v1.ServiceList, error)
+	ListServices(namespace string) (*v1.ServiceList, error)
 
 	// Config returns a config of API client
 	Config() *rest.Config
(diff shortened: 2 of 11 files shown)


invidian:
Still dependent on #163, but tests are now passing. I wonder if we need something extra.

src/ksm/client/discovery.go
```diff
@@ -216,17 +219,18 @@ func (sd *discoverer) nodeIP() (string, error) {
 
 // NewDiscoverer instantiates a new Discoverer required for discovering node IP
 // of kube-state-metrics pod and endpoint of kube-state-metrics service
-func NewDiscoverer(logger log.Logger, kubernetes client.Kubernetes) client.Discoverer {
-	return NewStaticEndpointDiscoverer("", logger, kubernetes)
+func NewDiscoverer(logger log.Logger, kubernetes client.Kubernetes, namespace string) client.Discoverer {
+	return NewStaticEndpointDiscoverer("", logger, kubernetes, namespace)
 }
```
roobre, line 224: Constructor with optional args triggers me (specially since now there are more than two args). Would it be feasible to add a `Namespace()` method to the object, that would set this internally? Or perhaps as an argument of the `Discover()` method?

src/ksm/client/distributed_pod_label_discovery.go
```diff
@@ -36,7 +35,7 @@ func (p *distributedPodLabelDiscoverer) findAllLabeledPodsRunningOnNode() ([]v1.
 		}
 
 		if pod.Status.HostIP == p.ownNodeIP {
-			p.logger.Debugf("Found KSM pod running on this code, pod IP: %s", pod.Status.PodIP)
+			p.logger.Debugf("Found KSM pod running on this node, pod IP: %s", pod.Status.PodIP)
```
roobre, line 38: There won't be more KSM pods running in this code anymore :(

src/ksm/client/distributed_pod_label_discovery.go
```diff
@@ -70,11 +69,12 @@ func (p *distributedPodLabelDiscoverer) Discover(timeout time.Duration) ([]clien
 }
 
 // NewDistributedPodLabelDiscoverer creates a new KSM discoverer that will find KSM pods using k8s labels
-func NewDistributedPodLabelDiscoverer(ksmPodLabel string, nodeIP string, logger log.Logger, k8sClient client.Kubernetes) client.MultiDiscoverer {
+func NewDistributedPodLabelDiscoverer(ksmPodLabel, ksmNamespace, nodeIP string, logger log.Logger, k8sClient client.Kubernetes) client.MultiDiscoverer {
```
roobre, line 72: So many args :(

src/ksm/client/distributed_pod_label_discovery.go
```diff
@@ -6,27 +6,26 @@ import (
...
-	pods, err := p.k8sClient.FindPodsByLabel(p.ksmPodLabel, "true")
-
+	pods, err := p.k8sClient.FindPodsByLabel(p.ksmNamespace, fmt.Sprintf("%s=true", p.ksmPodLabel))
 	if err != nil {
-		return nil, errors.Wrap(err, "could not query api server for pods")
+		return nil, fmt.Errorf("querying API server for pods: %w", err)
 	}
 	if len(pods.Items) == 0 {
-		return nil, errors.Wrapf(errNoKSMPodsFound, "no KSM pod found with label: '%s'", p.ksmPodLabel)
+		return nil, fmt.Errorf("no KSM pod found with label %q in namespace %q: %w", p.ksmPodLabel, p.ksmNamespace, errNoKSMPodsFound)
```
roobre, line 28: Wrapping will make the error message look like:

`no KSM pod found with label %q in namespace %q: no KSM pods found`

Which seems redundant.

What about ...?

```suggestion
		return nil, fmt.Errorf("discovering KSM with label %q in namespace %q: %w", p.ksmPodLabel, p.ksmNamespace, errNoKSMPodsFound)
```

src/ksm/client/distributed_pod_label_discovery.go
```diff
@@ -36,7 +35,7 @@ func (p *distributedPodLabelDiscoverer) findAllLabeledPodsRunningOnNode() ([]v1.
 		}
 
 		if pod.Status.HostIP == p.ownNodeIP {
-			p.logger.Debugf("Found KSM pod running on this code, pod IP: %s", pod.Status.PodIP)
+			p.logger.Debugf("Found KSM pod running on this node, pod IP: %s", pod.Status.PodIP)
 			foundPods = append(foundPods, pod)
 		}
```
roobre, line 40: Maybe a bit unrelated, but do you think it would be worth logging (`Debugf`) the pods that do not match this condition? Could be helpful for debugging.

src/ksm/client/pod_label_discovery.go
```diff
@@ -6,29 +6,28 @@ import (
...
-
+	pods, err := p.k8sClient.FindPodsByLabel(p.ksmNamespace, fmt.Sprintf("%s=true", p.ksmPodLabel))
 	if err != nil {
-		return nil, errors.Wrap(err, "could not query api server for pods")
+		return nil, fmt.Errorf("querying API server for Pods: %w", err)
 	}
+
 	if len(pods.Items) == 0 {
-		return nil, errors.Wrapf(errNoKSMPodsFound, "no KSM pod found with label: '%s'", p.ksmPodLabel)
+		return nil, fmt.Errorf("no KSM pods found with label %q in namespace %q: %w", p.ksmPodLabel, p.ksmNamespace, errNoKSMPodsFound)
```
roobre, line 30: Same as above, could this be reworded to avoid redundancy?

roobre (commented):
Left some nits and suggestions regarding where to put the namespace.

I also see that we got rid of `FindPodByName` and `FindPodsByHostname`. Wheren't those being used anywhere?

src/client/client.go
```diff
@@ -23,11 +23,7 @@ type Kubernetes interface {
 	// FindNode returns a Node reference containing the pod named as the argument, if any
 	FindNode(name string) (*v1.Node, error)
 	// FindPodsByLabel returns a PodList reference containing the pods matching the provided name/value label pair
-	FindPodsByLabel(name, value string) (*v1.PodList, error)
-	// FindPodByName returns a PodList reference that should contain the pod whose name matches with the name argument
-	FindPodByName(name string) (*v1.PodList, error)
-	// FindPodsByHostname returns a Podlist reference containing the pod or pods whose hostname matches the argument
-	FindPodsByHostname(hostname string) (*v1.PodList, error)
+	FindPodsByLabel(namespace, labelSelector string) (*v1.PodList, error)
```
roobre, line 26: I think it might be good to put an example in this comment regarding how `labelSelector` is supposed to look, just in case someone unfamiliar with the k8s concept reads this.

src/ksm/client/distributed_pod_label_discovery.go
invidian: Yes... Want me to refactor that before the addition?

src/ksm/client/distributed_pod_label_discovery.go
invidian: Sure.

invidian:
>I also see that we got rid of FindPodByName and FindPodsByHostname. Wheren't those being used anywhere?

Yup, not used anywhere.

src/client/client.go
invidian: I could actually make it strongly typed and use https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#LabelSelector. What do you think?

src/ksm/client/discovery.go
invidian: It's not really optional, `""` just gives you all namespaces and it's a valid value. But yeah, at this point it there could be a config struct for it.

src/ksm/client/distributed_pod_label_discovery.go
invidian: If you query all namespaces, it will be a lot. Maybe we can log a pod selector in namespace X + IP which we look for?

In case all pods on the cluster are queried (this is the default), it will be a lot of output.

src/client/client.go
roobre: I think it would be quite nice. How would the query behave if supplied with the zero value of the `LabelSelector`?

src/ksm/client/discovery.go
roobre: What I mean was more like you can construct the object without it, because all namespaces (`""`) is a very reasonable default.

src/ksm/client/distributed_pod_label_discovery.go
roobre: I was under the impression that this check is evaluated after filtering by the KSM label, isn't this the case?

src/ksm/client/distributed_pod_label_discovery.go
roobre: I think it might be worth a try, if the impact it's not too big. It's up to you :)

roobre:
Btw, would it be possible to base this PR on `main` rather than on the SDK v3 branch? I'd prefer to make a release with the new features and then another one with the sdk change, so we dont stack two major changes in the same release.

invidian:
>Btw, would it be possible to base this PR on main rather than on the SDK v3 branch? I'd prefer to make a release with the new features and then another one with the sdk change, so we dont stack two major changes in the same release.

Sure, but since I touch the CLI code, it will need to be carried over to SDK PR as well.

src/client/client.go
invidian: I'll make sure it selects all pods :)

roobre: Then I'd say let's go with the strong type

src/ksm/client/distributed_pod_label_discovery.go
invidian: Yes, it is. I'll see how it would look like.

src/client/client.go
```diff
@@ -61,21 +57,9 @@ func (ka *goClientImpl) FindNode(name string) (*v1.Node, error) {
...
-	return ka.client.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{
-		FieldSelector: fmt.Sprintf("metadata.name=%s", name),
-	})
-}
-
-func (ka *goClientImpl) FindPodsByHostname(hostname string) (*v1.PodList, error) {
-	return ka.client.CoreV1().Pods("").List(context.TODO(), metav1.ListOptions{
-		FieldSelector: fmt.Sprintf("spec.hostname=%s", hostname),
+func (ka *goClientImpl) FindPodsByLabel(namespace, labelSelector string) (*v1.PodList, error) {
+	return ka.client.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{
```
roobre, line 61:
```suggestion
	// nosemgrep: context-todo
	return ka.client.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{
```

invidian:
Added e2e tests for this feature. Now the refactoring part :unamused:

src/ksm/client/distributed_pod_label_discovery.go
invidian: Done in #172

src/ksm/client/discovery.go
invidian: Refactored in #172

src/client/client.go
invidian: Done in #172

invidian:
I'll still re-do the commits and add some extra unit tests.

src/ksm/group.go
```diff
@@ -22,7 +22,7 @@ type ksmGrouper struct {
 // addServiceSpecSelectorToGroup adds a new metric to the service group
 // which includes the selectors defined in the service spec.
 func (r *ksmGrouper) addServiceSpecSelectorToGroup(serviceGroup map[string]definition.RawMetrics) error {
-	services, err := r.k8sClient.ListServices()
+	services, err := r.k8sClient.ListServices("")
```
roobre, line 25: This line makes me wonder if we have any kind of caching set up for calls like this.

e2e/cmd/e2e.go
```diff
@@ -184,6 +192,9 @@ func main() {
 			// not the logic of KSM. This might change if KSM sharding becomes enabled by default.
 			s.TwoKSMInstances = true
 		},
+		"with_namespace_filtering": func(s *scenario.Scenario) {
+			s.ExtraArgs = []string{"-kube_state_metrics_namespace=default"}
```
roobre, line 196: Outside of the context I wouldn't be a fan of `Extra` stuff but I do think it's the lesser evil here. Nice job on that.

roobre (approved):

e2e/cmd/e2e.go
invidian: Can you elaborate on "Extra stuff"? I'm not sure what you mean.

src/ksm/group.go
invidian: I don't think so, but it would be easier if we could implement that once we have a daemon process, so we can cache in memory. Using e.g. informers.

PR #172, opened 2021-07-23
Refactor src/client and src/ksm/client a bit

invidian:
Partially addresses the review comments from #165.

Draft, as I'm still not 100% sure about exposing `LookupSRV` field.

### DIFF
--- a/src/ksm/client/discovery.go
+++ b/src/ksm/client/discovery.go
@@ -13,6 +13,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/transport"
 
 	"github.com/newrelic/nri-kubernetes/v2/src/client"
@@ -148,7 +149,11 @@ func (sd *discoverer) apiDiscover() (url.URL, error) {
 	var err error
 
 	for _, label := range ksmAppLabelNames {
-		services, err = sd.apiClient.FindServicesByLabel(label, ksmAppLabelValue)
+		services, err = sd.apiClient.FindServicesByLabel(metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				label: ksmAppLabelValue,
+			},
+		})
 		if err == nil && len(services.Items) > 0 {
 			break
 		}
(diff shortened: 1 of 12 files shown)


src/ksm/client/discovery.go
```diff
@@ -34,11 +35,42 @@ const (
...
+
+// DiscovererConfig holds parameters for creating discoverer.
+type DiscovererConfig struct {
+	LookupSRV         LookupSRVFunc
+	APIClient         client.Kubernetes
+	Logger            *logrus.Logger
+	OverridenEndpoint string
+}
+
+func (dc DiscovererConfig) New() (client.Discoverer, error) {
```
roobre, line 48: Having a config returning an instance of the object is an interesting concept. Normally I'm more used to do `func NewDiscovery(dc DiscovererConfig) (client.Discoverer, error)`.

Not that I'm against it or anything, but curious why you decided to do it this way :)

invidian: Hmm, I usually do this to avoid passing the parameters to function, as those tend to grow. It's also sometimes useful to make struct the _context_ of function call, while function arguments can act more as a "runtime" dependencies, like context or logger, which often cannot be serialized.

However, this approach has a drawback, as it's not explicit what will be returned from `DiscovererConfig.New()`. One may guess, but `NewDiscovery(dc DiscovererConfig` is more explicit.

I wonder about `DiscovererConfig.NewDiscoverer()` now :thinking:

roobre: Alternatively, just `DiscovererConfig.Discoverer()` could be a way as well.

But I do think that `NewDiscoverer(dc DiscovererConfig)` is perhaps what go developers are more used to see.

invidian: Changed to `NewDiscoverer(dc DiscovererConfig)` then.

src/ksm/client/discovery.go
```diff
@@ -34,11 +35,42 @@ const (
 	headlessServiceClusterIP = "None"
 )
 
-type lookupSRVFunc func(service, proto, name string) (cname string, addrs []*net.SRV, err error)
+type LookupSRVFunc func(service, proto, name string) (cname string, addrs []*net.SRV, err error)
+
+// DiscovererConfig holds parameters for creating discoverer.
+type DiscovererConfig struct {
+	LookupSRV         LookupSRVFunc
+	APIClient         client.Kubernetes
```
roobre, line 43: We call this `K8sClient` in other places, is there a reason for not unifying those?

roobre (approved):
Minor comment on naming, LGTM

src/ksm/client/discovery.go
invidian: Well, `discoverer` struct also has `apiClient`, but let's fix all that then. Pushed a fix :)

PR #178, opened 2021-07-26
e2e: improvements

invidian:
To simplify changes in #165.

### DIFF
--- a/.github/workflows/push_pr.yaml
+++ b/.github/workflows/push_pr.yaml
@@ -50,3 +50,16 @@ jobs:
         continue-on-error: ${{ github.event_name != 'pull_request' }}
         with:
           only-new-issues: true
+
+  codespell:
+    name: Codespell
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v2
+      - name: Codespell test
+        uses: codespell-project/actions-codespell@master
+        with:
+          skip: ./.git
+          check_filenames: true
+          check_hidden: true
--- a/e2e/charts/newrelic-infrastructure-k8s-e2e/templates/daemonset.yaml
+++ b/e2e/charts/newrelic-infrastructure-k8s-e2e/templates/daemonset.yaml
@@ -26,6 +26,8 @@ spec:
         - name: {{ .Values.daemonset.name }}
           image: "{{ .Values.daemonset.image.repository }}:{{ .Values.daemonset.image.tag }}"
           {{- if eq .Values.daemonset.clusterFlavor "Minikube" }}
+          # On minikube, agent image should be built using 'eval $(minikube -p minikube docker-env)',
+          # so it never has to be pulled from the registry.
           imagePullPolicy: Never
           {{- else }}
           imagePullPolicy: Always
@@ -53,6 +55,7 @@ spec:
               name: host-docker-socket
             {{- end }}
           env:
+            # On Minikube, also test scraping the controlplane via localhost.
             {{- if eq .Values.daemonset.clusterFlavor "Minikube" }}
               {{- if semverCompare ">=1.16-0" .Capabilities.KubeVersion.Version }}
             - name: "SCHEDULER_ENDPOINT_URL"
--- a/e2e/charts/newrelic-infrastructure-k8s-e2e/templates/daemonset.yaml
+++ b/e2e/charts/newrelic-infrastructure-k8s-e2e/templates/daemonset.yaml
@@ -54,12 +54,10 @@ spec:
             {{- end }}
           env:
             {{- if eq .Values.daemonset.clusterFlavor "Minikube" }}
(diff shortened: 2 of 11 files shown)


.github/workflows/push_pr.yaml
```diff
@@ -50,3 +50,16 @@ jobs:
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v2
+      - name: Codespell test
+        uses: codespell-project/actions-codespell@master
+        with:
+          skip: ./.git
+          check_filenames: true
+          check_hidden: true
```
roobre, line 65: Sneaky change :P

e2e/charts/newrelic-infrastructure-k8s-e2e/templates/daemonset.yaml
```diff
@@ -54,12 +54,10 @@ spec:
             {{- end }}
           env:
             {{- if eq .Values.daemonset.clusterFlavor "Minikube" }}
-              {{- if and (ne .Values.daemonset.clusterVersion "v1.15.x") (ne .Values.daemonset.clusterVersion "v1.14.x") (ne .Values.daemonset.clusterVersion "v1.13.x") }}
+              {{- if semverCompare ">=1.16-0" .Capabilities.KubeVersion.GitVersion }}
```
roobre, line 57: What does the `-0` do?

e2e/charts/newrelic-infrastructure-k8s-e2e/templates/daemonset.yaml
```diff
@@ -54,12 +54,10 @@ spec:
             {{- end }}
           env:
             {{- if eq .Values.daemonset.clusterFlavor "Minikube" }}
```
roobre, line 56: I'd take the advantage of your enlightment about what these checks do and add a small comment describing them.

e2e/cmd/e2e.go
```diff
@@ -287,117 +275,106 @@ func waitForKSM(c *k8s.Client, logger *logrus.Logger) (*v1.Pod, error) {
...
 	if err != nil {
 		execErr.errs = append(execErr.errs, err)
 	}
 
-	if currentScenario.ClusterFlavor == minikubeFlavor {
+	if se.scenario.ClusterFlavor == minikubeFlavor {
 		// Minikube use hostPath provisioner for PVCs, which makes kubelet to not report PVC volumes in /stats/summary
 		// See https://github.com/yashbhutwala/kubectl-df-pv/issues/2 for more info.
-		logger.Info("Skipping `testSpecificEntities` because you're running them in Minikube.")
+		se.logger.Info("Skipping `testSpecificEntities` because you're running them in Minikube.")
```
roobre, line 336: Again not quite a fan of the `testSpecificEntities` name, I think we should reflect what the comment above says in the log line (and then possibly remove the comment).

e2e/cmd/e2e.go
```diff
@@ -128,19 +86,16 @@ const (
 
 var allJobs = [...]job{jobKSM, jobKubelet, jobScheduler, jobEtcd, jobControllerManager, jobAPIServer}
 
-func execIntegration(pod v1.Pod, ksmPod *v1.Pod, dataChannel chan integrationData, wg *sync.WaitGroup, c *k8s.Client, logger *logrus.Logger) {
-	defer timer.Track(time.Now(), fmt.Sprintf("execIntegration func for pod %s", pod.Name), logger)
-	defer wg.Done()
```
roobre, line 133: Here we got rid of a waitgroup and a channel. What were these doing?

e2e/charts/newrelic-infrastructure-k8s-e2e/templates/daemonset.yaml
invidian: https://github.com/Masterminds/semver#working-with-prerelease-versions IIRC it includes the prereleases.

e2e/charts/newrelic-infrastructure-k8s-e2e/templates/daemonset.yaml
invidian: Sure, added some.

e2e/cmd/e2e.go
invidian: Sure, but perhaps out of scope for this already large PR?

e2e/cmd/e2e.go
invidian: So, integration was executed in parallel on all listed pods. Commit message explains that IIRC.

But as integration runs fast and usually on one pod only, there is little value it seems in executing it asynchronously.

e2e/cmd/e2e.go
roobre: Yup, you're right. This is already pretty big.

roobre (approved):
Looking great, thanks a lot @invidian

PR #179, opened 2021-07-27
metric: add nodeName attribute to node specs

### LINKED ISSUE #173: Add `nodeName` to sample containing node conditions

roobre:
Fixes #173

### DIFF
--- a/src/metric/definition.go
+++ b/src/metric/definition.go
@@ -718,7 +718,9 @@ var KSMSpecs = definition.SpecGroups{
 	"node": {
 		TypeGenerator: kubeletMetric.FromRawGroupsEntityTypeGenerator,
 		Specs: []definition.Spec{
+			// nodeName and Labels should be provided here as well so conditions can be queried and faceted by them
 			{Name: "nodeName", ValueFunc: prometheus.FromLabelValue("kube_node_info", "node"), Type: sdkMetric.ATTRIBUTE},
+			{Name: "label.*", ValueFunc: prometheus.InheritAllLabelsFrom("node", "kube_node_labels"), Type: sdkMetric.ATTRIBUTE},
 			{
 				Name: "condition.*",
 				ValueFunc: definition.Transform(
@@ -840,7 +842,9 @@ var KSMQueries = []prometheus.Query{
 	},
 	{MetricName: "kube_hpa_status_current_replicas"},
 	{MetricName: "kube_hpa_status_desired_replicas"},
+	// Node info and labels, so sample containing conditions also has this information
 	{MetricName: "kube_node_info"},
+	{MetricName: "kube_node_labels"},
 	// Node status condition
 	{MetricName: "kube_node_status_condition", Value: prometheus.QueryValue{
 		// Since we aggregate metrics which look like the following:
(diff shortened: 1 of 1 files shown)


invidian (approved):

src/metric/definition.go
```diff
@@ -839,6 +840,7 @@ var KSMQueries = []prometheus.Query{
 	},
 	{MetricName: "kube_hpa_status_current_replicas"},
 	{MetricName: "kube_hpa_status_desired_replicas"},
+	{MetricName: "kube_node_info"},
```
invidian, line 843: Should we have some tests for that?

invidian (approved):

src/metric/definition.go
roobre: Good point, I'm not sure where those would fit with the current architecture, since I haven't seen any obvious place for them. Do you have any suggestion?

src/metric/definition.go
```diff
@@ -718,7 +718,9 @@ var KSMSpecs = definition.SpecGroups{
 	"node": {
 		TypeGenerator: kubeletMetric.FromRawGroupsEntityTypeGenerator,
 		Specs: []definition.Spec{
+			// nodeName and Labels should be provided here as well so conditions can be queried and faceted by them
```
invidian, line 721:
```suggestion
			// nodeName and Labels should be provided here as well so conditions can be queried and faceted by them.
```

src/metric/definition.go
```diff
@@ -840,7 +842,9 @@ var KSMQueries = []prometheus.Query{
 	},
 	{MetricName: "kube_hpa_status_current_replicas"},
 	{MetricName: "kube_hpa_status_desired_replicas"},
+	// Node info and labels, so sample containing conditions also has this information
```
invidian, line 845:
```suggestion
	// Node info and labels, so sample containing conditions also has this information.
```

roobre:
Following the same logic, I've also added node labels, sourced from KSM as well.

src/metric/definition.go
```diff
@@ -718,6 +718,9 @@ var KSMSpecs = definition.SpecGroups{
 	"node": {
 		TypeGenerator: kubeletMetric.FromRawGroupsEntityTypeGenerator,
 		Specs: []definition.Spec{
+			// nodeName and Labels should be provided here as well so conditions can be queried and faceted by them.
+			{Name: "nodeName", ValueFunc: prometheus.FromLabelValue("kube_node_info", "node"), Type: sdkMetric.ATTRIBUTE},
+			{Name: "label.*", ValueFunc: prometheus.InheritAllLabelsFrom("node", "kube_node_labels"), Type: sdkMetric.ATTRIBUTE},
```
roobre, line 723: Just noticed that these two funcs have their arguments in opposite positions and now I won't be able to sleep at night.

src/metric/definition.go
invidian: Given the simplicity of the code and complexity of adding tests for this, I'm fine leaving it out for now :smile: 

The question was more of a bait rather than a requirement :smile:

PR #180, opened 2021-07-27
kubernetes-static: log with logrus rather than fmt.Println

roobre:
The former makes stdout not json-conformant and cannot be piped to jq, which annoys me to ontological levels

### DIFF
--- a/cmd/kubernetes-static/main.go
+++ b/cmd/kubernetes-static/main.go
@@ -225,7 +225,7 @@ func startStaticMetricsServer(content embed.FS, k8sMetricsVersion string) string
 	}
 
 	endpoint := fmt.Sprintf("http://%s", listener.Addr())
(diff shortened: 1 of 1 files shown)


cmd/kubernetes-static/main.go
```diff
@@ -225,7 +225,7 @@ func startStaticMetricsServer(content embed.FS, k8sMetricsVersion string) string
 	}
 
 	endpoint := fmt.Sprintf("http://%s", listener.Addr())
-	fmt.Println("Hosting Mock Metrics data on:", endpoint)
+	logrus.Infof("Hosting Mock Metrics data on: %s", endpoint)
```
invidian, line 228: Maybe
```suggestion
	logrus.Infof("Hosting Mock Metrics data on %s", endpoint)
```

invidian (approved):

PR #182, opened 2021-08-02
ci: use short commit hash as default TAG

gsanchezgavier:


roobre:
Looks great! My only concern with this is that we won't be able to know whether the integation was compiled in the nightly pipeline.

Do you see any downside to putting this in the workflow file instead?
Changing
https://github.com/newrelic/nri-kubernetes/blob/03c82f3a93e4fd053f22bb4e7718025327a77620/.github/workflows/nightly.yml#L29
To
`make TAG=nightly-$(git rev-parse --short HEAD)`

roobre (approved):
Awesome, thanks @gsanchezgavier !

PR #185, opened 2021-08-05
Add TTL jitter support to cache expiry

### LINKED ISSUE #177: Add jitter to discovery cache TTL

Right now, nri-kubernetes caches the output of the discovery operations for a certain period of time. However, this TTL is fixed for all instances of the DaemonSet. For clusters with a significant number of nodes, this causes caches to deterministically expire at the same time, which creates bursts of load in the control plane.

## Feature Description

We'd like to add some randomness to the expiration of this caches, so it is unlikely that all requests happen at the same time.

As a nice to have, it would be nice to be able to configure:
- The TTL itself
- The max amount of random deviation from the target TTL

invidian:
See commit messages for more details.

Closes #177

### DIFF
--- a/src/client/cached.go
+++ b/src/client/cached.go
@@ -41,7 +41,7 @@ func (d *DiscoveryCacher) Discover(timeout time.Duration) (HTTPClient, error) {
 	if err == nil {
 		d.Logger.Debugf("Found cached copy of %q stored at %s", d.StorageKey, time.Unix(creationTimestamp, 0))
 		// Check cached object TTL
-		if time.Now().Unix() < creationTimestamp+int64(d.TTL.Seconds()) {
+		if !expired(creationTimestamp, d.TTL) {
 			wrappedClient, err := d.Compose(d.CachedDataPtr, d, timeout)
 			if err != nil {
 				return nil, err
@@ -73,6 +73,10 @@ func (d *DiscoveryCacher) discoverAndCache(timeout time.Duration) (HTTPClient, e
 		d.Logger.WithError(err).Warnf("while storing %q in the cache", d.StorageKey)
 	}
 	return client, nil
+}
+
+func expired(creationTimestamp int64, ttl time.Duration) bool {
+	return time.Since(time.Unix(creationTimestamp, 0)) > ttl
 }
 
 func (d *DiscoveryCacher) wrap(client HTTPClient, timeout time.Duration) *cacheAwareClient {
@@ -143,7 +147,7 @@ func (d *MultiDiscoveryCacher) Discover(timeout time.Duration) ([]HTTPClient, er
 	if err == nil {
 		d.Logger.Debugf("Found cached copy of %q stored at %s", d.StorageKey, time.Unix(creationTimestamp, 0))
 		// Check cached object TTL
-		if time.Now().Unix() < creationTimestamp+int64(d.TTL.Seconds()) {
+		if !expired(creationTimestamp, d.TTL) {
 			clients, err := d.Compose(d.CachedDataPtr, d, timeout)
 			if err != nil {
 				return nil, errors.Wrap(err, "could not compose cache")
--- a/src/client/cached.go
+++ b/src/client/cached.go
@@ -37,11 +37,11 @@ type Composer func(source interface{}, cacher *DiscoveryCacher, timeout time.Dur
 // Discover tries to retrieve a HTTPClient from the cache, and otherwise engage the discovery process from the wrapped
 // Discoverer
 func (d *DiscoveryCacher) Discover(timeout time.Duration) (HTTPClient, error) {
-	ts, err := d.Storage.Read(d.StorageKey, d.CachedDataPtr)
+	creationTimestamp, err := d.Storage.Read(d.StorageKey, d.CachedDataPtr)
 	if err == nil {
-		d.Logger.Debugf("Found cached copy of %q stored at %s", d.StorageKey, time.Unix(ts, 0))
+		d.Logger.Debugf("Found cached copy of %q stored at %s", d.StorageKey, time.Unix(creationTimestamp, 0))
 		// Check cached object TTL
-		if time.Now().Unix() < ts+int64(d.TTL.Seconds()) {
+		if time.Now().Unix() < creationTimestamp+int64(d.TTL.Seconds()) {
 			wrappedClient, err := d.Compose(d.CachedDataPtr, d, timeout)
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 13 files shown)


src/client/cached.go
```diff
@@ -75,6 +75,12 @@ func (d *DiscoveryCacher) discoverAndCache(timeout time.Duration) (HTTPClient, e
 	return client, nil
 }
 
+func expired(creationTimestamp int64, ttl time.Duration) bool {
+	expirationTime := time.Unix(creationTimestamp, 0).Add(ttl)
+
+	return time.Now().After(expirationTime)
+}
```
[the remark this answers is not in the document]
invidian: I think `return time.Since(time.Unix(creationTimestamp, 0)) > ttl` is correct then: cache **expired** when time **since** **creation** is bigger than **TTL**.

src/client/cached_test.go
```diff
@@ -163,6 +108,155 @@ func TestCacheAwareClient_RediscoveryDoesntWork(t *testing.T) {
...
+		assert.Equal(t, 200, resp.StatusCode)
+	})
+
+	t.Run("does_not_perform_the_discovery", func(t *testing.T) {
+		// The Discovery process has been triggered once
+		discoverer.AssertExpectations(t)
+	})
+}
+
+func Test_CacheAwareClient_perform_the_discovery_again_when_cache_TTL_is_reached(t *testing.T) {
```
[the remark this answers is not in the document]
invidian: This seems a bit out of scope to me.

src/client/cached.go
```diff
@@ -139,11 +145,11 @@ type MultiDiscoveryCacher struct {
 // If the cache is not present or has expired, it will be written.
 // If the cache read fails, the underlying discovery will still run.
 func (d *MultiDiscoveryCacher) Discover(timeout time.Duration) ([]HTTPClient, error) {
```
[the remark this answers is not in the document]
invidian: Maybe, I'll have a look.

src/client/cached.go
```diff
@@ -37,11 +37,11 @@ type Composer func(source interface{}, cacher *DiscoveryCacher, timeout time.Dur
...
 func (d *DiscoveryCacher) Discover(timeout time.Duration) (HTTPClient, error) {
-	ts, err := d.Storage.Read(d.StorageKey, d.CachedDataPtr)
+	creationTimestamp, err := d.Storage.Read(d.StorageKey, d.CachedDataPtr)
 	if err == nil {
-		d.Logger.Debugf("Found cached copy of %q stored at %s", d.StorageKey, time.Unix(ts, 0))
+		d.Logger.Debugf("Found cached copy of %q stored at %s", d.StorageKey, time.Unix(creationTimestamp, 0))
 		// Check cached object TTL
-		if time.Now().Unix() < ts+int64(d.TTL.Seconds()) {
+		if !expired(creationTimestamp, d.TTL) {
 			wrappedClient, err := d.Compose(d.CachedDataPtr, d, timeout)
```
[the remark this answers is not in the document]
invidian: This is sort of implied by:
`d.Logger.Debugf("Found cached copy of %q stored at %s", d.StorageKey, time.Unix(creationTimestamp, 0))`
and 
`d.Logger.Debugf("Cached copy of %q expired. Refreshing", d.StorageKey)`
already.

src/client/cached.go
```diff
@@ -72,7 +74,11 @@ func (d *DiscoveryCacher) discoverAndCache(timeout time.Duration) (HTTPClient, e
 	if err != nil {
 		d.Logger.WithError(err).Warnf("while storing %q in the cache", d.StorageKey)
 	}
-	return client, nil
+	return d.wrap(client, timeout), nil
+}
+
+func expired(creationTimestamp int64, ttl time.Duration) bool {
+	return time.Since(time.Unix(creationTimestamp, 0)) > ttl
 }
```
roobre, line 82: Given that this is now a oneliner, is it worth to have a function wrapper for it?

invidian: I think it still provides some value with code dedup + naming it. I'd leave it.

src/client/cached.go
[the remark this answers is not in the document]
invidian: So, I don't think there is an easy way to compose that. Maybe a little bit, but that doesn't help much. I don't think it's worth spending much time either.

invidian:
I think it should be roughly working and ready for some reviews now. I didn't do a long term testing of that though.

src/client/cached.go
```diff
@@ -72,7 +82,24 @@ func (d *DiscoveryCacher) discoverAndCache(timeout time.Duration) (HTTPClient, e
...
+func Expired(currentTime time.Time, creationTimestamp int64, ttl time.Duration, jitterMaxPercentage uint) bool {
+	rand.Seed(time.Now().UTC().UnixNano())
+
+	selectedJitter := rand.Intn(int(jitterMaxPercentage)+1) - 1
+
+	if i := rand.Intn(2); i == 1 {
+		selectedJitter = selectedJitter * -1.0
+	}
+
+	jitterPercentage := (100 + float64(selectedJitter)) / 100
```
roobre, line 97: What about getting `rand.Float64`, substract `0.5` from it, and then multiply that by `float64(jitterMaxPercentage)/100` ?

roobre (approved):

src/client/cached.go
invidian: I decided to use `rand.Intn` to avoid using floating point as long as possible, which I think makes debugging simpler.

I'll actually fix `-1.0` to `-1`.

invidian: So `jitterPercentage := (100 + float64(selectedJitter)) / 100` produces a float as it does the division, then next line converts it back to time.Duration.

roobre: I would be more concerned about readability than debuggability. Are you sure it's better to have 4 lines of arithmetic an and if, as opposed to two?

invidian: >What about getting (rand.Float64() * 2) - 1, which would be a random number between -1 and 1, and then multiply that by float64(jitterMaxPercentage)/100 ?

With that, we still have to do an operation on the TTL, either addition or multiplication by factor.

Given that `(rand.Float64() * 2) - 1` gives us let's say `-0.5`.

Then `(jitterMaxPercentage/100)*-0.5` gives `−0.1`, with `jitterMaxPercentage = 20`.

How can we apply it to original TTL then?

invidian: Okay, it can be done this way:
```go
  // Convert e.g. 20% to 0.2
  jitterPercentage := float64(jitterMaxPercentage) / 100

  // Random number between -1 and 1.
  randomFactor := ((rand.Float64() * 2) - 1)

  // TTL multiplier must be a positive value for jitterMaxPercentage < 100, so add 1 to ensure that.
  //
  // With jitterMaxPercentage < 100:
  //
  // 0.2 * -0.75 = -0.15
  // -0.15 + 1 = 0.85
  //
  // With jitterMaxPercentage > 100:
  //
  // 1.2 * -1.5 = -1.8
  // -1.8 + 1 = -0.8
  //
  ttlMultiplier := (jitterPercentage * randomFactor) + 1

  // Multiply TTL as float with multiplier, then convert back to duration in seconds.
  ttlWithJitter := time.Duration(ttl.Seconds()*ttlMultiplier) * time.Second

  // As in documentation, time.Now().Sub() is the same as time.Since().
  return currentTime.Sub(time.Unix(creationTimestamp, 0)) > ttlWithJitter
```

invidian:
Ping @roobre

roobre (approved):
I thought I have approved this earlier, but it seems that I didn't :(
Sorry about that, LGTM and thanks a lot for the effort put here!

PR #189, opened 2021-08-09
Add Kubernetes v1.22.0 support

invidian:


.github/workflows/e2e.yaml
```diff
@@ -13,9 +13,9 @@ jobs:
       max-parallel: 7 # len(k8sVersion) is a good number to have here
       matrix:
         privilegedMode: [ "normal", "unprivileged" ]
-        # Latest patch version can be found in https://github.com/kubernetes/sig-release/blob/master/releases/patch-releases.md
+        # Latest patch version can be found in https://github.com/kubernetes/website/blob/main/content/en/releases/patch-releases.md
         # Some versions might not be available yet in https://storage.googleapis.com/kubernetes-release/release/v1.X.Y/bin/linux/amd64/kubelet
-        k8sVersion: [ "v1.15.12", "v1.16.15", "v1.17.17", "v1.18.18", "v1.19.10", "v1.20.6", "v1.21.0" ]
+        k8sVersion: [ "v1.15.12", "v1.16.15", "v1.17.17", "v1.18.18", "v1.19.10", "v1.20.6", "v1.21.0", "v1.22.0" ]
```
roobre, line 18: Can we bump patch for 1.21 as well?

invidian: Sure. Do we want to drop testing for v1.15 as well?

roobre: Definitely, let's get rid of it.

e2e/cmd/e2e.go
```diff
@@ -186,6 +186,7 @@ func main() {
 		"latest_single_instance": func(s *scenario.Scenario) {},
 		"latest_but_one_single_instance": func(s *scenario.Scenario) {
 			s.KSMVersion = "v1.9.7"
```
gsanchezgavier, line 188: shouldn't be 1.9.8?

gsanchezgavier (approved):
lgtm

e2e/cmd/e2e.go
invidian: No, as it tests last but one KSM version.

PR #194, opened 2021-08-20
Fetch node status conditions and unschedulable state from API server rather than KSM

roobre:
Replaces implementation in #155.

Previous implementation fetched this information from KSM metrics. This caused those metrics to be reported for one node by a different one on multi-node clusters, since there is only one instance of the DaemonSet fetching data for all the cluster objects. In turn, this caused some attributes to have unexpected values, such as `hostname` not matching `nodeName`.

This implementation fetches this information from the local node, as an additional step to parsing data from the kubelet.

Additionally, the `unschedulable` metric is also fetched this way.


src/kubelet/group.go
```diff
@@ -82,6 +83,32 @@ func (r *kubelet) Group(definition.SpecGroups) (definition.RawGroups, *data.Erro
...
+		case v1.ConditionFalse:
+			conditionValue = 0
+		case v1.ConditionUnknown:
+			conditionValue = -1
+		default:
+			// Should be unreachable as any other value is not allowed by the API. But if it were, we skip it.
+			continue
+		}
+
+		// Since conditions is a list, there could be duplicate conditions. Check ff we have added this condition before
```
invidian, line 103: Can the condition really be duplicate? I was not able to observe that in my tests, though I couldn't find code preventing it either.

src/apiserver/client.go
```diff
@@ -40,19 +40,23 @@ func (c clientImpl) GetNodeInfo(nodeName string) (*NodeInfo, error) {
...
-		NodeName:    node.ObjectMeta.Name,
-		Labels:      node.Labels,
-		Allocatable: node.Status.Allocatable,
-		Capacity:    node.Status.Capacity,
+		NodeName:      node.ObjectMeta.Name,
+		Labels:        node.Labels,
+		Allocatable:   node.Status.Allocatable,
+		Capacity:      node.Status.Capacity,
+		Conditions:    node.Status.Conditions,
+		Unschedulable: node.Spec.Unschedulable,
```
invidian, line 48: Commit message talks about conditions, but we also start fetching unschedulable here.

src/kubelet/group.go
```diff
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
```
invidian, line 7: This should be imported as `corev1`.

src/kubelet/group.go
```diff
@@ -82,6 +83,32 @@ func (r *kubelet) Group(definition.SpecGroups) (definition.RawGroups, *data.Erro
 		}
 	}
 
+	// Convert node conditions to a map so our ValueFuncs can work nicely with them
```
invidian, line 86: Dots could be consistent.
```suggestion
	// Convert node conditions to a map so our ValueFuncs can work nicely with them.
```

src/apiserver/client.go
```diff
@@ -40,19 +40,23 @@ func (c clientImpl) GetNodeInfo(nodeName string) (*NodeInfo, error) {
 	}
 
 	return &NodeInfo{
-		NodeName:    node.ObjectMeta.Name,
```
invidian, line 43: There is a typo in commit message: `kubelete` :fr:

src/kubelet/metric/pods.go
```diff
@@ -358,17 +358,28 @@ func replicasetNameToDeploymentName(rsName string) string {
...
-	labels, ok := rawLabels.(map[string]string)
-	if !ok {
-		return rawLabels, errors.New("error on creating kubelet label metrics")
-	}
+	return MapToPrefix("label.")(rawLabels)
+}
 
-	modified := make(definition.FetchedValues, len(labels))
-	for k, v := range labels {
-		modified[fmt.Sprintf("label.%v", k)] = v
```
invidian: Seems this refactoring is not related to this commit, could be done in separate commit.

src/kubelet/group_test.go
```diff
@@ -7,12 +7,13 @@ import (
...
 	"testing"
 
-	"github.com/newrelic/nri-kubernetes/v2/src/kubelet/metric"
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 
+	"github.com/newrelic/nri-kubernetes/v2/src/kubelet/metric"
+
```
invidian, line 16: Not needed blank line here.

src/kubelet/group_test.go
```diff
@@ -94,6 +95,8 @@ func TestGroup(t *testing.T) {
 				v1.ResourceEphemeralStorage: *resource.NewQuantity(18211586048, resource.BinarySI),
 				v1.ResourceMemory:           *resource.NewQuantity(2033283072, resource.BinarySI),
 			},
+			Conditions:    nil,
```
invidian: This should be committed together with code adding fields?

src/kubelet/metric/testdata/kubelet_group_expected.go
```diff
@@ -156,10 +157,10 @@ var ExpectedGroupData = definition.RawGroups{
...
 			"nodeIP":           "192.168.99.100",
-			//"restartCount": int32(7), // No restartCount since there is no restartCount in status field in the pod fetched from kubelet /pods.
-			//"isReady":              false, // No isReady since there is no isReady in status field in the pod fetched from kubelet /pods.
-			//"status":         "Running", // No Status since there is no ContainerStatuses field in the pod fetched from kubelet /pods.
-			//"startedAt":            parseTime("2018-02-27T15:21:37Z"), // No startedAt since there is no startedAt in status field in the pod fetched from kubelet /pods.
+			// "restartCount": int32(7), // No restartCount since there is no restartCount in status field in the pod fetched from kubelet /pods.
+			// "isReady":              false, // No isReady since there is no isReady in status field in the pod fetched from kubelet /pods.
+			// "status":         "Running", // No Status since there is no ContainerStatuses field in the pod fetched from kubelet /pods.
+			// "startedAt":            parseTime("2018-02-27T15:21:37Z"), // No startedAt since there is no startedAt in status field in the pod fetched from kubelet /pods.
```
invidian, line 163: Maybe we can align the values and comments like the formatter would do if we touch the formatting here already?

src/kubelet/metric/testdata/kubelet_group_expected.go
```diff
@@ -188,10 +189,10 @@ var ExpectedGroupData = definition.RawGroups{
 			"podName":          "kube-state-metrics-57f4659995-6n2qq",
 			"nodeName":         "minikube",
 			"nodeIP":           "192.168.99.100",
-			//"restartCount": int32(7), // No restartCount since there is no restartCount in status field in the pod fetched from kubelet /pods.
-			//"isReady":              false, // No isReady since there is no isReady in status field in the pod fetched from kubelet /pods.
-			//"status":         "Running", // No Status since there is no ContainerStatuses field in the pod fetched from kubelet /pods.
-			//"startedAt":            parseTime("2018-02-27T15:21:37Z"), // No startedAt since there is no startedAt in status field in the pod fetched from kubelet /pods.
+			// "restartCount": int32(7), // No restartCount since there is no restartCount in status field in the pod fetched from kubelet /pods.
```
invidian, line 192: Ditto formatting.

src/kubelet/metric/transform.go
```diff
@@ -0,0 +1,54 @@
...
+func OneMetricPerLabel(rawLabels definition.FetchedValue) (definition.FetchedValue, error) {
+	return PrefixFromMapString("label.")(rawLabels)
+}
+
+// PrefixFromMapString gets a value assumed to be a map and converts it to a set of values given a prefix, which should
+// includethe separating dot (`.`).
+// For example, calling PrefixFromMapString("prfx.") on a map containing {"foo": "bar", "boo": "baz"} will cause
+// { "prfx.foo": "bar", "prfx.boo": "baz"}
+// to be added to the group.
+func PrefixFromMapString(prefix string) func(mapValue definition.FetchedValue) (definition.FetchedValue, error) {
```
invidian, line 21: Can't this be casted to `map[string]interface{}` maybe, so the function can be de-duplicated with `PrefixFromMapInt`? Or maybe `switch` case could be used somehow?

src/kubelet/metric/transform.go
```diff
@@ -0,0 +1,54 @@
...
+	"fmt"
+
+	"github.com/newrelic/nri-kubernetes/v2/src/definition"
+)
+
+// OneMetricPerLabel transforms a map of labels to FetchedValues type,
+// which will be converted later to one metric per label.
+// It also prefix the labels with 'label.'
+func OneMetricPerLabel(rawLabels definition.FetchedValue) (definition.FetchedValue, error) {
+	return PrefixFromMapString("label.")(rawLabels)
```
invidian, line 13: Now that `PrefixFromMapString` is not used directly anywhere, what's the point of refactoring it?

Also, it's a good practice to move code in one commit, then change it in another, so the changes are nicely highlighted. Right now,   `MapToPrefix` -> `PrefixFromMapString` rename is hidden.

src/kubelet/group_test.go
```diff
@@ -95,7 +95,28 @@ func TestGroup(t *testing.T) {
 				v1.ResourceEphemeralStorage: *resource.NewQuantity(18211586048, resource.BinarySI),
 				v1.ResourceMemory:           *resource.NewQuantity(2033283072, resource.BinarySI),
 			},
-			Conditions:    nil,
+			Conditions: []v1.NodeCondition{
```
invidian, line 98: Again, could be added together with original code, which adds `Conditions: nil`, creating extra context for the reviewer.

invidian (commented):
Code looks mostly good, except some nits like refactoring, imports, dots and typos.

I also left some suggestions how commits could be improved to optimize for nice git history and for reviewing.

src/kubelet/group.go
roobre: I believe the API should disallow that, but took the defensive programming approach here and since the input is a list, there should be possible for that to happen.

src/apiserver/client.go
roobre: Nice point, I'll add it to the description.

src/apiserver/client.go
roobre: I saw it but was too lazy to re-do it 😊

src/kubelet/group.go
roobre: I could change it but it is imported as `v1` across the whole codebase :(
image

src/kubelet/metric/transform.go
roobre: Agreed, I missed out that opportunity. I was a bit annoyed by the whole `interface{}` thing tricking me I kind of rushed the commits.

src/kubelet/metric/testdata/kubelet_group_expected.go
roobre: Done!

src/kubelet/group.go
gsanchezgavier: ~would make sense to log when this happen?~ oh I see no log here

gsanchezgavier (approved):
lgtm

src/kubelet/group.go
invidian: It's consistent, so not ideal, but good enough :+1:

PR #207, opened 2021-09-06
no-merge: cherry-pick #194 on top of 1.21

roobre:
# Do not merge

This branch contains #194 on top of `main` prior to merging #189. We can release this as 2.7.x if we want to delay the 1.22 release, but want to include this fix asap.


roobre:
Superseded by #209

PR #208, opened 2021-09-15
feat: add kubeletVersion to the K8sNodeSample data

danielpanzella:
- Added kubeletVersion to the node data
- updated readme.md and skaffold.yaml as these values seemed necessary to build and deploy locally

Please let me know if there is anything else that needs to be updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ For in cluster development process [Minikube](https://kubernetes.io/docs/setup/l
 
 ### Run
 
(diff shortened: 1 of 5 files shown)


roobre:
This looks great, thanks for your efforts!

Overall seems good to me, I would only ask to add this new metric to the jsonschemas we use for e2e testing, likely [here](https://github.com/newrelic/nri-kubernetes/blob/main/e2e/schema/node-no-network.json) and [here](https://github.com/newrelic/nri-kubernetes/blob/main/e2e/schema/node.json). Otherwise we won't be able to identify regressions or incompatbilities.

dennisme:
Nice!

roobre (approved):

PR #248, opened 2021-10-26
HTTP Sink

paologallinaharbur:
This PR introduces a envVar to toggle the HTTP sink.

The new client can be used as 
```go
	iOptions := []integration.Option{integration.Args(&args)}

        c := pester.New()
        
        h, err := sink.NewHTTPSink(context.Background(), c, sink.DefaultAgentForwarderEndpoint, sink.DefaultTimeout)
        if err != nil {
	        log.Fatal(fmt.Errorf("creating HTTPSink: %w", err))
        }
        
        iOptions = append(iOptions, integration.Writer(h))

	i, err := integration.New(integrationName, integrationVersion, iOptions...)
	if err != nil {
		log.Fatal(fmt.Errorf("creating integration: %w", err))
	}

	logger := log.NewStdErr(args.Verbose)
```

Initially the PR was modifying the current main adding e2e as well, later on we decided to move to new main and e2e tests


src/sink/http.go
```diff
@@ -0,0 +1,117 @@
...
+
+	resp, err := h.client.Do(request)
+	if err != nil {
+		return 0, fmt.Errorf("performing HTTP request: %w", err)
+	}
+
+	defer cleanBody(resp)
+
+	if resp.StatusCode != http.StatusNoContent {
+		return 0, fmt.Errorf("unexpected statuscode: %d, expected: %d", resp.StatusCode, http.StatusNoContent)
```
gsanchezgavier, line 90:
```suggestion
		return 0, fmt.Errorf("unexpected status code: %d, expected: %d", resp.StatusCode, http.StatusNoContent)
```

src/sink/http.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	return len(p), nil
+}
+
+// DefaultPesterClient return a defaultPesterClient to be used with httpSink.
+func DefaultPesterClient(reqTimeout time.Duration) *pester.Client {
```
gsanchezgavier, line 97: Wouldn't make sense to move it outside the package for others to use or embedded it into this one without exporting it?

gsanchezgavier (approved):
looks good!

src/sink/http.go
paologallinaharbur: actually that function does not make any sense in such package, I'll remove it and keep it as a testHelper only

PR #263, opened 2021-11-12
Integration test and static main improvements

roobre:
This PR moves the test data from the `cmd/kubernetes-static` folder to an `internal/testutil` one, and adds some goodies on top. This new `testutil` package contains:

- A server based on `httptest.Server` which exposes all the required metrics from services for a particular version
- A helper to create a fake kubernetes client and add some data to it.

With this, a `WIP` test for the new `ksm.Scraper` has been added as well.


internal/testutil/testutil.go
```diff
@@ -0,0 +1,42 @@
...
+// - UPDATE LatestVersion() BELOW IF NEEDED
+
+const (
+	Testdata116 = "1_16"
+	Testdata118 = "1_18"
+)
+
+// LatestVersion returns the latest version we have test data for.
+func LatestVersion() Version {
+	return Testdata118
```
paologallinaharbur, line 33: Would make sense instead of returning a static field we need to update to order Version with sort and return the last element?

https://play.golang.org/p/Hy2x-W0eBT0

internal/testutil/testutil.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// AllVersions returns a list of versions we have test data for.
+func AllVersions() []Version {
+	return []Version{
+		Testdata116,
```
paologallinaharbur, line 39: should we add a comment saying to which KSM version corresponds to each of these?

internal/testutil/k8s.go
```diff
@@ -0,0 +1,64 @@
...
+			return nil, fmt.Errorf("creating service %q: %w", svc.Name, err)
+		}
+	}
+
+	// TODO: Populate more objects as needed:
+
+	return fakeCs, nil
+}
+
+func k8sCreateNamespaceIfMissing(k8s kubernetes.Interface, namespace string) error {
```
paologallinaharbur, line 49: Just curiosity  - In the past the fake client did not check if a ns existed when you were creating resources. Have you added it because they implemented such check or to be "extra clean" with the impl?

internal/testutil/k8s.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	// KSM scraper relies on a list of services to add extra data to KSM metrics
+	for _, svc := range services {
+		err := k8sCreateNamespaceIfMissing(fakeCs, svc.Namespace)
+		if err != nil {
+			return nil, fmt.Errorf("creating ns %q for service %q: %w", svc.Namespace, svc.Name, err)
+		}
+
+		_, err = fakeCs.CoreV1().Services(svc.Namespace).Create(context.Background(), svc, metav1.CreateOptions{})
```
paologallinaharbur, line 38: Instead of creating the objects one by one you can pass them as arg to the NewSimpleClientset function:

```
	client := testclient.NewSimpleClientset(getNamespaces()...,getServices()...)
```

It would simplify a lot this function

internal/testutil/k8s.go
```diff
@@ -0,0 +1,64 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+const servicesFile = "services.yml"
+
+func FakeK8sClient() (*fake.Clientset, error) {
+	fakeCs := fake.NewSimpleClientset()
+
+	k8sServicesYaml, err := testDataDir.ReadFile(filepath.Join(testDataRootDir, servicesFile))
```
paologallinaharbur, line 20: Why do we focus so much on services? Should we extract this to something like "setupServices".

cmd/kubernetes-static/data/1_18/controlplane/controller-manager/metrics
```diff
@@ -1,2448 +0,0 @@
-# HELP ClusterRoleAggregator_adds (Deprecated) Total number of adds handled by workqueue: ClusterRoleAggregator
-# TYPE ClusterRoleAggregator_adds counter
-ClusterRoleAggregator_adds 502
```
paologallinaharbur, line 3: Something weird happened. It shows that 1.18 was deleted and tha 1.15 was moved to be called 1.18 in an other folder.

I fear the files for all the version were the same and git got confused 😓

internal/testutil/assert.go
```diff
@@ -0,0 +1,165 @@
...
+
+func NewAsserter() Asserter {
+	return Asserter{
+		Mutex: &sync.Mutex{},
+	}
+}
+
+// Using returns an asserter that will use the supplied specGroups to assert entities.
+func (a Asserter) Using(groups definition.SpecGroups) Asserter {
+	a.specGroups = groups
```
paologallinaharbur, line 33: does it make sense to run any of the group without calling "Using"? Should we have it as a parameter of the builder?

internal/testutil/assert.go
```diff
@@ -0,0 +1,165 @@
...
+	return a
+}
+
+// On returns an asserter configured to check for existence on the supplied entities.
+func (a Asserter) On(entities []*integration.Entity) Asserter {
```
paologallinaharbur, line 38: same here, without entities the tests is basically no-op

src/ksm/ksm_test.go
```diff
@@ -0,0 +1,63 @@
...
+			})
+
+			i := testutil.NewIntegration(t)
+
+			err = scraper.Run(i)
+			if err != nil {
+				t.Fatalf("running scraper: %v", err)
+			}
+
+			asserter.On(i.Entities).Assert(t)
```
paologallinaharbur, line 60: here for example it is a bit confusing, you are using the asserter that has groups only because you previously set them up, however if also entities were set up you would have overwrote them. 

I personally find it clearer if the tests data is either in the constructor or in the Assert itself. Something like:
```
assert.Equal(t, Base2Bytes(1048577).String(), "1MiB1B")
```

internal/testutil/assert.go
paologallinaharbur: Or of the asserter

src/ksm/ksm_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/newrelic/nri-kubernetes/v2/src/metric"
+)
+
+func TestScraper(t *testing.T) {
+	// TODO(roobre): We should not exclude Optional, pod or hpa metrics. To be tackled in a follow-up PR.
+	asserter := testutil.NewAsserter().
+		Using(metric.KSMSpecs).
+		ExcludingOptional().
+		Excluding("pod").
+		Excluding("hpa")
```
paologallinaharbur, line 22: we are excluding them since in the output we have locally those metrics were not present?

internal/testutil/k8s.go
paologallinaharbur: If possible I would remove completely  FakeK8sClient and just call from the differente tests 	
client := testclient.NewSimpleClientset(getNamespaces()...,getServices()...)

paologallinaharbur (commented):
A nice improvement! 
I would try to reduce as much as possible "hidden" behaviour to make obvious what we are testing and using which data so that in 6 months we are not confused :)

src/ksm/ksm_test.go
roobre: Yes, this is temporary. The static files don't have those metrics. In an upcoming PR this, together with the `ExcludingOptional` will be removed and the static files updated.

internal/testutil/testutil.go
roobre: Not sure about the `sort` part (as I was thinking of the possibility of having "weirder" versions int he future), but you're totally right with using the last element of AllVersions.

internal/testutil/testutil.go
roobre: Yep, this is something I though about. My plan is to overwrite all the static files so all of them use the KSM we support, but in the future we might want to have weirder versions.

roobre: I will add a comment explaining this, and in the future, when I write the script to generate these files, I will keep in midn to log the version used somewhere.

internal/testutil/k8s.go
roobre: It was erroring for me, so I had to use this hack :(
Perhaps it errored about a different thing and overimplemented it.

internal/testutil/k8s.go
roobre: Interesting solution, I will give a try to creating the services like this.

internal/testutil/k8s.go
roobre: I focused on services because it's the only thing that KSM uses. The idea is to progressively add more things as we need them.

roobre: Moreover, in upcoming PRs the idea is to generate files like the `services.yaml` automatically.

cmd/kubernetes-static/data/1_18/controlplane/controller-manager/metrics
roobre: I thing git went bonkers. From what I can tell, I just moved the things and then deleted the old ones.

But in any case these will be deleted once we regenerate the static files, which will hapen in the next PR in this line.

internal/testutil/assert.go
roobre: Doing it in the builder is an option, but I wanted to keep everything through chainable methods. In fact in an previous iteration it _was_ in the `.Assert` method, but then you had to pass it several times. Like this, you can configure the asserter first with the groups (that will likely be constant across the whole test) and then just add the variable things (entities) after the test fans out in the tables and so on. This is also the reason to why the `Asserter` returns a copy of itself with each chainable method.

internal/testutil/assert.go
roobre: You are right, it makes sense for the asserter to fail if there are either no entities or no groups. I will patch that!

internal/testutil/testutil.go
roobre: Done! We still rely on the user for the sorting, as it could lead to weird effects if we start adding strange stuff, like `1_18KSM2_0` or the likes of that.

internal/testutil/k8s.go
roobre: Solved, now we have helpers returning `runtime.Object`s

internal/testutil/k8s.go
roobre: Sovled!

internal/testutil/k8s.go
roobre: Added also a helper for Namespaces, which for now is empty but will be populated in the follow-up PR.

internal/testutil/assert.go
roobre: I have clarified the comments on how the asserter work, as well as making it more explicit in this test since it's pretty simple (for now).

paologallinaharbur (approved):
Thanks for addressing everything!

PR #267, opened 2021-11-15
New config allowing file definition

alvarocabanas:
- Added new config separated by component.
- Allows both file or env variables using Viper.


paologallinaharbur:
Could make sense to support some basic flags common to all integrations? Currently this is the behaviour:
```
Usage of /var/db/newrelic-infra/newrelic-integrations/bin/nri-kubernetes-linux-amd64-scraper:
/ #
```

It would be nice to have at least an --help and a --version

PR #268, opened 2021-11-16
Deprecate old main

paologallinaharbur:
The main of this PR is to create a baseline for the refactor of kubelet and controlPlane components:

 - old main has been deleted, most of functionality have been moved to new main. (Possibility to change default values for cache, TTL, etc have been removed)
 - Instead of creating the unprivileged image the two env var and the greop/user have been moved to the manifest
 - static manifests have been deleted
 - local helm templated have been updated to leverage the new arch [agent+kubletAndControlplane] [forwarder+ksm]
 
 Open Questions, maybe next PR:
 - i.Publish() is not thread safe. 
 
   Therefore I would publish in a unique point for all components. 

 - We want to split setting up components and running them to avoid setting up the instances each iteration
 - I spitted cp scraping and daemonsets in two different containers, they mount the very same empty dir. I would prefer to split them as well as soon as we work on the chart. Moreover some of those empty dir are no longer useful


Dockerfile-scraper
```diff
@@ -0,0 +1,8 @@
+FROM alpine:latest
```
roobre, line 1: By the `-scraper` suffix I understand this is for KSM, right?

Dockerfile-scraper
```diff
@@ -0,0 +1,8 @@
...
+
+ADD --chmod=755 bin/nri-kubernetes-linux-amd64-scraper /var/db/newrelic-infra/newrelic-integrations/bin/
+
+# creating the nri-agent user used only in unprivileged mode
+RUN addgroup -g 2000 nri-agent && adduser -D -u 1000 -G nri-agent nri-agent
+
+ENTRYPOINT ["/var/db/newrelic-infra/newrelic-integrations/bin/nri-kubernetes-linux-amd64-scraper"]
```
roobre, line 8: Should we use `tini` here as well?

cmd/nri-kubernetes/main.go
```diff
@@ -49,59 +48,92 @@ func main() {
...
-		}
+	return ksmScraper, nil
+}
 
-		err = i.Publish()
-		if err != nil {
-			return fmt.Errorf("publishing integration: %w", err)
-		}
+func buildClients() (*clusterClients, error) {
+	k8s, err := client.NewKubernetes(true)
```
roobre, line 123: Will we get rid of the custom Kubernetes interface in this PR, or not yet? 🥺

cmd/nri-kubernetes/main.go
```diff
@@ -49,59 +48,92 @@ func main() {
...
+
+		time.Sleep(c.Interval)
 	}
 }
 
-func buildClients() (*clusterClients, error) {
-	k8s, err := client.NewKubernetes(true)
-	if err != nil {
-		return nil, fmt.Errorf("building kubernetes client: %w", err)
+func runScrapers(c config.Mock, ksmScraper *ksm.Scraper, i *integration.Integration, clients *clusterClients) error {
```
roobre, line 79: I wonder how this structure will end up being after implementing commands and subcommands as @alvarocabanas and @carlossscastro  is doing. But for the moment they look fine!

e2e/charts/newrelic-infrastructure/templates/daemonset.yaml
```diff
@@ -38,28 +38,63 @@ spec:
 {{ toYaml .Values.image.pullSecrets | indent 8 }}
       {{- end }}
       containers:
-        - name: {{ template "newrelic.name" . }}
-          image: {{ template "newrelic.image" . }}
-          imagePullPolicy: "{{ .Values.image.pullPolicy }}"
+        - name: kubelet-scraper
+          image: {{ .Values.scraperImage.repository }}:{{ .Values.scraperImage.tag }}
```
roobre, line 42:
```suggestion
          image: {{ .Values.scraperImage.repository }}:{{ .Values.scraperImage.tag | default .Chart.AppVersion }}
```

roobre: Since all images (except the agent) will come from this repo, it is safe to assume that the default version for all of them could be `Chart.AppVersion`, right?

e2e/charts/newrelic-infrastructure/templates/daemonset.yaml
```diff
@@ -38,28 +38,63 @@ spec:
...
+          imagePullPolicy: {{ .Values.scraperImage.pullPolicy }}
           securityContext:
             {{- if .Values.privileged }}
             privileged: true
             {{- else }}
             runAsUser: {{ .Values.runAsUser | default 1000 }}
+            runAsGroup: {{ .Values.runAsGroup | default 2000 }}
```
roobre, line 49: I think we should avoid hardcoded `default`s in templates and just make it so in the `values.yml`. I think helm users are already used to not remove stuff from the values as that could break the chart, so I generally write charts with that in mind.

Dockerfile-scraper
roobre: Ok, I remembered now, it will be the same image for all scrapers. Why jsut not call it `nri-kubernetes` then? The other one will just be the `forwarder`.

e2e/charts/newrelic-infrastructure/templates/daemonset.yaml
```diff
@@ -38,28 +38,63 @@ spec:
...
+          volumeMounts:
+            {{- if .Values.privileged }}
+            - name: dev
+              mountPath: /dev
+            - name: log
+              mountPath: /var/log
+            - name: host-volume
+              mountPath: /host
+              readOnly: true
+            {{- else }}
```
roobre, line 72: Do we need these for the kubelet?

e2e/charts/newrelic-infrastructure/templates/daemonset.yaml
```diff
@@ -38,28 +38,63 @@ spec:
...
+            - mountPath: /tmp
+              name: tmpfs-tmp
+            - mountPath: /var/cache/nr-kubernetes
+              name: tmpfs-cache
```
roobre, line 76: Could we generate a single volume and use `subPath` instead? Seems cleaner IMO.

e2e/charts/newrelic-infrastructure/templates/daemonset.yaml
```diff
@@ -38,28 +38,63 @@ spec:
...
+            {{- end }}
+          {{- if .Values.resources }}
+          resources:
+{{ toYaml .Values.resources | indent 12 }}
+          {{- end }}
+        {{- if .Values.privileged }}
+        - name: control-plane-scraper
```
roobre, line 83: Why deploy the control plane `if privileged`?

e2e/charts/newrelic-infrastructure/templates/daemonset.yaml
```diff
@@ -102,14 +137,58 @@ spec:
...
+          volumeMounts:
+            {{- if .Values.privileged }}
+            - name: dev
+              mountPath: /dev
+            - name: log
+              mountPath: /var/log
+            - name: host-volume
+              mountPath: /host
+              readOnly: true
+            {{- else }}
```
roobre, line 154: Ditto

e2e/charts/newrelic-infrastructure/templates/daemonset.yaml
```diff
@@ -149,6 +221,12 @@ spec:
             - name: NRIA_LOG_FILE
               value: {{ .Values.logFile }}
             {{- end }}
+            {{- if .Values.privileged }}
+            - name: "NRIA_OVERRIDE_HOST_ROOT"
+              value: ""
+            - name: "NRIA_IS_SECURE_FORWARD_ONLY"
+              value: "true"
+            {{- end }}
```
roobre, line 229: Shouldn't `NRIA_IS_SECURE_FORWARD_ONLY` be set to `true` if **not** privileged?

e2e/charts/newrelic-infrastructure/templates/deployment.yaml
```diff
@@ -34,17 +34,20 @@ spec:
...
         - name: forwarder
-          image: newrelic/k8s-events-forwarder:1.20.5
-          imagePullPolicy: IfNotPresent
+          image: {{ .Values.forwarderImage.repository }}:{{ .Values.forwarderImage.tag }}
+          imagePullPolicy: {{ .Values.forwarderImage.pullPolicy }}
           securityContext:
-            privileged: false
+            {{- if .Values.privileged }}
+            privileged: true
+            {{- else }}
```
roobre, line 42: Ditto, I'm not sure scrapers care about the `privileged` flag.

e2e/e2e-values.yml
```diff
@@ -1,12 +1,16 @@
 verboseLog: true
-image:
+forwarderImage:
+  pullPolicy: IfNotPresent
+  tag: 1.20.5
+  repository: newrelic/k8s-events-forwarder
+agentImage:
   pullPolicy: Never
   tag: e2e
   repository: e2e/newrelic-infrastructure
```
roobre, line 9: Why do we have two agents?

Dockerfile-scraper
```diff
@@ -0,0 +1,8 @@
+FROM alpine:latest
+
+ADD --chmod=755 bin/nri-kubernetes-linux-amd64-scraper /var/db/newrelic-infra/newrelic-integrations/bin/
+
+# creating the nri-agent user used only in unprivileged mode
+RUN addgroup -g 2000 nri-agent && adduser -D -u 1000 -G nri-agent nri-agent
```
roobre, line 6: I don't see any reason for the scraper (`nri-kubernetes`) to not be unprivileged by default, as it shouldn't need to mount anything. We could then change the `USER` here to the newly created one and save some redundant files in the manifests :)

roobre: Perhaps we coudl rename this user to something that is the same in both images, e.g. `newrelic`.

e2e/charts/newrelic-infrastructure/templates/deployment.yaml
paologallinaharbur: it does not, you are right

e2e/charts/newrelic-infrastructure/templates/daemonset.yaml
paologallinaharbur: otherwise it fails since you cannot contact it thorugh localhost

cmd/nri-kubernetes/main.go
paologallinaharbur: yes this is still work in progress that will addressed in the next PR when merging Alvaro and Carlos work

e2e/e2e-values.yml
paologallinaharbur: Since the "forwarderImage" has the "NRIA_IS_FORWARD_ONLY" flag and it is merely forwarding KSM data
The "agentImage" is creating the entity, monitoring the underlying host, forwarding kubelet and control plane data

Dockerfile-scraper
paologallinaharbur: The user is still root, I am just creating the user, not setting it up.

We need to add it to have the proper name when running the image in unprivileged with "runAsUser 1000"

cmd/nri-kubernetes/main.go
paologallinaharbur: not yet :(

e2e/charts/newrelic-infrastructure/templates/daemonset.yaml
paologallinaharbur: yes, but likely we are going to garbage collect tmpfs-cache, therefore I would not invest on that now

e2e/charts/newrelic-infrastructure/templates/daemonset.yaml
paologallinaharbur: no!

e2e/e2e-values.yml
roobre: Got it, for a moment I forgot both "agent flavours" belonging to the same chart and actually being different images.

Dockerfile-scraper
roobre: > The user is still root, I am just creating the user, not setting it up.

This is exactly what I wanted to say, why not switch the user in the dockerfile itself?

paologallinaharbur: oh, ahhh, oh. Let me try

paologallinaharbur: We would lose some metrics 
 - https://docs.newrelic.com/docs/infrastructure/install-infrastructure-agent/linux-installation/linux-agent-running-modes/

paologallinaharbur: Sure, I added it!

src/sink/http.go
```diff
@@ -17,7 +17,7 @@ const (
 	// DefaultRequestTimeout is the default IO timeout for each request.
 	DefaultRequestTimeout = 15 * time.Second
 	// DefaultAgentForwarderEndpoint holds the default endpoint of the agent forwarder.
-	DefaultAgentForwarderEndpoint = "http://localhost:8001/v1/data"
+	DefaultAgentForwarderEndpoint = "http://localhost:%s/v1/data"
```
roobre, line 20: Might easily break with v6 addresses. Could we split this into two consts, e.g. `DefaultAgentForwarderAddress` and DefaultAgentForwarderPath`? That way we could use `net.JoinHostPort` which is v6-safe.

roobre:
> i.Publish() is not thread safe.

As an alternative to this, we could make a small wrapper around the `integration.Integration`, e.g.

```go
type integration {
    *integration.Integration
    mtx *sync.Mutex
}

func (i *integratiion) Publish() error {
  i.mtx.Lock()
  defer i.mtx.Unlock()

  return i.Integration.Publish()
}
```

wDYT?

roobre (approved):

paologallinaharbur:
It is not only the publish() not "thread safe", it is the whole integration obj. 
What happens if a component has written already 2  entities and it is adding metrics and suddenly the other components publishes and deletes all the entities?

We should think about this carefully, if we want to publish from multiple places. I see it as a risk, but on the other hand it is the only way to have different intervals (even if all the components are currently splitted 😅 )

PR #269, opened 2021-11-18
data generator for integration tests

roobre:
# `datagen`

Contains a set of dummy deployments and charts that can be used to scrape static test data from a live cluster, which then developers can add to the `data` folder. This data will be used for integration tests as of https://github.com/newrelic/nri-kubernetes/pull/263

## How it works

`datagen.sh` script will deploy a series of dummy and required workloads to the cluster, making it a cluster containing at least one resource of all the resources we can possibly monitor.

More specifically, it will deploy:
- [x] The KSM chart, for the KSM services/endpoints discovery tests
- [x] The [`e2e-resources`](../../../e2e/charts/e2e-resources) chart, which includes multiple samples of resources we monitor:
  - [x] PVs and PVCs
  - [x] Stuck/crashing pods

After this, it will deploy KSM and programmatically hit all the endpoints required for the integration to work, and store them in a directory specified by the user:

- [x] Kubelet endpoints
  - [x] `/pods` 
  - [x] `/stats/summary`
  - [x] `/metrics/cadvisor`
- [x] KSM `/metrics`
- [x] Control plane components
  - [x] api-server `/metrics`
  - [x] controller-manager `/metrics`
  - [x] etcd `/metrics`
  - [x] scheduler `/metrics`
- [x] HPA
- [x] MetricServer (`minikube enable metricserver`)

It will do this by spawning a privileged, `hostNetwork` `alpine:latest` pod in the cluster and running itself from inside.


internal/testutil/datagen/deployments/scraper.yaml
```diff
@@ -0,0 +1,74 @@
+apiVersion: v1
```
roobre, line 1: TODO: The idea is to put this in the e2e chart, pending @paologallinaharbur thoughs on that.

internal/testutil/datagen/README.md
```diff
@@ -0,0 +1,37 @@
...
+  - api-server `/metrics`
+  - controller-manager `/metrics`
+  - etcd `/metrics`
+  - scheduler `/metrics`
+
+It will do this by spawning a privileged, `hostNetwork` `alpine:latest` pod in the cluster and running itself from inside.
+
+## Usage
+
+TBD
```
roobre, line 37: `// TODO`

internal/testutil/datagen/datagen.sh
```diff
@@ -0,0 +1,211 @@
...
+        kubectl exec -n scraper $pod -- datagen.sh scrape controlplane scheduler > controlplane/scheduler/metrics
+    else
+        echo "Skipping control plane metrics"
+    fi
+
+    # K8s objects
+    echo "Generating list of kubernetes resources"
+    kubedump namespaces
+    kubedump services
+    kubedump pods
```
roobre, line 69: We need to add more here, at least `endpoints`.

internal/testutil/datagen/datagen.sh
```diff
@@ -0,0 +1,211 @@
...
+}
+
+# scrape will curl the specified component and output the response body to standard output.
+function scrape() {
+    bearer="Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)"
+    endpoint=unsupported://
+
+    case $1 in
+    ksm)
+      endpoint=http://ksm-kube-state-metrics.ksm.svc:8080/metrics
```
roobre, line 125: It would be great if these could be overridden from env vars, like `KSM_ENDPOINT=${KSM_ENDPOINT:-http://ksm-kube-state-metrics.ksm.svc:8080/metrics}`

internal/testutil/datagen/README.md
```diff
@@ -0,0 +1,37 @@
...
+* `kubectl` and `helm` installed and configured
+* Current `kubectl` context pointed to a preferably freshly-created minikube cluster.
+
+## How it works
+
+`datagen.sh` script will deploy a series of dummy and required workloads to the cluster, making it a cluster containing at least one resource of all the resources we can possibly monitor.
+
+More specifically, it will deploy:
+- The KSM chart, for the KSM services/endpoints discovery tests
+- The [`e2e-resources`](../../../e2e/charts/e2e-resources) chart, which includes multiple samples of resources we monitor:
```
paologallinaharbur, line 16: would make sense to add ksm as a e2e resource? Also Guillermo had issue with that since it is not included by default

Actually the e2e will likely install nri-bundle whenever is available deploying as well the ksm

internal/testutil/datagen/README.md
```diff
@@ -0,0 +1,37 @@
...
+  - `/stats/summary`
+  - `/metrics/cadvisor`
+- KSM `/metrics`
+- Control plane components
+  - api-server `/metrics`
+  - controller-manager `/metrics`
+  - etcd `/metrics`
+  - scheduler `/metrics`
+
+It will do this by spawning a privileged, `hostNetwork` `alpine:latest` pod in the cluster and running itself from inside.
```
paologallinaharbur, line 33: running itself from inside.?

internal/testutil/datagen/README.md
paologallinaharbur: I would have the:
 - scraper chart or even manifest
 - e2e-resources chart including ksm till we have it in nri-bundle

internal/testutil/datagen/datagen.sh
paologallinaharbur: also nodes

internal/testutil/datagen/README.md
roobre: If we include KSM in the `e2e-resources` then 2 ksms will be deployed in the `e2e` tests, right? Things should not break but I'd prefer if the e2e scenario could have the most vanilla env possible.

roobre: re scraper, I thought about adding it to the `e2e-resources` chart hidden behind a value which defautlts to false. WDYT?

internal/testutil/datagen/README.md
roobre: Yes! The script copies itself to the alpine pod and then calls `kubectl exec` etc. from there.
I'm not sure why I did it this way, there could also be two different scripts. Perhaps I thought I would share some logic.

internal/testutil/datagen/datagen.sh
```diff
@@ -0,0 +1,211 @@
...
+        exit 1
+    fi
+
+    mkdir -p "$1" && cd "$1"
+    testinfo > README.md
+
+    # Kubelet endpoints
+    mkdir -p kubelet
+    echo "Extracting kubelet /pods"
+    kubectl exec -n scraper $pod -- datagen.sh scrape kubelet pods > kubelet/pods
```
paologallinaharbur, line 29: we could check if `minikube tunnel` would cover all use cases. I though it did not, but running it against the api-server seems reaching it:
```
curl -X GET https://192.168.183.33:8443/metrics --header "Authorization: Bearer $TOKEN" --insecure
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {

  },
  "status": "Failure",
  "message": "forbidden: User \"system:anonymous\" cannot get path \"/metrics\"",
  "reason": "Forbidden",
  "details": {

  },
  "code": 403
}%


However, no idea if you prefer to run it inside the cluster or having to discover the IP, moreover in that case we would be tied to minikube


```

internal/testutil/datagen/datagen.sh
```diff
@@ -0,0 +1,211 @@
...
+\`\`\`
+
+\`git status --short\`
+
+\`\`\`
+$(git status --short)
+\`\`\`
+
+### \`$ kubectl version\`
+\`\`\`
```
paologallinaharbur, line 189: I do not see any obvious issue with the code, but do not trust my bash opinion 😢

internal/testutil/datagen/datagen.sh
roobre: Done

internal/testutil/datagen/datagen.sh
roobre: Done in https://github.com/newrelic/nri-kubernetes/pull/269/commits/1d8bb1f6a3d21f17dc16e9e4f0fa0e0ea373a6c1

internal/testutil/datagen/datagen.sh
roobre: I tried to choose a balance between minikube defaults (because it's what we use) and something that could be useful in other environments if they are not too harsh. The scraper pod is ugly but I think it will work in more scenarios, like kubeadm and even most managed envs (ofc without the control plane metrics).

internal/testutil/datagen/README.md
paologallinaharbur: sounds good! I wanted to avoid if possible to have multiple charts in the repo

internal/testutil/datagen/deployments/scraper.yaml
roobre: Done

internal/testutil/datagen/README.md
roobre: Done

e2e/charts/ksm/values.yaml
```diff
@@ -0,0 +1 @@
+kube-state-metrics: {}
```
paologallinaharbur, line 1: I am a bit confused, why adding a new chart with single dependency for ksm?
Cannot we  deploy directly ksm chart?

paologallinaharbur: I would have understood if it was a dependency of the scraper to place it there 😃

internal/testutil/datagen/README.md
```diff
@@ -0,0 +1,90 @@
...
+By default, `scraper.sh` will attempt to reach the controlpane endpoints through `localhost`. This will not work on managed K8s environments like EKS or GKE. By setting `DISABLE_CONTROLPLANE` to a non-empty value, `scraper.sh` will not attempt to reach control plane components.
+
+> Note: For scraping EKS/GKE clusters, it's also recommended to set `IS_MINIKUBE` to false (see below). 
+
+#### `IS_MINIKUBE`
+
+Aiming to be extensive by default, `datagen.sh` will assume it runs in a minikube environment. This means it will:
+- Try to run `minikube addons enable metrics-server`
+- Deploy some PVCs with storageClass set to standard (minikube default)
+- Deploy the `LoadBalancer` service with a fake `loadBalancerIP`
```
paologallinaharbur, line 74: Is this needed to collect some metrics?

paologallinaharbur (approved):
my only doubt is the location of KSM chart

internal/testutil/datagen/datagen.sh
```diff
@@ -0,0 +1,250 @@
...
+
+    # K8s objects
+    echo "Generating list of kubernetes resources"
+    kubedump nodes
+    kubedump namespaces
+    kubedump endpoints
+    kubedump services
+    kubedump pods
+
+    cd -
```
paologallinaharbur, line 96: could we save the location and cd to it instead of a generic `cd -`

internal/testutil/datagen/datagen.sh
```diff
@@ -0,0 +1,250 @@
...
+        helm dependency update ../../../e2e/charts/e2e-resources > /dev/null
+        helm upgrade --install e2e ../../../e2e/charts/e2e-resources -n $scrapper_namespace --create-namespace \
+          --set scraper.enabled=true \
+          --set persistentVolume.enabled=true \
+          $minikube_args \
+          $HELM_E2E_ARGS
+
+        echo "Installing KSM"
+        helm dependency update ../../../e2e/charts/ksm > /dev/null
+        helm upgrade --install ksm ../../../e2e/charts/ksm -n ksm --create-namespace --wait
```
paologallinaharbur, line 126: I would change this name to avoid collision with other deployment, testing it locally I got:
```
Error: rendered manifests contain a resource that already exists. Unable to continue with install: ClusterRole "ksm-kube-state-metrics" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-namespace" must equal "ksm": current value is "kube-system"

```

e2e/charts/ksm/values.yaml
roobre: The scraper needs ksm to be there to scrape it, but now that the scraper is part of the `e2e-resources` chart perhaps we could add ksm as a dependency from the latter. WDYT?

internal/testutil/datagen/README.md
roobre: Correct! https://github.com/newrelic/nri-kubernetes/blob/5a318a81ae3275fbc99140fc206050e8b143a4fb/src/metric/definition.go#L602

One less exception in the integration test :)

internal/testutil/datagen/datagen.sh
roobre: Not sure myself if it is more idiomatic to make `oldwd=$(pwd)` and then `cd $oldwd`, or just `cd -` 🤔 
Summoning our bash expert @kang-makes

internal/testutil/datagen/datagen.sh
roobre: Did you have a release also named `ksm` in `kube-system`? 😆 
Makes sense to me, I'll change it to `scraper-ksm` for example.

e2e/charts/ksm/values.yaml
paologallinaharbur: exactly!

internal/testutil/datagen/datagen.sh
paologallinaharbur: I had it in a different namespace, but cluster resources are overlapping 😖 
if it becomes part of the e2e resources it is not an issue anymore 😄

roobre: Done in [`a4dcdbe` (#269)](https://github.com/newrelic/nri-kubernetes/pull/269/commits/a4dcdbed6b90ed85dd53697caec64451f0db2b40) :)

internal/testutil/datagen/datagen.sh
kang-makes:
```bash
#!/bin/bash

function strip_home() {
  sed -e "s|$HOME|\$HOME|" <<<$1
}

echo Before subshell: `strip_home $PWD`
(
  cd $HOME/.local/bin
  echo Inside subshell: `strip_home $PWD`
)
echo After subshell: `strip_home $PWD`
```
So...
```
Before subshell: $HOME/.local/tmp
Inside subshell: $HOME/.local/bin
After subshell: $HOME/.local/tmp
```

roobre: Wrapped collection in a subshell here: https://github.com/newrelic/nri-kubernetes/pull/269/commits/271a7dd94a2772c7194916cc46bc6e13d6537fc4. WDYT?

paologallinaharbur (approved):
LGTM

PR #271, opened 2021-11-19
ci: implement caching, allow to skip e2e tests

### LINKED ISSUE #264: Improve caching in CI/CD jobs
CI/CD jobs, specially E2E, can be pretty intensive. In order to improve performance and be more eco-friendly, we should cache stuff when possible. At least, we should consider:
- [ ] Caching go modules so runners don't need to download them each time
- ~Caching docker layers (https://github.com/satackey/action-docker-layer-caching)~

roobre:
Fixes #264

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -22,6 +22,14 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: '1.17'
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
       - name: Setup Minikube
         uses: manusa/actions-setup-minikube@v2.4.2
         with:
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -8,6 +8,8 @@ on:
 
 jobs:
   e2eTests:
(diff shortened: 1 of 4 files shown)


.github/workflows/e2e.yaml
```diff
@@ -8,6 +8,8 @@ on:
 
 jobs:
   e2eTests:
+    # Do not run e2e tests if commit message or PR has skip-e2e.
+    if: ! ( contains(github.event.head_commit.message, 'skip-e2e') || contains(github.event.pull_request.title, 'skip-e2e') )
```
roobre, line 12: There is a way to do this with labels as well I think, but I'm not sure if we prefer that. What do you think?

paologallinaharbur: It would be enough with title and commits, but also labels would be nice with the opt in/out feature.

https://stackoverflow.com/questions/62325286/run-github-actions-when-pull-requests-have-a-specific-label

the others are a bit more "static"

paologallinaharbur (approved):
It fine with what you implemented, however it would be nicer with labels to avoid having to push fake commit to re-trigger a pipeline.

Up to you if it worth to implement it or not 😄

.github/workflows/e2e.yaml
roobre: Switched to `github.event.pull_request.labels.*.name, 'ci/skip-e2e'`, seems to be working fine :)

I gave a thought to reacting on the `unlabeled` event (to see if someone removes the `ci/skip-e2e` label), but I think that might backfire on us causing the e2e to be re-run every time anybody changes the labels on the PR for any other reason, as I haven't found a way to react to the event only if that label is being removed :(

Issue #238, opened 2021-10-11
Support generic authentication beyond bearer token


invidian:
## Is your feature request related to a problem? Please describe.

Right now, scraping metrics does not work when client certificates are used instead of bearer token in given kubeconfig.

## Feature Description

To simplify code, supporting generic authentication mechanisms from kubeconfig would be preferred.

## Additional context

Found while working on #217 

## Priority
Please help us better understand this feature request by choosing a priority from the following options: 
[Nice to Have, Really Want, Must Have, Blocker] 
Nice to Have

paologallinaharbur:
Kubeconfig has been added as well as transportFor, this will work however only leveraging connection through proxy

Issue #219, opened 2021-10-04
Modify autodiscovery for kubelet considering new Arch


paologallinaharbur:
Taking into account that the process needs to be longrunning verify 
 - what we are actually caching/discovering regarding kubelet
 - if we can simplify that code or improve performances thanks the re-arch

Issue #239, opened 2021-10-11
Clean up ways of handling bearer tokens


invidian:
## Is your feature request related to a problem? Please describe.

Right now, the following code is used to create authenticated HTTP clients:
https://github.com/newrelic/nri-kubernetes/blob/67ede685d0ee0f5be00ec903c0a1806f03304d5d/src/ksm/client/discovery.go#L139-L143
https://github.com/newrelic/nri-kubernetes/blob/67ede685d0ee0f5be00ec903c0a1806f03304d5d/src/kubelet/client/discovery.go#L93-L95
https://github.com/newrelic/nri-kubernetes/blob/67ede685d0ee0f5be00ec903c0a1806f03304d5d/src/controlplane/client/discovery.go#L131-L132

## Feature Description

It would be good to unify this code to use https://pkg.go.dev/k8s.io/client-go/rest#TransportFor, which would also solve #238.

## Additional context

Found while working on #217 

## Priority
Please help us better understand this feature request by choosing a priority from the following options: 
[Nice to Have, Really Want, Must Have, Blocker] 
Nice to Have

paologallinaharbur:
I tried it, however I carry on facing the following when connecting to kubelet:
```
14:33:45.389666 [DEBUG] Kubelet connection with nodeIP failed: no connection succeeded through localhost: error trying to connect to: https://192.168.183.33:10250/healthz. Got error: Get "https://192.168.183.33:10250/healthz": x509: cannot validate certificate for 192.168.183.33 because it doesn't contain any IP SANs
```

The certificate in the kubeconfig is of the ApiServer and it cannot be used to authenticate against kubelet (at least in minikube)

Adding a Skip verify `inClusterConfig.TLSClientConfig.Insecure = true` (that was already present) we noticed a weird behaviour 

```
14:29:02.879227 [DEBUG] trying connecting to kubelet directly with nodeIP
14:29:02.900059 [DEBUG] Kubelet connection with nodeIP failed: no connection succeeded through localhost: error trying to connect to: https://192.168.183.33:10250/healthz. Got error: Get "https://192.168.183.33:10250/healthz": x509: cannot validate certificate for 192.168.183.33 because it doesn't contain any IP SANs
14:29:02.900392 [DEBUG] trying connecting to kubelet directly with API proxy
[...]

PR #272, opened 2021-11-22
Kubelet refactor

paologallinaharbur:
The main goal of the PR is refactoring the deprecated KSM code removing cache and custom interfaces
 - migraded kubelet code following KSM approach (scraper - group - client)
 - removed custom kubernetes interface 
 - removed custom apiServer with cache custom interface
 - removed disk cache for network interface data
 - added informer to retrieve node data 
 - renamed getFamilyMetrics to clarify it
 - hostnetwork is not needed anymore for kubelet

---
Still pending
- [x] improve client unit tests
- [x] clarify if network call should be performed multiple times
- [x] changing kubelet to hostnetwork=false changes the hostname

---
If needed we can split up this PR into smaller ones, I tried to split as much as possible the changes into different commits

Fix https://github.com/newrelic/nri-kubernetes/issues/221 
Fix https://github.com/newrelic/nri-kubernetes/issues/219 
Fix https://github.com/newrelic/nri-kubernetes/issues/238

Regarding TransportFor refer to https://github.com/newrelic/nri-kubernetes/issues/239


src/kubelet/grouper/group.go
```diff
@@ -1,41 +1,75 @@
...
+
+// WithLogger returns an OptionFunc to change the logger from the default noop logger.
+func WithLogger(logger log.Logger) OptionFunc {
+	return func(kc *grouper) error {
+		kc.logger = logger
+		return nil
+	}
+}
+
+// New returns a data.Grouper that groups KSM metrics.
```
gsanchezgavier, line 39: KSM metrics?

cmd/nri-kubernetes/main.go
```diff
@@ -120,20 +138,53 @@ func setupKSM(c config.Mock, clients *clusterClients) (*ksm.Scraper, error) {
 	return ksmScraper, nil
 }
 
-func buildClients() (*clusterClients, error) {
-	k8s, err := client.NewKubernetes(true)
+func setupKubelet(c config.Mock, clients *clusterClients) (*kubelet.Scraper, error) {
+	providers := kubelet.Providers{
+		// TODO: Get rid of custom client.Kubernetes wrapper and use kubernetes.Interface directly.
```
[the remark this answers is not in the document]
paologallinaharbur: yes!

cmd/nri-kubernetes/main.go
```diff
@@ -77,7 +96,7 @@ func main() {
 	}
 }
 
-func runScrapers(c config.Mock, ksmScraper *ksm.Scraper, i *integration.Integration, clients *clusterClients) error {
+func runScrapers(c config.Mock, ksmScraper *ksm.Scraper, kubeletScraper *kubelet.Scraper, i *integration.Integration, clients *clusterClients) error {
```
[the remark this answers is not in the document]
paologallinaharbur: Yes! that was my idea, once the controlPlane implements `.Run` we can pass an array.
Avoiding as well the `if component.enabled`.

I can include that in this PR or in a following one, what do you think?

internal/config/config.go
```diff
@@ -13,12 +13,14 @@ type Mock struct {
...
-	ClusterName    string
-	Interval       time.Duration
+	NodeIP           string
+	Verbose          bool
+	NodeName         string
+	HTTPServerPort   string
+	Timeout          time.Duration
+	ClusterName      string
+	Interval         time.Duration
+	NetworkRouteFile string
```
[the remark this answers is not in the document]
paologallinaharbur: that is still a mock, that will be fixed in the config PR 😄

internal/discovery/node_discoverer_test.go
```diff
@@ -0,0 +1,53 @@
...
+	defer close(closeChan)
+
+	// Discovery with no node
+	e, err := d.Get("test-node")
+	require.Error(t, err)
+
+	// Discovery after creating a node
+	_, err = client.CoreV1().Nodes().Create(context.Background(), getFirstNode(), metav1.CreateOptions{})
+	require.NoError(t, err)
+	time.Sleep(time.Second)
```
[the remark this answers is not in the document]
paologallinaharbur: Sadly yes, it is a limitation of the fake informer implementation (maybe also of the real one).

Just after the deletion of the node, the discover still finds a node 😕

internal/discovery/node_discoverer_test.go
```diff
@@ -0,0 +1,53 @@
...
+	require.NoError(t, err)
+	assert.Equal(t, getFirstNode(), e)
+
+	// Discovery after deleting such node
+	err = client.CoreV1().Nodes().Delete(context.Background(), "first-node", metav1.DeleteOptions{})
+	require.NoError(t, err)
+	time.Sleep(time.Second)
+
+	e, err = d.Get("first-node")
+	require.Error(t, err)
```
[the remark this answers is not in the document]
paologallinaharbur: 👍

internal/discovery/node_discoverer_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	testclient "k8s.io/client-go/kubernetes/fake"
+
+	"github.com/newrelic/nri-kubernetes/v2/internal/discovery"
+)
+
```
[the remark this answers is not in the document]
paologallinaharbur: I added a code verifying the following (even if it is a feature of the informer itself we could have been returning a wrong channel)
```
	// Run starts and runs the shared informer, returning after it stops.
	// The informer will be stopped when stopCh is closed.
```

src/kubelet/client/client.go
```diff
@@ -0,0 +1,183 @@
...
+	hostURL := net.JoinHostPort(nodeIP, port)
+
+	var conn connParams
+
+	switch portInt {
+	case defaultHTTPKubeletPort:
+		conn = defaultConnParamsHTTP(hostURL)
+		err = checkCall(conn)
+		if err == nil {
+			return &conn, nil
```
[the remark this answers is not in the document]
paologallinaharbur: makes sense!

src/kubelet/client/client.go
```diff
@@ -0,0 +1,183 @@
...
+func (c *Client) setupAPIConnection(tripperBearerToken http.RoundTripper, apiServer string, nodeName string) (*connParams, error) {
+	c.logger.Debugf("connecting to kubelet directly with API proxy")
+
+	conn, err := connectionAPIProxy(apiServer, nodeName, tripperBearerToken)
+	if err != nil {
+		return nil, fmt.Errorf("creating connection parameters for API proxy: %w", err)
+	}
+
+	err = checkCall(conn)
+	if err != nil {
```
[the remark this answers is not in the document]
paologallinaharbur: you are right, I usually forget about inlining things

src/kubelet/client/client_test.go
```diff
@@ -0,0 +1,263 @@
...
+const (
+	healthz          = "/healthz"
+	apiProxy         = "/api/v1/nodes/test-node/proxy"
+	prometheusMetric = "/metric"
+	kubeletMetric    = "/kubelet-metric"
+	nodeName         = "test-node"
+)
+
+func TestClientCalls(t *testing.T) {
+	s, requests := testHTTPSServerWithEndpoints(t, []string{healthz, prometheusMetric, kubeletMetric})
```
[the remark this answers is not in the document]
paologallinaharbur: sure!

internal/discovery/node_discoverer.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+)
+
+// NodeGetter get nodes with informers.
+type NodeGetter interface {
+	Get(name string) (ret *corev1.Node, err error)
+}
```
roobre, line 12: Since this file contains only contains the getter rather than the discoverer we could just rename it `node_getter.go`.

internal/discovery/node_discoverer.go
```diff
@@ -0,0 +1,36 @@
...
+
+// NewNodesGetter returns a NodeGetter to get nodes with informers.
+func NewNodesGetter(client kubernetes.Interface, options ...informers.SharedInformerOption) (NodeGetter, chan<- struct{}) {
```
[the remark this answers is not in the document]
roobre: Do we really need a custom interface for this? Perhaps we could use `k8s.io/client-go/listers/core/v1/NodeLister`, and in fact a similar thing for services and endpoints.

src/client/interfaces.go
```diff
@@ -32,8 +31,3 @@ type HTTPGetter interface {
 type HTTPDoer interface {
 	Do(*http.Request) (*http.Response, error)
 }
-
-// NodeIPGetter allows getting discovered Node IP.
-type NodeIPGetter interface {
-	NodeIP() string
-}
```
roobre, line 39: ❤️ ❤️

src/controlplane/client/discovery.go
```diff
@@ -48,7 +51,7 @@ type ControlPlaneComponentClient struct {
 	tlsSecretNamespace       string
 	logger                   log.Logger
 	IsComponentRunningOnNode bool
-	k8sClient                client.Kubernetes
+	k8sClient                kubernetes.Interface
```
roobre, line 54: Yassss, `client.Kubernetes` BEGONE

src/kubelet/client/client.go
```diff
@@ -0,0 +1,158 @@
...
+	c.endpoint = conn.url
+
+	return c, nil
+}
+
+func (c *Client) setupConnection(kc kubernetes.Interface, tripperBearerToken http.RoundTripper, config config.Mock) (*connParams, error) {
+	kubeletPort, err := getKubeletPort(kc, config.NodeName)
+	if err != nil {
+		return nil, fmt.Errorf("getting kubelet port: %w", err)
+	}
```
roobre, line 76: Do we know if it's safe to error out if this fails? That is, if K8s _guarantees_ that the port will always be there. Otherwise we'd need to resort to config/default.
@kang-makes

src/kubelet/client/client.go
```diff
@@ -0,0 +1,158 @@
...
+
+	return conn, nil
+}
+
+func (c *Client) setupLocalConnection(tripperWithBearerToken http.RoundTripper, nodeIP string, portInt int32) (*connParams, error) {
+	c.logger.Debugf("connecting to kubelet directly with nodeIP")
+	var err error
+
+	port := fmt.Sprintf("%d", portInt)
```
roobre, line 99:
```suggestion
	port := fmt.Sprint(portInt)
```

src/kubelet/client/client.go
```diff
@@ -0,0 +1,158 @@
...
+	switch portInt {
+	case defaultHTTPKubeletPort:
+		if conn, err = checkConnectionHTTP(hostURL); err == nil {
+			return conn, nil
+		}
+
+	case defaultHTTPSKubeletPort:
+		if conn, err = checkConnectionHTTPS(hostURL, tripperWithBearerToken); err == nil {
+			return conn, nil
+		}
```
roobre, line 113: This would mean that users won't be able to force https on the default HTTP port and viceversa.

It might be something strange to do but I'd prefer to not limit ourselves in that regard: If possible, users should be able to override the protocol and skip this autodetection.

src/kubelet/client/client_mock.go
```diff
@@ -0,0 +1,57 @@
...
+	"path"
+
+	"github.com/newrelic/infra-integrations-sdk/log"
+
+	"github.com/newrelic/nri-kubernetes/v2/src/client"
+	"github.com/newrelic/nri-kubernetes/v2/src/prometheus"
+)
+
+// Client implements a client for Kubelet, capable of retrieving prometheus metrics from a given endpoint.
+type Mock struct {
```
roobre, line 16: How does this mock compare to using the real client implementation, but with the `testutil` fake server as a backend?

src/kubelet/client/client.go
```diff
@@ -0,0 +1,158 @@
...
+		return nil, fmt.Errorf("connecting to kubelet: %w", err)
+	}
+
+	c.doer = conn.client
+	c.endpoint = conn.url
+
+	return c, nil
+}
+
+func (c *Client) setupConnection(kc kubernetes.Interface, tripperBearerToken http.RoundTripper, config config.Mock) (*connParams, error) {
```
roobre, line 72: All the functions to discover and return the actual client seem like a big chunk of logic. Perhaps we could put them in a different object? Something like a `Discoverer` or `Prober`.

src/kubelet/client/client.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	conn, err := c.setupLocalConnection(tripperBearerToken, config.NodeIP, kubeletPort)
+	if err == nil {
+		c.logger.Debugf("connected to Kubelet directly with nodeIP")
+		return conn, nil
+	}
+
+	c.logger.Debugf("Kubelet connection with nodeIP failed: %v", err)
+	c.logger.Debugf("Connecting to kubelet directly with API proxy")
```
roobre, line 85: These messages might be relevant for the user, perhaps we could word them slighly less scary (do not say "failed") and bump the log level to info.

roobre (commented):
I like the new shape a lot!

Some things I would change:
- Considering having a dedicated object to perform discovery/probing of different ports and protocols
- Allow users to override the port and protocol through the config (and tests for that logic, as it will be messy)

cmd/nri-kubernetes/main.go
```diff
@@ -5,79 +5,98 @@ import (
...
+	exitClients = iota
+	exitIntegration
+	exitLoop
+	exitSetup
 )
 
 type clusterClients struct {
-	k8s client.Kubernetes
-	ksm ksmClient.MetricFamiliesGetter
+	k8s      kubernetes.Interface
```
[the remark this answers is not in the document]
paologallinaharbur: of `clusterClients` or of `k8s      kubernetes.Interface`?

(that is already the new interface, not the custom one)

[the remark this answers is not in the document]
roobre: You are right, I have no idea why I wrote this. Maybe I looked at the red part? Closing!

cmd/nri-kubernetes/main.go
```diff
@@ -162,3 +211,18 @@ func createIntegrationWithHTTPSink(httpServerPort string) (*integration.Integrat
...
 	return integration.New("com.newrelic.kubernetes", "test-ksm", integration.Writer(h))
 }
+
+func getK8sConfig(tryLocalKubeConfig bool) (*rest.Config, error) {
+	c, err := rest.InClusterConfig()
+	if err == nil || !tryLocalKubeConfig {
+		return c, nil
+	}
+
+	kubeconf := path.Join(homedir.HomeDir(), ".kube", "config")
```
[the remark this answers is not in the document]
paologallinaharbur: sure!

internal/discovery/node_discoverer.go
paologallinaharbur: I renamed it to node_lister.go since I changed it to leverage the k8s interface that is "lister"

src/kubelet/client/client.go
paologallinaharbur: I created a new object "connector"

src/kubelet/client/client_mock.go
paologallinaharbur: I removed that mock, but I kept the connector mock for now, mainly because when testing the client setting up all objs needed to perform discovery is a bit annoying (/healtz page), node with IP of the server, etc

src/kubelet/client/client.go
paologallinaharbur: Currently the user can set a port ignoring this field.
However, I believe that is "de facto" guaranteed that is populated even if I was not able to find that written somewhere 😕

paologallinaharbur:
I implemented your suggestion let me know what you think @roobre 

You can check only the last commits 😅 , it shows a lot of changes but it is merely moving the code around

src/kubelet/client/client.go
roobre: Do you think it's worth implementing a fallback to the default known port? Not sure myself if that would complicate the logic too much.

paologallinaharbur: So far it never happen that a single customer had issues with this, I would try to keep the probing logic smaller if possible to avoid hiding such error. By the way it is not a blocker since the user could still specify a port.

cmd/nri-kubernetes/main.go
```diff
@@ -214,17 +213,21 @@ func createIntegrationWithHTTPSink(httpServerPort string) (*integration.Integrat
...
 }
 
-func getK8sConfig(tryLocalKubeConfig bool) (*rest.Config, error) {
-	c, err := rest.InClusterConfig()
+func getK8sConfig(tryLocalKubeConfig bool, c config.Mock) (*rest.Config, error) {
+	inclusterConfig, err := rest.InClusterConfig()
 	if err == nil || !tryLocalKubeConfig {
-		return c, nil
+		return inclusterConfig, nil
 	}
```
roobre: I do not get the use case we are covering here: Is there a reason for not trying local kubeconfig in all cases?

internal/discovery/node_lister_test.go
```diff
@@ -0,0 +1,75 @@
...
+	require.NoError(t, err)
+	time.Sleep(time.Second)
+
+	// Discovery with closed informer
+	e, err := d.Get(nodeName)
+	require.Error(t, err)
+	require.Nil(t, e)
+}
+
+func getFirstNode() *corev1.Node {
```
roobre, line 69:
```suggestion
func fakeNode() *corev1.Node {
```

internal/discovery/services_lister.go
```diff
@@ -0,0 +1,25 @@
+package discovery
+
+import (
+	"time"
+
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	v1 "k8s.io/client-go/listers/core/v1"
```
roobre, line 8: We named this import `listersv1` in the other package. I don't mind either but we probably should choose only one.

internal/testutil/data/nodes.yaml
```diff
@@ -0,0 +1,83 @@
...
+        kubeProxyVersion: v1.22.2
+        kubeletVersion: v1.22.2
+        machineID: 92def845a9f14019b01864052bcea13b
+        operatingSystem: linux
+        osImage: Buildroot 2021.02.4
+        systemUUID: 6b6311ec-0000-0000-95bb-faffc2368d15
+kind: List
+metadata:
+  resourceVersion: ""
+  selfLink: ""
```
roobre, line 83: Genuinenly curious about how did you manage to autogenerate this without the trailing newline 😆

src/kubelet/client/connector.go
```diff
@@ -0,0 +1,252 @@
...
+	"github.com/newrelic/nri-kubernetes/v2/src/client"
+)
+
+const (
+	apiProxyPath = "/api/v1/nodes/%s/proxy/"
+	httpSchema   = "http"
+	httpsSchema  = "https"
+)
+
+type Connector interface {
```
roobre, line 27: It would be great to have some comments on exported types and functions. What does `Connector` do?

src/kubelet/client/connector.go
```diff
@@ -0,0 +1,252 @@
...
+	connect() (*connParams, error)
+}
+
+type defaultConnector struct {
+	// TODO: Use a non-sdk logger
+	logger             log.Logger
+	apiServerHost      string
```
roobre, line 34: Can we get this from the `kubernetes.Interface` instead?

src/kubelet/client/connector.go
```diff
@@ -0,0 +1,252 @@
...
+	kc                 kubernetes.Interface
+	tripperBearerToken http.RoundTripper
+	config             config.Mock
```
roobre, line 37: `config` is quite big, could we make it a pointer?

src/kubelet/client/connector.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+func (dp defaultConnector) connect() (*connParams, error) {
```
roobre, line 40: `defaultConnector` is also reasonably big, I'd consider using a pointer receiver to avoid copying too much.

src/kubelet/client/connector.go
```diff
@@ -0,0 +1,252 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("getting kubelet port: %w", err)
+	}
+
+	kubeletSchema := dp.getKubeletSchema(kubeletPort)
+	hostURL := net.JoinHostPort(dp.config.NodeIP, fmt.Sprint(kubeletPort))
+
+	conn, err := dp.setupLocalConnection(dp.tripperBearerToken, kubeletSchema, hostURL)
+	if err == nil {
+		dp.logger.Debugf("connected to Kubelet directly with nodeIP")
```
roobre, line 52: I'd promote this to info and include hostURL and schema)
```suggestion
		dp.logger.Infof("Connected to Kubelet through nodeIP %s", hostURL)
```

src/kubelet/client/connector.go
```diff
@@ -0,0 +1,252 @@
...
+
+	kubeletPort, err := dp.getKubeletPort()
+	if err != nil {
+		return nil, fmt.Errorf("getting kubelet port: %w", err)
+	}
+
+	kubeletSchema := dp.getKubeletSchema(kubeletPort)
+	hostURL := net.JoinHostPort(dp.config.NodeIP, fmt.Sprint(kubeletPort))
+
+	conn, err := dp.setupLocalConnection(dp.tripperBearerToken, kubeletSchema, hostURL)
```
roobre, line 50: I'd add a `log.Debugf` with `Trying to connect using...`

src/kubelet/client/connector.go
```diff
@@ -0,0 +1,252 @@
...
+	if err == nil {
+		dp.logger.Debugf("connected to Kubelet directly with nodeIP")
+		return conn, nil
+	}
+
+	dp.logger.Infof("Kubelet connection with nodeIP not working, falling back to API proxy: %v", err)
```
roobre, line 56: I'd log which URL will be attempted here as well as it includes potentially useful debug info such as the node name

src/kubelet/client/connector.go
```diff
@@ -0,0 +1,252 @@
...
+		if conn, err = dp.checkConnectionHTTPS(hostURL, tripperWithBearerToken); err == nil {
+			return conn, nil
+		}
+
+		if conn, err = dp.checkConnectionHTTP(hostURL); err == nil {
+			return conn, nil
+		}
+	}
+
+	return nil, fmt.Errorf("no connection succeeded through localhost: %w", err)
```
roobre, line 91:
```suggestion
	return nil, fmt.Errorf("no connection succeeded through nodeIP %s: %w", nodeIpWithSchema, err)
```

src/kubelet/client/connector.go
```diff
@@ -0,0 +1,252 @@
...
+}
+
+func (dp defaultConnector) getKubeletPort() (int32, error) {
+	if dp.config.Kubelet.Port != 0 {
+		dp.logger.Debugf("Setting Port %d as specified by user config", dp.config.Kubelet.Port)
+		return dp.config.Kubelet.Port, nil
+	}
+
+	//We pay the price of a single call getting a node to avoid asking the user the Kubelet port if different from the standard one
```
roobre, line 100:
```suggestion
	// We pay the price of a single call getting a node to avoid asking the user the Kubelet port.
```

src/kubelet/client/connector.go
```diff
@@ -0,0 +1,252 @@
...
+	node, err := dp.kc.CoreV1().Nodes().Get(context.Background(), dp.config.NodeName, metav1.GetOptions{})
+	if err != nil {
+		return 0, fmt.Errorf("getting node %q: %w", dp.config.NodeName, err)
+	}
+
+	port := node.Status.DaemonEndpoints.KubeletEndpoint.Port
```
roobre, line 106: Just to be sure, none of these fields is a pointer, right?

src/kubelet/client/connector.go
```diff
@@ -0,0 +1,252 @@
...
+
+	switch kubeletPort {
+	case defaultHTTPKubeletPort:
+		dp.logger.Debugf("Setting Kubelet Endpoint Schema http since kubeletPort is %d", kubeletPort)
+		return httpSchema
+	case defaultHTTPSKubeletPort:
+		dp.logger.Debugf("Setting Kubelet Endpoint Schema https since kubeletPort is %d", kubeletPort)
+		return httpsSchema
+	default:
+		dp.logger.Debugf("Schema is unknown since kubeletPort is %d", kubeletPort)
```
roobre, line 126:
```suggestion
		dp.logger.Errorf("Cannot automatically figure out schema from non-standard port %d, please set kubelet.schema in the config file.", kubeletPort)
```

src/kubelet/client/connector.go
```diff
@@ -0,0 +1,252 @@
...
+
+	if err = checkConnection(conn); err != nil {
+		return nil, fmt.Errorf("checking connection via API proxy: %w", err)
+	}
+
+	return &conn, nil
+}
+
+func (dp defaultConnector) checkConnectionHTTP(hostURL string) (*connParams, error) {
+	dp.logger.Debugf("testing kubelet connection with http to host: %s", hostURL)
```
roobre, line 161:
```suggestion
	dp.logger.Debugf("testing kubelet connection over plain http to %s", hostURL)
```

src/kubelet/client/connector.go
```diff
@@ -0,0 +1,252 @@
...
+	conn := defaultConnParamsHTTP(hostURL)
+	if err := checkConnection(conn); err != nil {
+		return nil, fmt.Errorf("checking connection via API proxy: %w", err)
+	}
+
+	return &conn, nil
+}
+
+func (dp defaultConnector) checkConnectionHTTPS(hostURL string, tripperBearer http.RoundTripper) (*connParams, error) {
+	dp.logger.Debugf("testing kubelet connection with https to host: %s", hostURL)
```
roobre, line 172:
```suggestion
	dp.logger.Debugf("testing kubelet connection over https to %s", hostURL)
```

src/kubelet/client/connector.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	resp, err := conn.client.Do(r)
+	if err != nil {
+		return fmt.Errorf("connecting to: %s. Got error: %s ", conn.url.String(), err)
+	}
+	defer resp.Body.Close() // nolint: errcheck
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("error calling endpoint %s. Got status code: %d", conn.url.String(), resp.StatusCode)
```
roobre, line 197:
```suggestion
		return fmt.Errorf("got non-200 status code: %d calling %s", resp.StatusCode, conn.url.String())
```

src/kubelet/client/connector.go
```diff
@@ -0,0 +1,252 @@
...
+	conn.url.Path = path.Join(conn.url.Path, healthzPath)
+
+	r, err := http.NewRequest(http.MethodGet, conn.url.String(), nil)
+	if err != nil {
+		return fmt.Errorf("creating request to: %s. Got error: %s ", conn.url.String(), err)
+	}
+
+	resp, err := conn.client.Do(r)
+	if err != nil {
+		return fmt.Errorf("connecting to: %s. Got error: %s ", conn.url.String(), err)
```
roobre, line 192:
```suggestion
		return fmt.Errorf("connecting to: %s: %w ", conn.url.String(), err)
```

src/kubelet/client/connector.go
```diff
@@ -0,0 +1,252 @@
...
+
+	return &conn, nil
+}
+
+func checkConnection(conn connParams) error {
+	conn.url.Path = path.Join(conn.url.Path, healthzPath)
+
+	r, err := http.NewRequest(http.MethodGet, conn.url.String(), nil)
+	if err != nil {
+		return fmt.Errorf("creating request to: %s. Got error: %s ", conn.url.String(), err)
```
roobre, line 187:
```suggestion
		return fmt.Errorf("connecting to: %s: %w ", conn.url.String(), err)
```

roobre (commented):
Review still not finished, but reached timeout for the day

cmd/nri-kubernetes/main.go
paologallinaharbur: mmm no real reason, it was a piece of code migrated from the old config

src/kubelet/client/connector.go
paologallinaharbur: no pointers

src/kubelet/client/connector.go
paologallinaharbur: at run-time you mean? I tried and I was not able to get it from the interface.

cmd/kubernetes-static/main.go
```diff
@@ -55,58 +54,47 @@ func main() {
...
+	mc := kubletClient.MockConnector{
+		URL:    *u,
+		Client: &http.Client{Timeout: time.Minute * 10},
+		Err:    nil,
+	}
+	kubeletClient, err := kubletClient.New(nil, config.Mock{}, &rest.Config{}, kubletClient.WithCustomConnector(mc))
+	if err != nil {
+		log.Fatal(err)
+	}
+
```
gsanchezgavier, line 82: wouldn't be possible to create the scraper for kubelet after this point? i'm a bit concerned that we have some logic in the scraper that needs to be reimplemented here, and also to make this main as simple and similar to main as possible.

cmd/nri-kubernetes/main.go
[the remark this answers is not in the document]
gsanchezgavier: let's do it in other ones  i think

cmd/kubernetes-static/main.go
paologallinaharbur: Actually it should never fail, since it is using the mock connector locally. However, if it fails there should be for sure a mistake in the way the mock client was populated

paologallinaharbur: regarding refactoring the static-main I agree, but I would do it in a different pr :)

src/kubelet/client/connector.go
```diff
@@ -78,7 +101,9 @@ func (dp defaultConnector) setupLocalConnection(tripperWithBearerToken http.Roun
 			return conn, nil
 		}
 	default:
-		// we were not able to infer the schema and the user did not provided it.
+		dp.logger.Errorf("Checking both HTTP and HTTPS since the schema was not detected automatically, " +
```
roobre, line 104: Is this an `Error`, or perhaps `Warn` or even `Info`?

src/kubelet/client/connector.go
```diff
@@ -184,17 +212,17 @@ func checkConnection(conn connParams) error {
 
 	r, err := http.NewRequest(http.MethodGet, conn.url.String(), nil)
 	if err != nil {
-		return fmt.Errorf("creating request to: %s. Got error: %s ", conn.url.String(), err)
+		return fmt.Errorf("creating request to %q : %s ", conn.url.String(), err)
```
roobre, line 215:
```suggestion
		return fmt.Errorf("creating request to %q: %v", conn.url.String(), err)
```

src/kubelet/client/connector.go
```diff
@@ -184,17 +212,17 @@ func checkConnection(conn connParams) error {
...
 	}
 
 	resp, err := conn.client.Do(r)
 	if err != nil {
-		return fmt.Errorf("connecting to: %s. Got error: %s ", conn.url.String(), err)
+		return fmt.Errorf("connecting to %q: %w ", conn.url.String(), err)
```
roobre, line 220:
```suggestion
		return fmt.Errorf("connecting to %q: %w", conn.url.String(), err)
```

src/kubelet/client/connector.go
```diff
@@ -123,7 +148,7 @@ func (dp defaultConnector) getKubeletSchema(kubeletPort int32) string {
 		dp.logger.Debugf("Setting Kubelet Endpoint Schema https since kubeletPort is %d", kubeletPort)
 		return httpsSchema
 	default:
-		dp.logger.Debugf("Schema is unknown since kubeletPort is %d", kubeletPort)
+		dp.logger.Errorf("Cannot automatically figure out schema from non-standard port %d, please set kubelet.schema in the config file.", kubeletPort)
```
roobre, line 151: If we cannot figure out the schema we still go on, so perhaps we could downgrade this to `Info` rather than error.

src/kubelet/kubelet_test.go
```diff
@@ -28,10 +26,11 @@ func TestScraper(t *testing.T) {
 	asserter := testutil.NewAsserter().
 		Using(metric.KubeletSpecs).
 		Silently().
```
roobre, line 28: I'd get rid of `Silently()` (in fact I had serious doubts about implementing it in the first place) as it will hide what we are not testing from the test output.

roobre (approved):
LGTM other than the logs/naming nits!

cmd/kubernetes-static/main.go
```diff
@@ -70,18 +67,13 @@ func main() {
 	}
 
 	// Kubelet
-	mc := kubletClient.MockConnector{
-		URL:    *u,
-		Client: &http.Client{Timeout: time.Minute * 10},
-		Err:    nil,
-	}
-	kubeletClient, err := kubletClient.New(nil, config.Mock{}, &rest.Config{}, kubletClient.WithCustomConnector(mc))
+	kubeletClient, err := kubletClient.New(kubletClient.FixedConnector(&http.Client{Timeout: time.Minute * 10}, *u))
```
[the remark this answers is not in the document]
paologallinaharbur: sure, makes sense

src/kubelet/client/connector.go
paologallinaharbur: mmm info!

PR #273, opened 2021-11-23
Allow testutil.Asserter to use user-defined excludeFuncs

roobre:
This ~tentative~ PR introduces `ExcludeFunc`, a function that can be user-defined and returns true if a metric should be allowed to be missing from the test.

This can be used to handle complex scenarios, like excluding/including a metric only if an entity matches a pattern. For example, with the previously known problem of not failing for container memory limit metrics one could do:

```go
	asserter := testutil.NewAsserter().
		Using(metric.KubeletSpecs).
		Excluding(func(group string, spec *definition.Spec, ent *integration.Entity) bool {
			return group == "container" && spec.Name == "memoryLimitBytes" && ent.Metadata.Name != "container-with-limits"  
		})
```

This way we will exclude the `memoryLimitBytes` metric for `group == "container"` if the entity is not named `container-with-limits`, which could be a container we specifically deploy in our `e2e-resources` chart with limits associated.

`testutil` package already includes some wrappers that return `ExcludeFunc`s, like `ExcludeOptional`, or `ExcludeMetric` which implement the same behavior that previously was built-in into the asserter.

As a not-so-pretty drawback of this approach, whole metric groups need to be excluded through a different mechanism, `ExcludingGroups`. The reason for this is that in order for the test to not fail where there are not any entities for an excluded group, the asserter needs to check this before iterating over the entities mapped to that group. This means it cannot call an `ExcludeFunc` as `entity` and `spec` would be `nil`. This could be set as a constraint for the `ExcludeFunc` API, but we thought this would be more unexpected and confusing than having a separate mechanism for excluding groups.

~In the current state this apprach is broken, as the asserter cannot apply groupExclusions before attempting to check if any entity is there.~

### DIFF
--- a/internal/testutil/assert.go
+++ b/internal/testutil/assert.go
@@ -9,15 +9,20 @@ import (
 	"github.com/newrelic/nri-kubernetes/v2/src/definition"
 )
 
-// ExcludeFunc is a function that returns true if a particular metric (spec) should be excluded from being asserted on ent.
+// ExcludeFunc is a function that returns true if a particular metric (spec) should be excluded from being asserted on
+// ent.
+// If an ExcludeFunc returns true for a given group, metric (spec) and entity, Asserter will not fail even if the metric
+// is not found.
 type ExcludeFunc func(group string, spec *definition.Spec, ent *integration.Entity) bool
 
+// ExcludeOptional returns an ExcludeFunc that excludes metrics marked as Optional.
 func ExcludeOptional() ExcludeFunc {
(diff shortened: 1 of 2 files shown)


paologallinaharbur (approved):
🎉

paologallinaharbur:
everything look good even if the title is still "wip"

roobre:
Requesting feedback from @gsanchezgavier as well since he was also in the conversation for this :)

gsanchezgavier (approved):
looks good !

PR #275, opened 2021-11-23
Disable renovate dashboard

kang-makes:
Renovate dashboard is a feature that creates an issue that updates with the pending updates that this repository has: https://docs.renovatebot.com/key-concepts/dashboard/

We follow these PRs renovate (and dependabot) creates using Zenhub.

Solves #205

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,6 +1,9 @@
 {
   "extends": [
-    "config:base"
+    "config:base",
+    // Disable the creation of this issue that renovate updates with the pending issue we follow with Zenhub:
+    // https://github.com/newrelic/nri-kubernetes/issues/205
+    ":disableDependencyDashboard"
   ],
   "enabledManagers": [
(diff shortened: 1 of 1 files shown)


gsanchezgavier (approved):
thanks for this!

roobre:
The dependency dasboard is however a way for renovate to tell us if something goes wrong when fetching updates. Do you know if it will still be created in this case?

kang-makes:
No. In the docs I put in the description you can see all that you can see in the dashboard. The only use case I could see useful is that you can approve so dependencies are automatically merged. I am not sure but I bet it will fail in repositories with protected branches.

It is "basically" an issue where all the PRs are listed and where you can check the checkbox that is in that issue to run renovate instantly.

roobre:
I cannot find a link because the issues were already solved, but in the past I have seen this dependency dashboard showing issues renovate itself found. For example, if after bumping a dependency `go mod` fails, or things like that.

kang-makes:
We decided that this is not needed as renovatebot still open PRs and the use cases that the docs gives us are not in out work flow.

PR #278, opened 2021-11-29
control plane refactor

gsanchezgavier:
First PR for control plane refactor:

- Changes the component discovery mechanism by a new one based on informers 
    - Removed: old discovery, discovery cache
- Moves component to new scraper structure


cmd/nri-kubernetes/main.go
```diff
@@ -97,7 +107,7 @@ func main() {
 	}
 }
 
-func runScrapers(c config.Mock, ksmScraper *ksm.Scraper, kubeletScraper *kubelet.Scraper, i *integration.Integration, clients *clusterClients) error {
+func runScrapers(c config.Mock, ksmScraper *ksm.Scraper, kubeletScraper *kubelet.Scraper, controlplaneScraper *controlplane.Scraper, i *integration.Integration, clients *clusterClients) error {
```
gsanchezgavier, line 110: just added extra parameter, but we discussed to refactor this in the future to have a scraper interface array but I didn't want to add an extra refactor to this PR

e2e/test-specs.yml
```diff
@@ -28,26 +28,6 @@ scenarios:
 
             - k8s.controllermanager.leaderElectionMasterStatus
 
-            - k8s.etcd.mvccDbTotalSizeInBytes
-            - k8s.etcd.networkClientGrpcReceivedBytesRate
-            - k8s.etcd.networkClientGrpcSentBytesRate
```
paologallinaharbur, line 33: 🎉

paologallinaharbur:
if you want we can merge before the refactor e2e and the pod lister

internal/deprecated/control-plane.go
```diff
@@ -44,11 +45,12 @@ func RunControlPlane(config config.Mock, k8s kubernetes.Interface, i *integratio
 		return err
 	}
 
-	nodeName := os.Getenv(nodeNameEnvVar)
-	if nodeName == "" {
-		logger.Errorf("%s env var should be provided by Kubernetes and is mandatory", nodeNameEnvVar)
-		os.Exit(1)
-	}
+	nodeName := config.NodeName
+	// nodeName := os.Getenv(nodeNameEnvVar)
```
paologallinaharbur: why commenting out and not deleting? (however I guess this disappear in the next commit)

cmd/nri-kubernetes/main.go
```diff
@@ -79,9 +79,19 @@ func main() {
...
+			logger.Errorf("setting up control plane scraper: %w", err)
+			os.Exit(exitSetup)
+		}
+		defer controlplaneScraper.Close()
+	}
+
 	for {
 		// TODO think carefully to the signature of this function
-		err := runScrapers(c, ksmScraper, kubeletScraper, i, clients)
+		err := runScrapers(c, ksmScraper, kubeletScraper, controlplaneScraper, i, clients)
```
paologallinaharbur, line 94: could we already add an interface here and call the function against []scrapers?

paologallinaharbur: I missed your comment literally two lines below

src/controlplane/components.go
```diff
@@ -26,7 +23,9 @@ type Component struct {
 	UseMTLSAuthentication           bool
 	Specs                           definition.SpecGroups
 	Queries                         []prometheus.Query
-	Labels                          []labels
+	Labels                          []componentLabels
+	// TODO this will be refactored with config per label group.
+	Namespace string
```
paologallinaharbur, line 28: In this file WithAPIServerSecurePort is not unused, is it intended? If so we should delete it

src/controlplane/client/discovery.go
```diff
@@ -35,28 +32,55 @@ func (i invalidTLSConfig) Error() string {
 	return i.message
 }
 
-type authenticationMethod string
```
paologallinaharbur, line 35: in this file const podEntityType = "pod" is unused

src/controlplane/client/discovery.go
```diff
@@ -35,28 +32,55 @@ func (i invalidTLSConfig) Error() string {
...
+type AuthenticationMethod string
 
 const (
-	none           authenticationMethod = "None (http)"
-	mTLS           authenticationMethod = "Mutual TLS"
-	serviceAccount authenticationMethod = "Service account (Bearer token)"
+	None           AuthenticationMethod = "None (http)"
+	MTLS           AuthenticationMethod = "Mutual TLS"
```
paologallinaharbur, line 39: should we comment these?

src/controlplane/controlplane_test.go
```diff
@@ -0,0 +1,184 @@
...
+		},
+		{
+			Name:      controlplane.APIServer,
+			NameSpace: "kube-system",
+			Labels:    map[string]string{"k8s-app": "kube-apiserver"},
+		},
+	}
+}
+
+func createControlPlainPods(t *testing.T, client *fake.Clientset, testServer *testutil.Server, testComponents []testComponent) {
```
paologallinaharbur, line 167: testServer *testutil.Server is unused

src/controlplane/components.go
```diff
@@ -115,15 +114,15 @@ func findComponentByName(name ComponentName, components []Component) *Component
 	return nil
 }
 
-// labels is a collection of labels, key-value style
-type labels map[string]string
+// componentLabels is a collection of componentLabels, key-value style
+type componentLabels map[string]string
```
paologallinaharbur, line 118: Should we use "labels.Set{}" or any k8s client object?

src/controlplane/grouper/group.go
```diff
@@ -1,4 +1,4 @@
-package controlplane
+package grouper
```
paologallinaharbur, line 1: if possible I would try to move to tests _test, ignore this if too refactor is required

src/controlplane/scraper.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (s *Scraper) deprecatedClient(c Component, podName string) client.HTTPClient {
+	timeout := 500 * time.Millisecond
+
+	authMethod := controlplaneClient.None
```
paologallinaharbur, line 125: controlplane or controlPlane. The same comment apply to the whole file :)

src/controlplane/scraper.go
```diff
@@ -0,0 +1,230 @@
...
+		selector := labels.SelectorFromSet(labels.Set(l))
+		pods, err := podLister.List(selector)
+		if err != nil {
+			return nil, fmt.Errorf("fail to list pods for component:%v selector: %v", c.Name, l)
+		}
+
+		// Validation of returned pods.
+		for _, pod := range pods {
+			if pod.Spec.NodeName != s.config.NodeName {
+				s.logger.Debugf("discarding pod: %s running outside the node", pod.Name)
```
paologallinaharbur, line 169: could we extract this to a function with a single return avoiding the continue and the break?

paologallinaharbur: maybe also return an error or a bool to indicate no pod was found to avoid `discoveredPod == nil {`

src/controlplane/scraper.go
```diff
@@ -0,0 +1,230 @@
...
+				continue
+			}
+			discoveredPod = pod
+			break
+		}
+
+	}
+
+	if discoveredPod == nil {
+		return nil, nil
```
paologallinaharbur, line 179: should be this an error? Since the user can opt in/out each component we might decide to fail if we are not able to retrieve data

internal/deprecated/control-plane.go
gsanchezgavier: I deleted this file actually, this will be vanished after rebase

src/controlplane/components.go
gsanchezgavier: yes all this will be removed on the next refactor when using the new config schema, I already removed all this func in the next branch :)

src/controlplane/components.go
gsanchezgavier: this will be gone in next PR with new config

src/controlplane/scraper.go
gsanchezgavier: this is the main issue of this func, the result could be pod, error , or no pod found but no error. Mainly because if there is an error on the lister i think we need to stop with error other wise issues related to configuration will just be noted on debug logs.

src/controlplane/client/discovery.go
gsanchezgavier: mmm this will probably gone when refactored the client

src/controlplane/scraper.go
gsanchezgavier: not sure about this convention, i though it was sub+package, 
`"github.com/newrelic/nri-kubernetes/v2/src/controlplane/client"` = `controlplaneClient`

src/controlplane/components.go
gsanchezgavier: just removed from here too

internal/discovery/pods_lister.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+type PodsListerConfig struct {
+	// Namespace can be used to restric the search to a particular namespace.
+	Namespace string
+	// Client is the Kubernetes client.Interface used to build informers.
+	Client kubernetes.Interface
+}
+
+func NewPodsLister(config PodsListerConfig) (v1.PodLister, chan<- struct{}) {
```
paologallinaharbur, line 18: All informes accept `options ...informers.SharedInformerOption` but none is actually leveraging that, either we include it here as well or we garbage collect them in the other places (I would do so)

internal/discovery/pods_lister.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"time"
+
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	v1 "k8s.io/client-go/listers/core/v1"
+)
+
+type PodsListerConfig struct {
```
paologallinaharbur, line 11: Everything is pod`S`lister but we are returning a v1.PodLister

internal/discovery/pods_lister.go
```diff
@@ -0,0 +1,34 @@
...
+	Client kubernetes.Interface
+}
+
+func NewPodsLister(config PodsListerConfig) (v1.PodLister, chan<- struct{}) {
+	// Arbitrary value, same used in Prometheus.
+	resyncDuration := 10 * time.Minute
+	factory := informers.NewSharedInformerFactoryWithOptions(
+		config.Client,
+		resyncDuration,
+		informers.WithNamespace(config.Namespace),
```
paologallinaharbur, line 24: if we are specifying the namespace here, should we return a "v1.PodNamespaceLister"?

src/controlplane/grouper/group.go
gsanchezgavier: not sure what you mean here

src/controlplane/scraper.go
gsanchezgavier: i think this is more readeable now since there is no break any more , wdyt?

internal/discovery/pods_lister.go
gsanchezgavier: yes make sense i did not looked at that

e2e/charts/newrelic-infrastructure/templates/clusterrole.yaml
```diff
@@ -20,9 +20,9 @@ rules:
       - "services"
       - "nodes"
       - "pods"
-    verbs: [ "get", "list","watch" ]
+    verbs: [ "get", "list", "watch" ]
   - nonResourceURLs: ["/metrics"]
-    verbs: ["get"]
+    verbs: ["get", "head"]
```
paologallinaharbur, line 25: Are we using this permission to check connection?

e2e/e2e-values.yml
```diff
@@ -38,25 +38,38 @@ scrapersConfig:
         - namespace: kube-system
           selector: "tier=control-plane,component=etcd"
           matchNode: true
-          url: http://localhost:2381
+          endpoints:
+            - url: http://localhost:2381
```
paologallinaharbur, line 42: this is what I meant, checking the nil pointer no need for `type` so specify no-auth :)

paologallinaharbur: actually, now that I think about it, you had the same idea

internal/testutil/server.go
```diff
@@ -17,11 +16,13 @@ func (s *Server) KSMEndpoint() string {
...
 	return s.Server.URL + "/ksm/metrics"
 }
+
 func (s *Server) KubeletEndpoint() string {
 	return s.Server.URL + "/kubelet"
 }
+
 func (s *Server) ControlPlaneEndpoint(component string) string {
-	return s.Server.URL + "/controlplane/" + strings.ReplaceAll(component, "/", "")
+	return s.Server.URL + "/controlplane/" + component + "/metrics"
```
paologallinaharbur, line 25: wouldn't path.Join()  safer?

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,274 @@
...
+		RootCAs:            clientCertPool,
+		InsecureSkipVerify: insecureSkipVerify,
+	}
+
+	tlsConfig.BuildNameToCertificate()
+
+	return tlsConfig, nil
+}
+
+type connParams struct {
```
paologallinaharbur, line 271: in the Kubelet client we take connParams and save into client.url and client.doer the data coming from that. 
I am fine with both approaches, but we need to follow the same

src/controlplane/client/client.go
```diff
@@ -1,203 +1,59 @@
...
-		k8sClient:       cfg.K8sClient,
-		InClusterConfig: cfg.InClusterConfig,
-		auth:            cfg.Auth,
-		endpoint:        u,
-	}
-
-	if err = c.configureAuthentication(); err != nil {
-		return nil, fmt.Errorf("fail configuring auth method: %w", err)
+		logger:           cfg.Logger,
+		connectionParams: cp,
```
paologallinaharbur, line 39: here in the other client the approach is different, we can agree on one and implement it :)

src/controlplane/client/client.go
```diff
@@ -1,203 +1,59 @@
...
 	}
 
 	return c, nil
 }
 
-// Get implements HTTPGetter interface by selecting proper authentication strategy for request
-// based on client configuration.
-//
-// TODO If secure request fails and insecure fallback is configured, request will be attempted over HTTP.
 func (c *Client) Get(urlPath string) (*http.Response, error) {
```
paologallinaharbur, line 45: urlPath is not used in this function, is there a bug?

paologallinaharbur: The others prometheus clients have the method:
```
func (c *Client) MetricFamiliesGetFunc(url string) prometheus.FetchAndFilterMetricsFamilies {
```

implementing: MetricFamiliesGetFunc interface, could we follow the same approach?

src/controlplane/scraper.go
```diff
@@ -100,24 +106,31 @@ func (s *Scraper) Run(i *integration.Integration) error {
...
+				autodiscover.Endpoints,
+				s.K8s,
+				s.inClusterConfig,
+				s.logger,
+			)
+			if err != nil {
+				return fmt.Errorf("control plane component %q failed creating connector: %v", component.Name, err)
+			}
+
 			client, err := controlplaneClient.New(controlplaneClient.Config{
```
paologallinaharbur: Each iteration we need to create the clients, I guess this is needed since the endpoints could change. 

In KSM the client is not created each time and it is a simple http pester, I guess we cannot do the same due to auth.

I realize that CP as both complexity of KSM (where to connect) and Kubelet (how to connect)

With informers this task is not too demanding, however, I would add some comments to explain why the approach is a bit different

src/controlplane/client/client.go
paologallinaharbur: or is it waiting for the grouper refactor?

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,274 @@
...
+	return nil
+}
+
+func validateAuth(auth *config.Auth) error {
+	if auth == nil {
+		return nil
+	}
+
+	switch {
+	case auth.Type == plainAuth:
```
paologallinaharbur, line 124: is it needed to have "plainAuth" if auth config can be nil?

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,274 @@
...
+		kc:              kc,
+		endpoints:       endpoints,
+	}, nil
+}
+
+// Connect iterates over the endpoints list and returns the connection parameters of the
+// first endpoint that respond Status OK.
+func (dp *defaultConnector) Connect() (*connParams, error) {
+	for _, e := range dp.endpoints {
+
```
paologallinaharbur, line 62: I would add a log.debug specifying what we are checking

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,274 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("creating HTTP client for endpoint %q: %w", e.URL, err)
+		}
+
+		if err := dp.probeEndpoint(u.String(), httpClient); err != nil {
+			dp.logger.Debugf("Autodiscover endpoint %q probe failed: %v", e.URL, err)
+			continue
+		}
+
+		// First endpoint with OK probe results returns.
```
paologallinaharbur, line 83: I would add a log.debug specifying that an endpoint worked

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,274 @@
...
+		return fmt.Errorf("mTLS config must exist")
+	}
+
+	if mTLS.TLSSecretName == "" {
+		return fmt.Errorf("TLSSecretName cannot be empty")
+	}
+	return nil
+}
+
+func (dp *defaultConnector) newHTTPClient(endpoint config.Endpoint) (*http.Client, error) {
```
paologallinaharbur, line 148: nit: I would reorder function to group all (dp *defaultConnector) ones

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,274 @@
...
+
+	secret, err := dp.kc.CoreV1().Secrets(namespace).Get(context.Background(), secretName, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("could not find secret %q containing TLS configuration: %w", secretName, err)
+	}
+
+	var cert, key, cacert []byte
+
+	var ok bool
+	if cert, ok = secret.Data["cert"]; !ok {
```
paologallinaharbur, line 219: We are unifying the use cases of mTLS and TLS, I believe that right now it is not possible to configure an endpoint with normal HTTPS. Do we want that?

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,274 @@
...
+	insecureSkipVerifyRaw, hasInsecureSkipVerify := secret.Data["insecureSkipVerify"]
+
+	if !hasCACert && !hasInsecureSkipVerify {
+		return nil, fmt.Errorf("both cacert and insecureSkipVerify are not set. One of them need to be set to be able to call ETCD metrics")
+	}
+
+	// insecureSkipVerify is set to false by default, and can be overridden with the insecureSkipVerify field
+	insecureSkipVerify := false
+	if hasInsecureSkipVerify {
+		insecureSkipVerify = strings.ToLower(string(insecureSkipVerifyRaw)) == "true"
```
paologallinaharbur, line 237: this is a bit obscure, why insecureSkipVerifyRaw is coming from the secret? Aren't the user setting it up in the config already?

src/controlplane/client/client_tls_test.go
```diff
@@ -58,36 +61,36 @@ func TestMutualTLSCalls(t *testing.T) {
...
-		{
-			name: "No config should fail",
-			assert: func(t *testing.T, resp *http.Response, err error) {
-				// todo: check if it's really the correct error
-				require.Error(t, err)
-			},
-		},
+		// {
+		// 	name:               "InsecureSkipVerify or CaCert should be set",
+		// 	insecureSkipVerify: boolPtr(false),
```
paologallinaharbur, line 66: we can just delete them if not needed anymore

src/controlplane/scraper.go
```diff
@@ -0,0 +1,270 @@
...
+		if logger == nil {
+			return fmt.Errorf("logger canont be nil")
+		}
+		s.logger = logger
+		return nil
+	}
+}
+
+// WithRestConfig returns an OptionFunc to change the restConfig from default empty config.
+func WithRestConfig(restConfig *rest.Config) ScraperOpt {
```
paologallinaharbur, line 55: would the scraper work if WithRestConfig is not passed? would make sense to have this as a parameter?

src/controlplane/scraper.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	// TODO If this could change without a restart of the pod we should run it each time we scrape data,
+	// possibly with a reasonable cache Es: NewCachedDiscoveryClientForConfig
+	s.k8sVersion, err = providers.K8s.Discovery().ServerVersion()
+	if err != nil {
+		return nil, fmt.Errorf("fetching K8s version: %w", err)
+	}
+
+	s.buildLister()
```
paologallinaharbur, line 98: I would add a comment for the us of the future

paologallinaharbur (commented):
Overall looks good, I would just iterate some small differences among clients we can agree on Tuesday.

I would add also more comments and logs since it is a complex piece of code

src/controlplane/client/connector.go
paologallinaharbur: and in general more logs so that in case of weird behaviours we can follow expactly in debug mode what is happening

cmd/nri-kubernetes/main.go
```diff
@@ -107,7 +117,7 @@ func main() {
 	}
 }
 
-func runScrapers(c *config.Config, ksmScraper *ksm.Scraper, kubeletScraper *kubelet.Scraper, i *integration.Integration, clients *clusterClients) error {
+func runScrapers(c *config.Config, ksmScraper *ksm.Scraper, kubeletScraper *kubelet.Scraper, controlplaneScraper *controlplane.Scraper, i *integration.Integration, clients *clusterClients) error {
 	if c.KSM.Enabled {
 		err := ksmScraper.Run(i)
```
alvarocabanas, line 122: Wouldn't it make more sense to have the function in the scrappers to be called Scrape instead of Run in all the scrappers, also, I agree with the interface scrapper as Paolo suggests and I see that is marked to fix it.

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,274 @@
...
+	default:
+		return fmt.Errorf("authorization type not supported: %q", auth.Type)
+	}
+
+	return nil
+}
+
+func validateMTLS(mTLS *config.MTLS) error {
+	if mTLS == nil {
+		return fmt.Errorf("mTLS config must exist")
```
alvarocabanas, line 139: In the cases where a new error is created, I prefer to create a custom Error at the top of the file like errConfigMustExist = errors.new("mTLS config must exist") and return it here. Then, it can be checked afterwards with errors.is(errConfigMustExist), for example on a unitTest.

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,274 @@
...
+}
+
+func validateAuth(auth *config.Auth) error {
+	if auth == nil {
+		return nil
+	}
+
+	switch {
+	case auth.Type == plainAuth:
+		break
```
alvarocabanas, line 125: You don't need the break, did you add it for readability?

src/controlplane/scraper.go
```diff
@@ -0,0 +1,270 @@
...
+func (s *Scraper) Run(i *integration.Integration) error {
+	var jobs []*scrape.Job
+	for _, component := range s.components {
+
+		job, err := s.externalEndpoint(component)
+		if err != nil {
+			return fmt.Errorf("configuring %q external endpoint: %w", component.Name, err)
+		}
+
+		if component.StaticEndpointConfig == nil {
```
alvarocabanas, line 113: Shouldn't this be put first on the loop and then on the externalEndpoint() removed the first condition? I point it bellow.

src/controlplane/scraper.go
```diff
@@ -0,0 +1,270 @@
...
+
+	}
+
+	return nil
+}
+
+// externalEndpoint builds the client based on the StaticEndpointConfig and fails if
+// the client probe cannot reach the endpoint.
+func (s *Scraper) externalEndpoint(c component) (*scrape.Job, error) {
+	if c.StaticEndpointConfig == nil {
```
alvarocabanas, line 142: I meant this condition in the previous comment.

src/controlplane/client/client.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+type Client struct {
+	logger           log.Logger
+	connectionParams *connParams
+}
+
+type Config struct {
+	Logger    log.Logger
+	Connector Connector
```
alvarocabanas, line 20: I don't understand why the connector is decoupled, could there be other connectors? Wouldn't it make more sense to pass all the connector config here and make the client construct the Default connector in the New? As I said, maybe I am loosing something and there can be other connectors.

alvarocabanas (approved):
Good job, I put a few comments.

src/controlplane/client/client.go
paologallinaharbur: In the kubelet it was embedded, in the first iteration, in the client itself and we ended up extracting it to reduce its logic keeping it "clean". 😃 
Right now the only other connected that exists is the fixedConnector, used in tests only.

src/controlplane/scraper.go
gsanchezgavier: I have divided this into autodicover and external endpoint functions with descriptions , let me know if it's clear enough otherwise i could add some inline comments.

src/controlplane/client/client.go
alvarocabanas: oks, I just saw the place where the client is created is really long with the population of the connector config, the creation of the connector, then the population of the client config and the client creation, and just wondered if it could be done differently. One possibility is to have the Client config accept the defaultConnector parameters, plus a Connector pointer, if the connector is nil, then the DefaultConnector is created inside the client's new, if not the Connector passed is used. But really, it's not important, it's just a matter of taste.

src/controlplane/client/connector.go
alvarocabanas: oks, yes, for sure make it the same approach everywhere is the best.

alvarocabanas: Maybe can be discussed for a final refactor.

internal/discovery/pods_lister.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+type PodListerConfig struct {
+	// Namespace can be used to limit the search to a particular namespace.
+	Namespace string
+	// Client is the Kubernetes client.Interface used to build informers.
+	Client kubernetes.Interface
+}
+
+func NewPodLister(config PodListerConfig) (v1.PodNamespaceLister, chan<- struct{}) {
```
paologallinaharbur, line 18: NewPodNamespaceLister instead?

internal/testutil/server.go
```diff
@@ -17,11 +17,13 @@ func (s *Server) KSMEndpoint() string {
...
 	return s.Server.URL + "/ksm/metrics"
 }
+
 func (s *Server) KubeletEndpoint() string {
 	return s.Server.URL + "/kubelet"
 }
+
 func (s *Server) ControlPlaneEndpoint(component string) string {
-	return s.Server.URL + "/controlplane/" + strings.ReplaceAll(component, "/", "")
+	return s.Server.URL + path.Join("/controlplane", component, "metrics")
```
paologallinaharbur, line 26: why keeping the `s.Server.URL + path.Join` and not everything in the join?

src/controlplane/client/connector.go
```diff
@@ -43,6 +42,22 @@ type defaultConnector struct {
...
+
+	if logger == nil {
+		return nil, fmt.Errorf("logger cannot be nil")
+	}
+
+	if kc == nil {
+		return nil, fmt.Errorf("kubernetes interface cannot be nil")
+	}
+
+	if len(endpoints) == 0 {
```
paologallinaharbur, line 57: could we move it to validateEndpointConfig? Or maybe is better to keep it with other checks

src/controlplane/client/connector.go
paologallinaharbur: Talked offline

src/controlplane/scraper.go
paologallinaharbur: If I am not wrong rest config is passed to the Defalut connector that fails if it is nil, no?
```
	if inClusterConfig == nil {
		return nil, fmt.Errorf("inClusterConfig cannot be nil")
	}
```

Therefore it seems required

paologallinaharbur (approved):
Thanks the whole work!
There are still some differences with the other clients, but we can address them in the grouper PR

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,294 @@
...
+// Connector provides an interface to retrieve []connParams to connect to a Control Plane instance.
+type Connector interface {
+	Connect() (*connParams, error)
+}
+
+type defaultConnector struct {
+	// TODO: Use a non-sdk logger
+	logger          log.Logger
+	kc              kubernetes.Interface
+	inClusterConfig *rest.Config
```
roobre, line 39: By seeing this I get the feel that the connector is doing too many things: Checking connectivity and handling authentication. Perhaps we could build a separate object, some sort of "authenticator"? This object could receive these two args, and together with the `auth` section of the config, take care of authenticating requests. WDYT?

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,294 @@
...
+	endpoints       []config.Endpoint
+}
+
+// DefaultConnector returns a defaultConnector that probes all endpoints in the list and return the first responding status OK.
+func DefaultConnector(endpoints []config.Endpoint, kc kubernetes.Interface, inClusterConfig *rest.Config, logger log.Logger) (Connector, error) {
```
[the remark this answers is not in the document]
roobre: As mentioned above, perhaps we could split this responsibility away from the connector.

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,294 @@
...
+	for _, e := range dp.endpoints {
+		dp.logger.Debugf("Configuring autodiscover endpoint %q for probing", e.URL)
+
+		u, err := url.Parse(e.URL)
+		if err != nil {
+			return nil, fmt.Errorf("parsing endpoint url %q: %w", e.URL, err)
+		}
+
+		// If no path is defined on the config, default is set.
+		if u.Path == "" || u.Path == "/" {
```
roobre, line 85:
```suggestion
		if strings.TrimSuffix(u.Path, "/") == "" {
```

So we don't mess up if path is `http://localhost///`

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,294 @@
...
+func (dp *defaultConnector) Connect() (*connParams, error) {
+	for _, e := range dp.endpoints {
+		dp.logger.Debugf("Configuring autodiscover endpoint %q for probing", e.URL)
+
+		u, err := url.Parse(e.URL)
+		if err != nil {
+			return nil, fmt.Errorf("parsing endpoint url %q: %w", e.URL, err)
+		}
+
+		// If no path is defined on the config, default is set.
```
roobre, line 84: Comment seems redundant, `log.` line below is already telling me what the if does

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,294 @@
...
+			u.Path = defaultMetricsPath
+		}
+
+		httpClient, err := dp.newHTTPClient(e)
+		if err != nil {
+			return nil, fmt.Errorf("creating HTTP client for endpoint %q: %w", e.URL, err)
+		}
+
+		if err := dp.probeEndpoint(u.String(), httpClient); err != nil {
+			dp.logger.Debugf("Autodiscover endpoint %q probe failed: %v", e.URL, err)
```
roobre, line 96: I'd clarify that we'll try the next thing to avoid making the message sound more alarming than it is.
```suggestion
			dp.logger.Debugf("Autodiscover endpoint %q probe failed, skipping: %v", e.URL, err)
```

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,294 @@
...
+		return &connParams{url: *u, client: httpClient}, nil
+	}
+
+	return nil, fmt.Errorf("all autodiscover endpoints probe failed to response")
+}
+
+func (dp *defaultConnector) probeEndpoint(url string, client *http.Client) error {
+	resp, err := client.Head(url)
+	if err != nil {
+		return fmt.Errorf("http request failed with error: %w", err)
```
roobre, line 111:
```suggestion
		return fmt.Errorf("http HEAD request failed: %w", err)
```

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,294 @@
...
+	if err := dp.configureAuthentication(client, endpoint); err != nil {
+		return nil, fmt.Errorf("configuring auth: %w", err)
+	}
+
+	return client, nil
+}
+
+func (dp *defaultConnector) configureAuthentication(httpClient *http.Client, endpoint config.Endpoint) error {
+	if endpoint.Auth == nil {
+		dp.logger.Debugf("Plain auth is used for endpoint %q", endpoint.URL)
```
roobre, line 141:
```suggestion
		dp.logger.Debugf("No authentication configured for %q, connection will be attempted anonymously", endpoint.URL)
```

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,294 @@
...
+
+		return nil
+	}
+
+	if strings.EqualFold(endpoint.Auth.Type, bearerAuth) {
+		dp.logger.Debugf("kubernetes Bearer token auth is used for endpoint %q", endpoint.URL)
```
roobre, line 147:
```suggestion
		dp.logger.Debugf("Using kubernetes token to authenticate request to %q", endpoint.URL)
```

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,294 @@
...
+
+		httpClient.Transport = transport.NewBearerAuthRoundTripper(dp.inClusterConfig.BearerToken, httpClient.Transport)
+
+		return nil
+	}
+
+	if strings.EqualFold(endpoint.Auth.Type, mTLSAuth) {
+		dp.logger.Debugf("MTLS auth is used for endpoint %q", endpoint.URL)
```
roobre, line 155:
```suggestion
		dp.logger.Debugf("Using mTLS to authenticate request to %q", endpoint.URL)
```

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,294 @@
...
+		}
+
+		httpClient.Transport = &http.Transport{
+			TLSClientConfig: tlsConfig,
+		}
+
+		return nil
+	}
+
+	return fmt.Errorf("authorization type not supported: %q", endpoint.Auth.Type)
```
roobre, line 169:
```suggestion
	return fmt.Errorf("unknown authorization type %q", endpoint.Auth.Type)
```

src/controlplane/client/connector.go
roobre: This is certainly weird. I'd suggest to keep compatibility (overriding user config) but add a `log.Warn` and a `// TODO` saying the feature is deprecated and slated for removal soon.

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,294 @@
...
+		return nil
+	}
+
+	switch {
+	case strings.EqualFold(auth.Type, bearerAuth):
+		break
+	case strings.EqualFold(auth.Type, mTLSAuth):
+		return validateMTLS(auth.MTLS)
+	default:
+		return fmt.Errorf("authorization type not supported: %q", auth.Type)
```
roobre, line 273: We are doing this twice, right? I'd get rid of early validation and just fail when we attempt to use it.

src/controlplane/controlplane_test.go
```diff
@@ -0,0 +1,310 @@
...
+	controlPlainSpecs["scheduler"] = metric.SchedulerSpecs["scheduler"]
+	controlPlainSpecs["api-server"] = metric.APIServerSpecs["api-server"]
+
+	asserter := testutil.NewAsserter().
+		Using(controlPlainSpecs).
+		Excluding(
+			testutil.ExcludeMetrics("controller-manager", excludeCM...),
+			testutil.ExcludeMetrics("etcd", excludeETCD...),
+			testutil.ExcludeMetrics("scheduler", excludeS...),
+			testutil.ExcludeMetrics("api-server", excludeAS...),
```
roobre, line 48: Why do we exclude this? I'm sure we have good reasons to, but cannot figure out which. Could we leave a comment explaining them?

src/controlplane/controlplane_test.go
```diff
@@ -0,0 +1,310 @@
...
+		)
+
+	for _, version := range testutil.AllVersions() {
+		version := version
```
roobre, line 52: Does this work if variables are named the same? 🤯

src/controlplane/controlplane_test.go
```diff
@@ -0,0 +1,310 @@
...
+		controlplane.Providers{
+			K8s: fakeK8s,
+		},
+	)
+	if err != nil {
+		t.Fatalf("building scraper should not fail: %v", err)
+	}
+
+	// create a scheduler pod on different node
+	createControlPlainPod(t, fakeK8s, controlplane.Scheduler, discoveryConfig[controlplane.Scheduler], "masterNode2")
```
roobre, line 126:
```suggestion
	createControlPlanePod(t, fakeK8s, controlplane.Scheduler, discoveryConfig[controlplane.Scheduler], "masterNode2")
```

src/controlplane/controlplane_test.go
```diff
@@ -0,0 +1,310 @@
...
+				Enabled:   true,
+				Scheduler: testConfig.ControlPlane.Scheduler,
+			},
+		},
+		controlplane.Providers{
+			K8s: fakeK8s,
+		},
+	)
+	if err != nil {
+		t.Fatalf("building scraper should not fail: %v", err)
```
roobre, line 122:
```suggestion
		t.Fatalf("error building scraper: %v", err)
```

src/controlplane/controlplane_test.go
```diff
@@ -0,0 +1,310 @@
...
+	client *fake.Clientset,
+	componentName controlplane.ComponentName,
+	autodiscovery config.AutodiscoverControlPlane,
+	nodeName string,
+) {
+	t.Helper()
+	labelsSet, _ := labels.ConvertSelectorToLabelsMap(autodiscovery.Selector)
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      string(componentName) + fmt.Sprintf("%d", rand.Int()), // nosemgrep // add rand to allow create same component multiple times.
```
roobre, line 297: Good ole mateusz whispered this insane trick to me: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/rand#String

src/controlplane/controlplane_test.go
```diff
@@ -0,0 +1,310 @@
...
+			Namespace: autodiscovery.Namespace,
+			Labels:    labelsSet,
+		},
+		Spec: corev1.PodSpec{
+			NodeName: nodeName,
+		},
+	}
+	if _, err := client.CoreV1().Pods(autodiscovery.Namespace).Create(context.Background(), pod, metav1.CreateOptions{}); err != nil {
+		t.Fail()
```
roobre, line 306:
```suggestion
		t.Fatalf("error creating pods in fake client: %v", err)
```

src/controlplane/scraper.go
```diff
@@ -0,0 +1,282 @@
...
+
+	for _, component := range s.components {
+		var job *scrape.Job
+
+		var err error
+
+		// Static endpoint take precedence over autodisover and fails if external endpoint
+		// cannot be scraped.
+		if component.StaticEndpointConfig != nil {
+			s.logger.Debugf("Building %q external endpoint job")
```
roobre, line 119:
```suggestion
			s.logger.Infof("Using static endpoint for component %s")
```

⬆️  Argument for `%q` seems to be missing

src/controlplane/scraper.go
```diff
@@ -0,0 +1,282 @@
...
+
+			job, err = s.externalEndpoint(component)
+			if err != nil {
+				return fmt.Errorf("configuring %q external endpoint: %w", component.Name, err)
+			}
+		} else {
+			s.logger.Debugf("Building %q autodiscovered endpoint job")
```
roobre, line 126:
```suggestion
			s.logger.Debugf("Autodiscovering pods for component %s")
```

src/controlplane/scraper.go
```diff
@@ -0,0 +1,282 @@
...
+	var jobs []*scrape.Job
+
+	for _, component := range s.components {
+		var job *scrape.Job
+
+		var err error
+
+		// Static endpoint take precedence over autodisover and fails if external endpoint
+		// cannot be scraped.
+		if component.StaticEndpointConfig != nil {
```
roobre, line 118: I wonder why we if-else based on this. Doesn't the connector handle both scenarios on its own? Why does the scraper need to check this?

src/controlplane/scraper.go
```diff
@@ -0,0 +1,282 @@
...
+
+	return scrape.NewScrapeJob(string(c.Name), grouper, c.Specs), nil
+}
+
+// autodiscover will iterate over the Autodiscovery configs from a component and for each:
+//  - Discover if any pod matches the selector.
+//  - Build the client, which probes all the endpoints in the list.
+// It uses the first autodiscovery config that can satisfy conditions above.
+// It doesn't fail if no autodiscovery satisfy the contitions.
+func (s *Scraper) autodiscover(c component) (*scrape.Job, error) {
```
roobre, line 191: Oh, I see now: Autodiscovery is not handled by the connector but by the scraper. It might be that the scraper is doing too many things then. Do you think it would be worth it to extract out a component discoverer object, which in turn should be able to handle both autodiscovery and static endpoints?

src/controlplane/scraper.go
```diff
@@ -0,0 +1,282 @@
...
+	for _, autodiscover := range c.AutodiscoverConfigs {
+		podName, err := s.discoverPod(autodiscover)
```
roobre, line 193: I would make the discovery operation return a `*corev1.Pod`. Right now we're only using the labels but in the future we might want to check more info.

src/controlplane/scraper.go
```diff
@@ -0,0 +1,282 @@
...
+		s.logger.Debugf("Found %q pod %q with labels %q", c.Name, podName, autodiscover.Selector)
+
+		connector, err := controlplaneClient.DefaultConnector(
+			autodiscover.Endpoints,
+			s.K8s,
+			s.inClusterConfig,
+			s.logger,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("control plane component %q failed creating connector: %v", c.Name, err)
```
roobre, line 212:
```suggestion
			return nil, fmt.Errorf("creating connector for %s: %v", c.Name, err)
```

src/controlplane/scraper.go
```diff
@@ -0,0 +1,282 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("control plane component %q discovery failed: %v", c.Name, err)
+		}
+
+		if podName == "" {
+			s.logger.Debugf("No %q pod found with labels %q", c.Name, autodiscover.Selector)
+			continue
+		}
+
+		s.logger.Debugf("Found %q pod %q with labels %q", c.Name, podName, autodiscover.Selector)
```
roobre, line 203:
```suggestion
		s.logger.Debugf("Found pod %q for %s with labels %q", podName, c.Name, autodiscover.Selector)
```

roobre (changes requested):
This looks fantastic overall, saving some nits and messages!
My only concern is that both the Scraper and the Connector seem to be doing many things. Could we consider splitting them into smaller objects?

src/controlplane/controlplane_test.go
paologallinaharbur: golang magic

internal/testutil/server.go
gsanchezgavier: doesn't work for url just for path https://go.dev/play/p/qX93BzDNKXY

src/controlplane/client/connector.go
```diff
@@ -0,0 +1,230 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("creating HTTP client for endpoint %q: %w", e.URL, err)
+		}
+
+		if err := dp.probeEndpoint(u.String(), httpClient); err != nil {
+			dp.logger.Debugf("Endpoint %q probe failed, skipping: %v", e.URL, err)
+			continue
+		}
+
+		dp.logger.Debugf("Endpoint %q probed successfully", e.URL)
```
roobre, line 79: I'd bump this to `Infof` if we're not logging the chosen endpoint anywhere else, it could be useful information.

src/controlplane/client/connector.go
roobre: On a second thought, I do wonder what happens if I want to point the integration to a custom endpoint that _does_ serve metrics on `/` 🤔

src/controlplane/controlplane_test.go
roobre: And don't forget the most important feature of kubernetes' `rand.String`: It excludes vowels and vowel-like numbers so your pods don't contain suffixes like `mydeployment-fuck4`.

src/controlplane/scraper.go
roobre: Uh, didn't see it was being printed on each iteration. Definitely should not be `Info` then, good catch!

roobre (approved):
LGTM, only things I'd change are the connector handling out and perhaps the scraper handling discovery, but we can leave those for a follow-up pr (i.e. #289 )

src/controlplane/client/connector.go
gsanchezgavier: the probing happen on each run of the scraper, wouldn't be this a bit verbose?

gsanchezgavier:
codespell error is on the generated data form testutil

PR #279, opened 2021-11-29
Feat/config file

paologallinaharbur:
This superseeds https://github.com/newrelic/nri-kubernetes/pull/267 following the same approach, but using the new config.

I would like as well to sort of support hot-reload:
```
viper.OnConfigChange(func(e fsnotify.Event) {
	fmt.Println("Config file changed:", e.Name)
})
viper.WatchConfig()
```

it is not straightforward since  many things should be considered, however we could also trigger a restart of the pod with os.exit(0)-configChanged


paologallinaharbur:
I'll fix imports/linting issues before merging to reduce toil

internal/config/config.go
```diff
@@ -1,144 +1,126 @@
...
+	Autodiscover   []AutodiscoverControlPlane `mapstructure:"autodiscover"`
 }
 
-type ControllerManager struct {
-	ControllerManagerEndpointURL string
+type AutodiscoverControlPlane struct {
+	Namespace string `mapstructure:"namespace"`
+	Selector  string `mapstructure:"selector"`
+	MatchNode bool   `mapstructure:"matchNode"`
+	URL       string `mapstructure:"url"`
```
gsanchezgavier, line 82: I think this will need to be `[]string` to support different common options. [currently](https://github.com/newrelic/nri-kubernetes/blob/main/src/controlplane/components.go#L186) API server has 2 preconfigured. But perhaps @roobre  or @kang-makes has other opinions.

internal/config/example_data/config.yml
```diff
@@ -0,0 +1,32 @@
+clusterName: dummy_cluster
```
alvarocabanas, line 1: I would call this folder as testdata instead of example_data, the Go tool will ignore any directory in your $GOPATH that starts with a period, an underscore, or matches the word testdata, as stated in https://dave.cheney.net/2016/05/10/test-fixtures-in-go

paologallinaharbur: 🚀  Thanks! I did not know that!

internal/config/config.go
```diff
@@ -1,144 +1,126 @@
...
+	Enabled        bool                       `mapstructure:"enabled"`
+	StaticEndpoint StaticEndpoint             `mapstructure:"staticEndpoint"`
+	Autodiscover   []AutodiscoverControlPlane `mapstructure:"autodiscover"`
 }
 
 type Scheduler struct {
-	SchedulerEndpointURL string
+	Enabled        bool                       `mapstructure:"enabled"`
+	StaticEndpoint StaticEndpoint             `mapstructure:"staticEndpoint"`
+	Autodiscover   []AutodiscoverControlPlane `mapstructure:"autodiscover"`
```
gsanchezgavier, line 75: would make sense to create a controlplane config abstraction and load a map of that? instead of having a struct per each component just have a single one i mean.

e2e/e2e-values.yml
```diff
@@ -18,6 +17,27 @@ resources:
...
+  networkRouteFile: "/proc/net/route"
+
+  kubelet:
+    enabled: true
+
+  ksm:
+    enabled: true
+
+  controlPlane:
+    enabled: true
```
gsanchezgavier, line 34: should we keep the enabled config in true here? i think we are deploying this into 2 nodes en minikube and one of this is not the master

gsanchezgavier: anyway it should not fail :) so perhaps is good to test that

internal/config/config.go
```diff
@@ -1,144 +1,126 @@
...
-	ETCD              ETCD
-	APIServer         APIServer
-	Scheduler         Scheduler
-	ControllerManager ControllerManager
-	Enabled           bool
+	Enabled           bool              `mapstructure:"enabled"`
+	ETCD              ETCD              `mapstructure:"etcd"`
+	APIServer         APIServer         `mapstructure:"apiServer"`
+	ControllerManager ControllerManager `mapstructure:"controllerManager"`
+	Scheduler         Scheduler         `mapstructure:"scheduler"`
```
gsanchezgavier, line 51: just as a reminder to refator this to be al `ControllerManager`

internal/config/config.go
gsanchezgavier: I can refactor this in my pr if we decided to change it

gsanchezgavier (approved):
LGTM just a reminder and a question regarding activating the control plane in the config.

e2e/e2e-values.yml
paologallinaharbur: there are node selectors, therefore the cp scraper is not deployed to the not-master nodes

PR #287, opened 2021-12-15
New Architecture image

alvarocabanas:
Creation of the new image in the repo https://hub.docker.com/r/newrelic/nri-kubernetes.
Modification of the agent to use the infrastructure-bundle directly


e2e/charts/newrelic-infrastructure/templates/daemonset-kubelet.yaml
```diff
@@ -80,6 +80,8 @@ spec:
           {{- end }}
         - name: agent
           image: {{ .Values.agentImage.repository }}:{{ .Values.agentImage.tag }}
+          command: [ "/sbin/tini" ]
+          args: [ "--", "/usr/bin/newrelic-infra" ]
```
roobre, line 84: Mostly a minor thing, but I would just mash all this together in the `command` line (which is the equivalent of `ENTRYPOINT`) rather than having both `command` and `args`

e2e/charts/newrelic-infrastructure/templates/daemonset-kubelet.yaml
```diff
@@ -98,6 +100,10 @@ spec:
                 fieldRef:
                   apiVersion: "v1"
                   fieldPath: "spec.nodeName"
+            - name: "NRIA_FEATURES"
+              value: "docker_enabled:false"
```
roobre, line 104: Aw yess, I love this!

Maybe we could add a small comment so we don't forget why we do this :)

e2e/e2e-values.yml
```diff
@@ -3,9 +3,9 @@ forwarderImage:
   tag: 1.20.5
   repository: newrelic/k8s-events-forwarder
 agentImage:
-  pullPolicy: Never
-  tag: e2e
-  repository: e2e/newrelic-infrastructure
```
roobre, line 8: I think these were meant to be unpullable since we "inject" this image as a previous step for the e2e test, but I'm not sure if this is still the case. Perhaps @paologallinaharbur knows more about if this is still used.

.github/workflows/release_beta.yml
```diff
@@ -0,0 +1,109 @@
+name: Release
```
roobre, line 1: Is there any reason to add a new `-beta` workflow, rather than just replacing the old one?

These should be versioned so a release from the v2 branch should use the workflow present on that branch.

e2e/e2e-values.yml
paologallinaharbur: Those were unpullable since we were building them and otherwise minikube tries to find them -failing-.
However, if the image is not build we can simply get it from the repo and use it

e2e/charts/newrelic-infrastructure/templates/daemonset-kubelet.yaml
gsanchezgavier: couldn't this be a constraint if the future if we want to tweak that FF from the backend? I think local FF takes precedence over remote ones

roobre: I think it could be hard to do this kind of detection from the backend, meaning: disabling docker `if kubernetes && (unprivileged || !processMetricsEnabled)`. Seems like a complex logic for the backend to handle.

e2e/charts/newrelic-infrastructure/templates/daemonset-kubelet.yaml
```diff
@@ -98,6 +100,10 @@ spec:
                 fieldRef:
                   apiVersion: "v1"
                   fieldPath: "spec.nodeName"
+            - name: "NRIA_FEATURES"
```
roobre, line 103: Looking at this again makes me want to confirm: Does this disable the docker integration (`nri-docker`), or the docker plugin in the agent related to correlating process data and containers?
We'd like to disable both.

e2e/charts/newrelic-infrastructure/templates/daemonset-kubelet.yaml
gsanchezgavier: sure, what i meant is that we are going to override a parameter that the backend could (if needed for some reason) modify to manage ingestion. But actually we are disabling the feature so it wouldn't make sense that the backend enable this again. Just discard my comment :)

paologallinaharbur: I am not a super fan of this since I have the feeling that we should do the very same of nri-ecs:
```
integrations:
  - name: nri-ecs
    when:
      env_exists:
        ENABLE_NRI_ECS: "true"
    interval: 15s
```
However, if caos tell us it is safe I am fine with that 🙃

e2e/charts/newrelic-infrastructure/templates/daemonset-kubelet.yaml
alvarocabanas: Only the integration.

.github/workflows/release_beta.yml
alvarocabanas: yes, true, changed.

e2e/charts/newrelic-infrastructure/templates/daemonset-kubelet.yaml
alvarocabanas: done

paologallinaharbur (approved):
thanks for taking care of this!

PR #289, opened 2021-12-16
Control plane authenticator

gsanchezgavier:
- Creates secret and pod `Listerer` that can be initialized with multiple namespaces and removes this logic from the scraper. Uses the secret lister informer in order to fetch the certificates for mTLS configuration.

- Creates the Authenticator responsable of generate the round tripper for the client based on the endpoint authorization configuration.

- Separates client test according to new structure.

- Codespell now uses local config file and there is a make target to run it (manual installation though)

### DIFF
--- a/.github/workflows/push_pr.yaml
+++ b/.github/workflows/push_pr.yaml
@@ -75,8 +75,3 @@ jobs:
       - uses: actions/checkout@v2
       - name: Codespell test
         uses: codespell-project/actions-codespell@master
-        with:
-          skip: ./.git
-          check_filenames: true
-          check_hidden: true
-          ignore_words_list: inout
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -81,6 +81,14 @@ type MTLS struct {
 	TLSSecretNamespace string `mapstructure:"secretNamespace"`
 }
 
+func (m MTLS) SecretNamespace() string {
+	if m.TLSSecretNamespace == "" {
+		return "default"
+	}
+
+	return m.TLSSecretNamespace
+}
+
 func LoadConfig(filePath string, fileName string) (*Config, error) {
 	v := viper.New()
 
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -81,14 +81,6 @@ type MTLS struct {
 	TLSSecretNamespace string `mapstructure:"secretNamespace"`
 }
 
-func (m MTLS) SecretNamespace() string {
-	if m.TLSSecretNamespace == "" {
-		return "default"
-	}
-
-	return m.TLSSecretNamespace
-}
-
 func LoadConfig(filePath string, fileName string) (*Config, error) {
 	v := viper.New()
 
--- a/internal/discovery/secret_lister.go
+++ b/internal/discovery/secret_lister.go
@@ -6,27 +6,46 @@ import (
 	listersv1 "k8s.io/client-go/listers/core/v1"
 )
 
+type SecretListerer interface {
+	Lister(namespace string) (listersv1.SecretNamespaceLister, bool)
+}
+
 type SecretListerConfig struct {
 	// Namespace can be used to restric the search to a particular namespace.
-	Namespace string
+	Namespaces []string
 	// Client is the Kubernetes client.Interface used to build informers.
 	Client kubernetes.Interface
 }
 
+type MultiNamespaceSecretListerer struct {
+	listers map[string]listersv1.SecretNamespaceLister
+}
+
+func (l MultiNamespaceSecretListerer) Lister(namespace string) (listersv1.SecretNamespaceLister, bool) {
+	lister, ok := l.listers[namespace]
+
+	return lister, ok
+}
+
 // NewSecretNamespaceLister returns a SecretGetter to get secrets with informers.
-func NewSecretNamespaceLister(config SecretListerConfig) (listersv1.SecretNamespaceLister, chan<- struct{}) {
+func NewSecretNamespaceLister(config SecretListerConfig) (*MultiNamespaceSecretListerer, chan<- struct{}) {
 	stopCh := make(chan struct{})
 
 	factory := informers.NewSharedInformerFactoryWithOptions(
 		config.Client,
 		defaultResyncDuration,
-		informers.WithNamespace(config.Namespace),
 	)
 
-	secretLister := factory.Core().V1().Secrets().Lister().Secrets(config.Namespace)
+	multiNamespaceSecretListerer := &MultiNamespaceSecretListerer{
+		listers: make(map[string]listersv1.SecretNamespaceLister),
+	}
+
+	for _, namespace := range config.Namespaces {
+		multiNamespaceSecretListerer.listers[namespace] = factory.Core().V1().Secrets().Lister().Secrets(namespace)
+	}
 
 	factory.Start(stopCh)
 	factory.WaitForCacheSync(stopCh)
 
-	return secretLister, stopCh
+	return multiNamespaceSecretListerer, stopCh
 }
--- a/internal/discovery/secret_lister.go
+++ b/internal/discovery/secret_lister.go
@@ -6,29 +6,36 @@ import (
 	listersv1 "k8s.io/client-go/listers/core/v1"
 )
 
+// SecretListerer return namespaced secret listers.
 type SecretListerer interface {
+	// Lister ruturns the secret lister for the specified namespaces
+	// and true if the lister exist in the listerer.
 	Lister(namespace string) (listersv1.SecretNamespaceLister, bool)
 }
 
-type SecretListerConfig struct {
-	// Namespace can be used to restric the search to a particular namespace.
+type SecretListererConfig struct {
+	// Namespaces supported by the listerer.
 	Namespaces []string
 	// Client is the Kubernetes client.Interface used to build informers.
 	Client kubernetes.Interface
 }
 
+// MultiNamespaceSecretListerer impelements SecretListerer interface
+// for a group of listers pre-build on initialization.
 type MultiNamespaceSecretListerer struct {
 	listers map[string]listersv1.SecretNamespaceLister
 }
 
+// Lister returns the available lister based on the namespace if exists in the listerer.
 func (l MultiNamespaceSecretListerer) Lister(namespace string) (listersv1.SecretNamespaceLister, bool) {
 	lister, ok := l.listers[namespace]
 
 	return lister, ok
 }
 
-// NewSecretNamespaceLister returns a SecretGetter to get secrets with informers.
-func NewSecretNamespaceLister(config SecretListerConfig) (*MultiNamespaceSecretListerer, chan<- struct{}) {
+// NewNamespaceSecretListerer returns a MultiNamespaceSecretListerer with listers for all
+// namespaces on config.Namespaces.
+func NewNamespaceSecretListerer(config SecretListererConfig) (*MultiNamespaceSecretListerer, chan<- struct{}) {
 	stopCh := make(chan struct{})
 
 	multiNamespaceSecretListerer := &MultiNamespaceSecretListerer{
--- a/internal/discovery/secret_lister.go
+++ b/internal/discovery/secret_lister.go
@@ -20,7 +20,7 @@ type SecretListererConfig struct {
 	Client kubernetes.Interface
 }
 
-// MultiNamespaceSecretListerer impelements SecretListerer interface
+// MultiNamespaceSecretListerer implements SecretListerer interface
 // for a group of listers pre-build on initialization.
 type MultiNamespaceSecretListerer struct {
 	listers map[string]listersv1.SecretNamespaceLister
--- a/internal/discovery/secret_lister.go
+++ b/internal/discovery/secret_lister.go
@@ -31,21 +31,22 @@ func (l MultiNamespaceSecretListerer) Lister(namespace string) (listersv1.Secret
 func NewSecretNamespaceLister(config SecretListerConfig) (*MultiNamespaceSecretListerer, chan<- struct{}) {
 	stopCh := make(chan struct{})
 
-	factory := informers.NewSharedInformerFactoryWithOptions(
-		config.Client,
-		defaultResyncDuration,
-	)
-
 	multiNamespaceSecretListerer := &MultiNamespaceSecretListerer{
 		listers: make(map[string]listersv1.SecretNamespaceLister),
 	}
 
 	for _, namespace := range config.Namespaces {
-		multiNamespaceSecretListerer.listers[namespace] = factory.Core().V1().Secrets().Lister().Secrets(namespace)
-	}
+		factory := informers.NewSharedInformerFactoryWithOptions(
+			config.Client,
+			defaultResyncDuration,
+			informers.WithNamespace(namespace),
+		)
 
-	factory.Start(stopCh)
-	factory.WaitForCacheSync(stopCh)
+		multiNamespaceSecretListerer.listers[namespace] = factory.Core().V1().Secrets().Lister().Secrets(namespace)
+
+		factory.Start(stopCh)
+		factory.WaitForCacheSync(stopCh)
+	}
 
 	return multiNamespaceSecretListerer, stopCh
 }
--- a/src/controlplane/client/authenticator.go
+++ b/src/controlplane/client/authenticator.go
@@ -1,0 +1,133 @@
+package client
+
+import (
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/newrelic/infra-integrations-sdk/log"
+	"github.com/newrelic/nri-kubernetes/v2/internal/config"
+	v1 "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/transport"
+)
+
+// Authenticator provides an interface to generate a authorized round tripper.
+type Authenticator interface {
+	Authenticate(endpoint config.Endpoint) (http.RoundTripper, error)
+}
+
+type authenticator struct {
+	logger                  log.Logger
+	secretListerByNamespace map[string]v1.SecretNamespaceLister
+	inClusterConfig         *rest.Config
+}
+
+// NewAuthenticator returns an Authenticator that supports plain, bearer token and mTLS.
+func NewAuthenticator(logger log.Logger, secretListerByNamespace map[string]v1.SecretNamespaceLister, inClusterConfig *rest.Config) Authenticator {
+	return &authenticator{
+		logger:                  logger,
+		secretListerByNamespace: secretListerByNamespace,
+		inClusterConfig:         inClusterConfig,
+	}
+}
+
+// Authenticate retruns a round tripper according to the endpoint config.
+// For mTLS configuration it fetches the certificates from the secret.
+func (a authenticator) Authenticate(endpoint config.Endpoint) (http.RoundTripper, error) {
+	transportConfig := &transport.Config{}
+	tlsConfig := transport.TLSConfig{
+		Insecure: endpoint.InsecureSkipVerify,
+	}
+
+	switch {
+	case endpoint.Auth == nil:
+		a.logger.Debugf("No authentication configured for %q, connection will be attempted anonymously", endpoint.URL)
+
+	case strings.EqualFold(endpoint.Auth.Type, bearerAuth):
+		a.logger.Debugf("Using kubernetes token to authenticate request to %q", endpoint.URL)
+
+		transportConfig.BearerToken = a.inClusterConfig.BearerToken
+
+	case strings.EqualFold(endpoint.Auth.Type, mTLSAuth) && endpoint.Auth.MTLS != nil:
+		a.logger.Debugf("Using mTLS to authenticate request to %q", endpoint.URL)
+
+		certs, err := a.getTLSCertificatesFromSecret(endpoint.Auth.MTLS)
+		if err != nil {
+			return nil, fmt.Errorf("could not load TLS configuration: %w", err)
+		}
+
+		tlsConfig.CertData = certs.cert
+		tlsConfig.KeyData = certs.key
+		// CAData could be empty if insecureSkipVerify is true.
+		tlsConfig.CAData = certs.ca
+
+	default:
+		return nil, fmt.Errorf("unknown authorization type %q", endpoint.Auth.Type)
+	}
+
+	transportConfig.TLS = tlsConfig
+
+	rt, err := transport.New(transportConfig)
+	if err != nil {
+		return nil, fmt.Errorf("creating the round tripper: %w", err)
+	}
+
+	return rt, nil
+}
+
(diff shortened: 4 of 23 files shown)


src/controlplane/client/authenticator.go
```diff
@@ -0,0 +1,133 @@
...
+	logger                  log.Logger
+	secretListerByNamespace map[string]v1.SecretNamespaceLister
+	inClusterConfig         *rest.Config
+}
+
+// NewAuthenticator returns an Authenticator that supports plain, bearer token and mTLS.
+func NewAuthenticator(logger log.Logger, secretListerByNamespace map[string]v1.SecretNamespaceLister, inClusterConfig *rest.Config) Authenticator {
+	return &authenticator{
+		logger:                  logger,
+		secretListerByNamespace: secretListerByNamespace,
```
gsanchezgavier, line 30: I was thinking to move the creation of this map inside the authenticator, but i'm not sure if this kind of namespaced listers should already provide some sort of multi-namespace initialization (perhaps managed in the discovery package)

src/controlplane/client/connector.go
```diff
@@ -98,133 +98,3 @@ func (dp *defaultConnector) probeEndpoint(url string, client *http.Client) error
...
-
-	tlsConfig, err := parseTLSConfig(cert, key, cacert, insecureSkipVerify)
-	if err != nil {
-		return nil, fmt.Errorf("parsing TLS config: %w", err)
-	}
-
-	return tlsConfig, nil
-}
-
-func parseTLSConfig(certPEMBlock, keyPEMBlock, cacertPEMBlock []byte, insecureSkipVerify bool) (*tls.Config, error) {
```
gsanchezgavier, line 204: this was completely removed since k8s transport package handle this.

src/controlplane/client/connector.go
```diff
@@ -98,133 +98,3 @@ func (dp *defaultConnector) probeEndpoint(url string, client *http.Client) error
...
-	if cert, ok = secret.Data["cert"]; !ok {
-		return nil, fmt.Errorf("could not find TLS certificate in `cert` field in secret %q", secretName)
-	}
-
-	if key, ok = secret.Data["key"]; !ok {
-		return nil, fmt.Errorf("could not find TLS key in `key` field in secret %q", secretName)
-	}
-
-	cacert, hasCACert := secret.Data["cacert"]
-	insecureSkipVerifyRaw, hasInsecureSkipVerify := secret.Data["insecureSkipVerify"]
```
gsanchezgavier, line 184: the posibility to configure the ca verification has been removed from the secrets, now it needs to be configured in the endpoint configuration (as the others auth methods)

.github/workflows/push_pr.yaml
```diff
@@ -76,7 +76,7 @@ jobs:
       - name: Codespell test
         uses: codespell-project/actions-codespell@master
         with:
-          skip: ./.git
+          skip: ./.git,internal/testutil/data
```
paologallinaharbur, line 79: 🔝

src/controlplane/client/authenticator.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+type authenticator struct {
+	logger                  log.Logger
+	secretListerByNamespace map[string]v1.SecretNamespaceLister
+	inClusterConfig         *rest.Config
+}
+
+// NewAuthenticator returns an Authenticator that supports plain, bearer token and mTLS.
+func NewAuthenticator(logger log.Logger, secretListerByNamespace map[string]v1.SecretNamespaceLister, inClusterConfig *rest.Config) Authenticator {
```
roobre, line 27: Logger in the constructor list makes it look a bit boilerplate-y. Could we use an Options pattern, here?

roobre: Not sure if we have a reason to hide the K8s implementation for Authenticator. Would it be a problem to return it (exported) rather than the interface? Taking advantage of Go's fantastic implicit interfaces :)

roobre (commented):
Looking awesome! 💖 
Some minor nits & stuff.

.github/workflows/push_pr.yaml
gsanchezgavier: didn't work 😭

src/controlplane/scraper.go
```diff
@@ -99,8 +101,10 @@ func NewScraper(config *config.Config, providers Providers, options ...ScraperOp
 		return nil, fmt.Errorf("fetching K8s version: %w", err)
 	}
 
-	// Building pod lister and closers for pod autodisover.
-	s.buildLister()
+	// Building pod and secret lister and closers.
+	s.buildListers()
```
paologallinaharbur, line 105: This still confuses me a bit. I do not like that it saves the lister into the object in a hidden way, moreover the client doesn't need to save those object, since it merely passes them two lines below

.github/workflows/push_pr.yaml
roobre: Seems like codespell is ignoring it though: https://github.com/newrelic/nri-kubernetes/runs/4562163854?check_suite_focus=true#step:4:20
From the [docs](https://github.com/codespell-project/actions-codespell#parameter-skip) it says it expects a glob, perhaps we have to use that syntax?
```suggestion
          skip: ./.git,internal/testutil/data/**
```

src/controlplane/scraper.go
paologallinaharbur: (at least the secret lister), wouldn't be better to return the two and saving explicitely the podLister and passing by the secret one without saving it into the struct?

src/controlplane/client/authenticator.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/newrelic/nri-kubernetes/v2/internal/config"
+	v1 "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/transport"
+)
+
+// Authenticator provides an interface to generate a authorized round tripper.
+type Authenticator interface {
+	Authenticate(endpoint config.Endpoint) (http.RoundTripper, error)
+}
```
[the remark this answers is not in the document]
gsanchezgavier: I was not convinced either... the k8s client have something similar in a transport package , this is kind of a wrapper of that so `AuthenticatedTransport` make sense to me

src/controlplane/client/authenticator.go
```diff
@@ -0,0 +1,133 @@
...
+		a.logger.Debugf("Using mTLS to authenticate request to %q", endpoint.URL)
+
+		certs, err := a.getTLSCertificatesFromSecret(endpoint.Auth.MTLS)
+		if err != nil {
+			return nil, fmt.Errorf("could not load TLS configuration: %w", err)
+		}
+
+		tlsConfig.CertData = certs.cert
+		tlsConfig.KeyData = certs.key
+		// CAData could be empty if insecureSkipVerify is true.
```
paologallinaharbur, line 62: should we log something or erroring if it is empty and insecureSkipVerify is not set?

.github/workflows/push_pr.yaml
gsanchezgavier: i'll give a try

src/controlplane/client/authenticator.go
```diff
@@ -0,0 +1,133 @@
...
+		secretListerByNamespace: secretListerByNamespace,
+		inClusterConfig:         inClusterConfig,
+	}
+}
+
+// Authenticate retruns a round tripper according to the endpoint config.
+// For mTLS configuration it fetches the certificates from the secret.
+func (a authenticator) Authenticate(endpoint config.Endpoint) (http.RoundTripper, error) {
+	transportConfig := &transport.Config{}
+	tlsConfig := transport.TLSConfig{
```
paologallinaharbur, line 39: I would avoid a new variable, and instead I would place `transport.TLSConfig` directly inside the `transportConfig` so that in the cases we can modify the initial object and avoid `transportConfig.TLS = tlsConfig`.

I find it more readable, but maybe an other person would not

src/controlplane/client/authenticator.go
```diff
@@ -0,0 +1,133 @@
...
+		a.logger.Debugf("TLS Secret name configured, but not TLS Secret namespace. Defaulting to `default` namespace.")
+
+		namespace = DefaultSecretNamespace
+	}
+
+	var secretLister v1.SecretNamespaceLister
+
+	var ok bool
+
+	if secretLister, ok = a.secretListerByNamespace[namespace]; !ok {
```
paologallinaharbur, line 103: we are inlining this to save 1 line, but we need 2 to initialize the vars 😄

src/controlplane/client/client.go
```diff
@@ -47,10 +47,6 @@ func New(connector Connector, opts ...OptionFunc) (*Client, error) {
 		}
 	}
 
-	if connector == nil {
```
paologallinaharbur, line 50: why removing this check?

src/controlplane/client/connector.go
```diff
@@ -12,42 +9,41 @@ import (
...
 const (
-	DefaultTimout      = 5000 * time.Millisecond
-	defaultMetricsPath = "/metrics"
-	mTLSAuth           = "mTLS"
-	bearerAuth         = "bearer"
+	DefaultTimout          = 5000 * time.Millisecond
+	DefaultSecretNamespace = "default"
+	defaultMetricsPath     = "/metrics"
+	mTLSAuth               = "mTLS"
+	bearerAuth             = "bearer"
```
paologallinaharbur, line 20: I would move these two to the authenticator

src/controlplane/client/connector.go
paologallinaharbur: 🎉

src/controlplane/scraper.go
```diff
@@ -29,13 +29,15 @@ type Providers struct {
...
-	informerClosers      []chan<- struct{}
-	podListerByNamespace map[string]v1.PodNamespaceLister
-	inClusterConfig      *rest.Config
+	logger                  log.Logger
+	config                  *config.Config
+	k8sVersion              *version.Info
+	components              []component
+	informerClosers         []chan<- struct{}
+	podListerByNamespace    map[string]v1.PodNamespaceLister
+	secretListerByNamespace map[string]v1.SecretNamespaceLister
```
paologallinaharbur, line 38: I would not save this into the scraper if possible

paologallinaharbur (commented):
Nice job, few suggestions, but overall looks good! and it is way simpler the way certs are managed now

src/controlplane/client/authenticator.go
gsanchezgavier: what about modify the secret discovery to implement something like.
Perhaps just to simplify the thing just support array of namespaces at creation (not adding nor deleting)
```
type SecretListerer interface {
	Lister(namespace string) (listersv1.SecretNamespaceLister, error)
}
```
Then the authenticator accept this interface.
Probably the intitialization of this still could be handle by the `scraper` maybe moving the nasty loops to the `component` , this should unload the scraper i think.

gsanchezgavier: would be easier if this is implemented i guess https://github.com/kubernetes/kubernetes/issues/74415

.github/workflows/push_pr.yaml
```diff
@@ -76,7 +76,7 @@ jobs:
       - name: Codespell test
         uses: codespell-project/actions-codespell@master
         with:
-          skip: ./.git
+          skip: ./.git,internal/testutil/data/**
```
gsanchezgavier, line 79: not working either 😢

src/controlplane/client/authenticator.go
gsanchezgavier: just pushed a commit with this approach , LMK WDYT

src/controlplane/client/authenticator.go
```diff
@@ -0,0 +1,133 @@
...
+		return nil, fmt.Errorf("could not find secret lister for namespace %q", namespace)
+	}
```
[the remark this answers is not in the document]
gsanchezgavier: 🤦‍♂️  thanks for caching that

src/controlplane/client/authenticator.go
```diff
@@ -0,0 +1,133 @@
...
+	if cert, ok = secret.Data["cert"]; !ok {
+		return nil, fmt.Errorf("could not find TLS certificate in `cert` field in secret %q", mTLSConfig.TLSSecretName)
+	}
+
+	if key, ok = secret.Data["key"]; !ok {
+		return nil, fmt.Errorf("could not find TLS key in `key` field in secret %q", mTLSConfig.TLSSecretName)
+	}
+
+	if cacert, ok = secret.Data["cacert"]; !ok {
+		a.logger.Debugf("CA certificate is not present in secret %q on namespace %q", mTLSConfig.TLSSecretName, namespace)
```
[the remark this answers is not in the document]
gsanchezgavier: just moved the check outside this func. There is also a test case for this

src/controlplane/client/authenticator.go
gsanchezgavier: changed this and the connector to unify the code

src/controlplane/client/authenticator.go
gsanchezgavier: added the check here

src/controlplane/client/authenticator.go
gsanchezgavier: 🤦‍♂️  thanks for caching that

src/controlplane/scraper.go
gsanchezgavier: resolve this comment, is part of the discussion above

.github/workflows/push_pr.yaml
gsanchezgavier: working know , the action interface doesn't work as expected. I just added the local make target and `.codespellrc` with de config, so we can run locally and is "the same" as the ci runs

internal/config/config.go
```diff
@@ -81,6 +81,14 @@ type MTLS struct {
 	TLSSecretNamespace string `mapstructure:"secretNamespace"`
 }
 
+func (m MTLS) SecretNamespace() string {
```
paologallinaharbur, line 84: why not setting it as:
 - `v.SetDefault("clusterName", "cluster")`

Or in helm charts directly?

src/controlplane/client/connector/connector.go
```diff
@@ -30,27 +30,47 @@ type ConnParams struct {
...
+}
+
+// DefaultConnector implements Connector interface for the Control Plane components.
+type DefaultConnector struct {
+	Config
+	logger log.Logger
+}
+
+// New returns a DefaultConnector.
+func New(config Config, opts ...OptionFunc) (*DefaultConnector, error) {
```
paologallinaharbur, line 55: I am not a super fan of this, since the logic is basically the same as before, but more code is needed.

I imagine we introduce it since after moving to a new package structure

src/controlplane/client/connector/connector.go
```diff
@@ -30,27 +30,47 @@ type ConnParams struct {
 	Client client.HTTPDoer
 }
 
-type defaultConnector struct {
-	// TODO: Use a non-sdk logger
-	logger        log.Logger
-	authenticator authenticator.Authenticator
-	endpoints     []config.Endpoint
+type Config struct {
```
paologallinaharbur, line 33: could we add some comments since they are exported?

src/controlplane/client/connector/connector.go
paologallinaharbur: we ask to provide config, but then in the .New we simply save config and apply logger..

internal/discovery/secret_lister.go
```diff
@@ -6,27 +6,46 @@ import (
...
+
 // NewSecretNamespaceLister returns a SecretGetter to get secrets with informers.
-func NewSecretNamespaceLister(config SecretListerConfig) (listersv1.SecretNamespaceLister, chan<- struct{}) {
+func NewSecretNamespaceLister(config SecretListerConfig) (*MultiNamespaceSecretListerer, chan<- struct{}) {
 	stopCh := make(chan struct{})
 
 	factory := informers.NewSharedInformerFactoryWithOptions(
 		config.Client,
 		defaultResyncDuration,
-		informers.WithNamespace(config.Namespace),
```
paologallinaharbur, line 23: I fear that with this change the permission needed is to watch the secrets over the all clusters, before we needed such permission over the single namespace selected 😕  

The helm chart however is not smart enough to make any difference

paologallinaharbur: creating a `informers.NewSharedInformerFactoryWithOptions` per namespace would be a solution... would it worth the overhead?

src/controlplane/scraper.go
```diff
@@ -92,15 +94,35 @@ func NewScraper(config *config.Config, providers Providers, options ...ScraperOp
...
+			SecretListerer:  secretListerer,
+			InClusterConfig: s.inClusterConfig,
+		},
+		authenticator.WithLogger(s.logger),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("creating authenticator: %w", err)
+	}
+
+	// TODO refactor pod discovery to add multiple namespaces and implement similar
```
paologallinaharbur, line 123: would make sense to address it now?

src/controlplane/client/connector/connector.go
gsanchezgavier: yes I introduced in order to unify a bit the way of objects are created, didn't make much sense to have it different between authenticator and connector IMO. 
I personally like this approach , is easy to extend and to refactor when changes are introduced.

src/controlplane/scraper.go
gsanchezgavier: yes but i wanted to validate that the `SecretListerer` approach is ok

internal/discovery/secret_lister.go
gsanchezgavier: my understanding is that the only namespace that is going to be get,listed,watched is the one we create a lister for (few lines below).
I still need to add some test to this , I can test that too.

internal/config/config.go
gsanchezgavier: thanks yes i need to get rid of this, we decided that we are not going to use this default

src/controlplane/client/connector/connector.go
gsanchezgavier: i think the commets will add more noice in this case, i think the variable names are clear self explained wdyt?

internal/discovery/secret_lister.go
gsanchezgavier: yes your fear was confirmed 😢 is not working as I expected. 
I need to build an informer/namespace i think

gsanchezgavier: added some test and include the multi factory thing , it doesn't look to complex in the code and we could use the same closer at least

internal/discovery/secret_lister.go
```diff
@@ -6,27 +6,47 @@ import (
 	listersv1 "k8s.io/client-go/listers/core/v1"
 )
 
+type SecretListerer interface {
```
roobre, line 9: The `Listerer` name seems slightly strange to me. Seeing that this interface seems very similar to https://pkg.go.dev/k8s.io/client-go/listers/core/v1#SecretLister, perhaps we could borrow the names used upstream

internal/discovery/secret_lister.go
```diff
@@ -6,27 +6,47 @@ import (
...
+	Namespaces []string
 	// Client is the Kubernetes client.Interface used to build informers.
 	Client kubernetes.Interface
 }
 
+type MultiNamespaceSecretListerer struct {
+	listers map[string]listersv1.SecretNamespaceLister
+}
+
+func (l MultiNamespaceSecretListerer) Lister(namespace string) (listersv1.SecretNamespaceLister, bool) {
```
roobre, line 24: Right now we are creating a collection of listers, indexed by namespace, and returning it. That doesn't sound bad, but I wonder if returning a listerer capable of creating listers lazily would have any advantages.

This is, if we make this method to first check if there is a lister already for a naespace, and create it if there isn't.

WDYT about this approach? I'm not positive it's a better option, just wondering about it.

src/controlplane/client/connector/connector.go
```diff
@@ -0,0 +1,121 @@
...
+
+		return &ConnParams{URL: *u, Client: httpClient}, nil
+	}
+
+	return nil, fmt.Errorf("all endpoints in the list failed to response")
+}
+
+// probeEndpoint executes a HEAD request to the url and fails if the response code
+// is not StatusOK.
+func (dp *DefaultConnector) probeEndpoint(url string, client *http.Client) error {
```
roobre, line 108: I think this name could be shorter if we do something like this:
```suggestion
func (dp *DefaultConnector) probe(endpoint string, client *http.Client) error {
```

src/controlplane/client/authenticator/authenticator.go
```diff
@@ -0,0 +1,154 @@
...
+
+	"github.com/newrelic/infra-integrations-sdk/log"
+	"github.com/newrelic/nri-kubernetes/v2/internal/config"
+	"github.com/newrelic/nri-kubernetes/v2/internal/discovery"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/transport"
+)
+
+const (
+	DefaultSecretNamespace = "default"
```
paologallinaharbur, line 16: this is not used anymore

src/controlplane/client/connector/connector.go
paologallinaharbur: mmmm true :)

internal/discovery/secret_lister.go
gsanchezgavier: this was the most simple solution i imaged ( not even mutex for the map :) ) since we have all namespaces in the config ( no need to dynamically build listers ), and unless there is use case for that I would keep it like this.

internal/discovery/secret_lister.go
gsanchezgavier: I'm not convinced either of the name but because `listerer` is not a valid word and my spell checker in vscode complains :).
I thought about it but in the case of this interface it returns lister , it doesn't list . I would expect that a lister list so could be confusing i think.

src/controlplane/client/authenticator/authenticator.go
gsanchezgavier: damm it we need linters to be a bit more strict !

src/controlplane/scraper.go
gsanchezgavier: just pushed this change

paologallinaharbur (approved):
Everything looks good!

internal/discovery/secret_lister.go
roobre: I guess the authors of go-client faced a similar issue. I think the way they resolved it is by making the top-level `Lister` not having a `List` method, but rather a `Secrets` method that returns the real `Lister`.

Might be confusing but if we cnanot thing about a better name I'd just do what they did :)

Moreover if we name the method the same way, this means their object will also satisfy our interface. Which I'm not sure if it will be useful at any point but who knows :)

roobre (approved):
LGTM, only thing I'd change is the name of the master lister :)

PR #292, opened 2021-12-21
Fix verbose setting a default

paologallinaharbur:
It was not possible to set verbose as Env Var since we needed to set in in Viper first

Adding networkRouteFile as well


alvarocabanas (approved):
LGTM

paologallinaharbur:
code spell failing for an already existing issue

PR #294, opened 2021-12-22
Control plane Discoverer

### LINKED ISSUE #241: Make discoverer easier to test
At the moment `Discoverer` interface https://github.com/newrelic/nri-kubernetes/blob/0a78aa518374cc8d250122d26268ba5e1858a754/src/client/discovery.go#L8-L11 returns an `HTTPClient` interface https://github.com/newrelic/nri-kubernetes/blob/0a78aa518374cc8d250122d26268ba5e1858a754/src/client/discovery.go#L20-L24 which makes it difficult to test, as it is obfuscated where the request will be sent.

Most likely we should decompose `Discoverer` interface to make it have an endpoint discovery part and client configuration part, so we can only test discovered endpoints without caring about client configuration and performing the requests.

gsanchezgavier:
Moves pod discover logic to Discoverer, and add some corner cases testing.
Fixes #241

### DIFF
--- a/src/controlplane/discoverer/discoverer.go
+++ b/src/controlplane/discoverer/discoverer.go
@@ -1,0 +1,96 @@
+package discoverer
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/newrelic/infra-integrations-sdk/log"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+
+	"github.com/newrelic/nri-kubernetes/v2/internal/config"
+	"github.com/newrelic/nri-kubernetes/v2/internal/discovery"
+)
+
+var ErrPodNotFound = errors.New("pod not found")
+
+// PodDiscoverer is used to discover control plane components.
+type PodDiscoverer interface {
+	// Discover returns a pod mathching the selector, namespaces and
+	// is in the same node if matchNode is true.
+	Discover(config.AutodiscoverControlPlane) (*corev1.Pod, error)
+}
+
+type Config struct {
+	PodListerer discovery.PodListerer
+	NodeName    string
+}
+
+type OptionFunc func(c *ControlplanePodDiscoverer) error
+
+// WithLogger returns an OptionFunc to change the logger from the default noop logger.
+func WithLogger(logger log.Logger) OptionFunc {
+	return func(c *ControlplanePodDiscoverer) error {
+		c.logger = logger
+		return nil
+	}
+}
+
+// ControlplanePodDiscoverer implements PodDiscoverer interface.
+type ControlplanePodDiscoverer struct {
+	Config
+	logger log.Logger
+}
+
+// New returns an ControlplanePodDiscoverer.
+func New(config Config, opts ...OptionFunc) (*ControlplanePodDiscoverer, error) {
+	c := &ControlplanePodDiscoverer{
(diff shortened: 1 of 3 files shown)


src/controlplane/discoverer/discoverer_test.go
```diff
@@ -0,0 +1,193 @@
...
+)
+
+func Test_Discoverer_does_not_fail(t *testing.T) {
+	t.Parallel()
+
+	namespace := "testNamespace"
+	nodeName := "testNode"
+	selector := "foo=bar"
+
+	testCases := []struct {
```
paologallinaharbur, line 27: I might missed it, but I would write a tests and add a comments saying what happens if the list of selector provided is empty

paologallinaharbur (approved):
Only a doubt regarding a behaviour I am not sure we want to allow permit, everything else looks good!

src/controlplane/discoverer/discoverer.go
```diff
@@ -0,0 +1,93 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+
+	"github.com/newrelic/nri-kubernetes/v2/internal/config"
+	"github.com/newrelic/nri-kubernetes/v2/internal/discovery"
+)
+
+// PodDiscoverer is used to discover control plane components.
+type PodDiscoverer interface {
+	// Discover returns looks for a pod mathching the selector, namespaces and
```
roobre, line 16: returns or looks?

src/controlplane/discoverer/discoverer.go
```diff
@@ -0,0 +1,93 @@
...
+	labelsSet, err := labels.ConvertSelectorToLabelsMap(ad.Selector)
+	if err != nil {
+		return nil, fmt.Errorf("invalid selector %q: %w", ad.Selector, err)
+	}
+
+	selector := labels.SelectorFromSet(labelsSet)
+
+	pods, err := podLister.List(selector)
+	if err != nil {
+		return nil, fmt.Errorf("fail to list pods for selector: %v", labelsSet)
```
roobre, line 78:
```suggestion
		return nil, fmt.Errorf("listing pods with selector %q: %w", labelsSet, err)
```

src/controlplane/discoverer/discoverer.go
```diff
@@ -0,0 +1,93 @@
...
+	for _, pod := range pods {
+		if ad.MatchNode && pod.Spec.NodeName != c.Config.NodeName {
+			c.logger.Debugf("Discarding pod: %s running outside the node", pod.Name)
+			continue
+		}
+		// first pod matching all conditions is returned.
+		return pod, nil
+	}
+
+	return nil, nil
```
roobre, line 92: Should we return a non-nil error here? Like a `ErrNoPodsFound`.

roobre (approved):
Looking great so far! Just a couple nits

src/controlplane/discoverer/discoverer.go
gsanchezgavier: yes it make sense! at least just to have this custom error 👍

src/controlplane/discoverer/discoverer_test.go
gsanchezgavier: just added on "when_empty_selector", thanks

PR #295, opened 2021-12-22
testutil/asserter: improve composability of exclude functions

roobre:
This PR makes the `ExcludeFunc`s included with the `asserter` smaller and more composable. This is done by means of a very simple ExcludeFunc:

```go
// Exclude returns an ExcludeFunc that returns true if all the supplied ExcludeFuncs return true.
// Input ExcludeFuncs are evaluated in order, so this function makes easy to compose exclusion rules.
func Exclude(funcs ...ExcludeFunc) ExcludeFunc {
```

Which allow to chain other excludeFuncs:
```go
exclude.Exclude(
	// for pods...
	exclude.Groups("pod"),
	// that are waiting...
	func(_ string, _ *definition.Spec, ent *integration.Entity) bool {
		return asserter.EntityMetricIs(ent, "status", "waiting")
	},
	// exclude metrics defined in this `metricsToExcludeForPendingPods`
	exclude.Metrics(metricsToExcludeForPendingPods...),
),
```

Achieving complex things without the need to write that much boilerplate.

### DIFF
--- a/internal/testutil/asserter/assert.go
+++ b/internal/testutil/asserter/assert.go
@@ -1,4 +1,4 @@
-package testutil
+package asserter
 
 import (
 	"strings"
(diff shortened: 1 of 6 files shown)


src/kubelet/kubelet_test.go
```diff
@@ -24,18 +28,100 @@ import (
...
+
+			// Exclude PVC metrics for volumes that are not named "pv"
+			exclude.Exclude(
+				exclude.Groups("volume"),
+				func(_ string, spec *definition.Spec, e *integration.Entity) bool {
+					return e.Metadata.Name != "pv" && strings.HasPrefix(spec.Name, "pvc")
+				},
+			),
+			// Optional metrics.
+			//exclude.Optional(),
```
paologallinaharbur, line 123: leftovers

src/kubelet/kubelet_test.go
```diff
@@ -24,18 +28,100 @@ import (
...
 		Excluding(
-			testutil.ExcludeMetrics("pod", commonMetricsToExclude...),
-			testutil.ExcludeMetrics("node", nodeMetricsToExclude...),
-			testutil.ExcludeOptional(),
-		).Excluding(ExcludeMissingMetricsPendingPod)
+			// Network metrics
+			exclude.Metrics(networkMetrics...),
+
+			// Node utilization  metrics.
+			exclude.Exclude(exclude.Groups("node"), exclude.Metrics(nodeUtilizationMetrics...)),
```
paologallinaharbur, line 59: I would add an helper function as getKubeletAsserter, just to divide logically some code and keep the test readable

roobre: Addressed in https://github.com/newrelic/nri-kubernetes/pull/303/commits/dfdd7d9b14c380399cf3fe3106814b62236ba23c

internal/testutil/asserter/exclude/exclude.go
```diff
@@ -0,0 +1,97 @@
...
+				return true
+			}
+		}
+
+		return false
+	}
+}
+
+// MetricsGroup returns a Func that excludes the specified metric names belonging for the specified group.
+// Deprecated: Use Exclude(Groups("group"), Metrics("...")) instead.
```
paologallinaharbur, line 64: Does it makes sense to keep a Deprecated function in tests? 
I would try at least to get rid of it in https://github.com/newrelic/nri-kubernetes/pull/303/files

internal/testutil/asserter/assert.go
```diff
@@ -175,32 +150,44 @@ func entitiesFor(entities []*integration.Entity, pseudotype string) []*integrati
...
-		}
-
-		if !isString && mType == metric.ATTRIBUTE {
-			continue
-		}
+	return entityMetric(e, m) == value
+}
 
+// EntityContains returns true if supplied entity has metric m with type _similar_ to mType, false otherwise.
+func EntityContains(e *integration.Entity, m string, mType metric.SourceType) bool {
```
paologallinaharbur, line 177: I do not like particularly this name, EntityContains since it is actually checking if the mType ~matches the value type

internal/testutil/asserter/assert.go
```diff
@@ -175,32 +150,44 @@ func entitiesFor(entities []*integration.Entity, pseudotype string) []*integrati
...
+		if found {
+			return entityMetric
+		}
+	}
+
+	return nil
+}
+
+// EntityHas returns true if the specified entity has a metric of name m equal to value.
+func EntityHas(e *integration.Entity, m string, value interface{}) bool {
```
paologallinaharbur, line 166: Whats about `entityMetricWithWildcard`? Just a suggestion, feel free to ignore it :)

paologallinaharbur (approved):

internal/testutil/asserter/exclude/exclude.go
roobre: I marked it as such to avoid modifying other tests in the same PR, but I can fix them as well :)

internal/testutil/asserter/assert.go
roobre: I'm not very convinced either about the value of those two funcs 🤔 

WDYT about `EntityHasValue` and `EntityHasType`?

roobre: A better idea: `EntityMetricIs` and `EntityMetricKindIs`

paologallinaharbur: I like them all no bit preference 😄

internal/testutil/asserter/exclude/exclude.go
paologallinaharbur: it should be a quick fix and without any logic change, no?

internal/testutil/asserter/assert.go
roobre: I'll go for the last one then :)

PR #282, opened 2021-11-30
Update tests to new testdata format introduced in #281

roobre:
Built on top of #281

Fake K8s object retrievers from the `testutil` package have been slightly modified to return objects by version, rather than sharing the same for all of them. This aligns better with the testdata tree. KSM and Kubelet scraper tests have been migrated accordingly.

Finally, during these tests, some bugs have been fixed and squashed, related to a missing `namespaceName` metric in HPA sample, and missing `LoadBalancerIP` in ServiceSample.

PR #296, opened 2021-12-22
fix loadBalancerIP for services and namespaces for HPA

roobre:
During the development of #282, we noticed the `loadBalancerIP` metric of the `K8sServiceSample` and `namespace` from `K8sHpaSample` were not being populated.

For the first case, this was being caused by attempting to get the `ip` label from the `kube_service_status_load_balancer_ingress` metric (from KSM). This metric will not be present if an ingress is not involved with the loadBalancer, so the label was not retrieved.

In the second case the root cause was very similar, retrieving the `namespace` label from the `kube_hpa_status_condition` metric which will not always be present.

This PR simply changes the source metric for those two attributes.

### DIFF
--- a/src/metric/definition.go
+++ b/src/metric/definition.go
@@ -600,7 +600,7 @@ var KSMSpecs = definition.SpecGroups{
 			},
 			{
 				Name:      "loadBalancerIP",
-				ValueFunc: prometheus.FromLabelValue("kube_service_status_load_balancer_ingress", "ip"),
+				ValueFunc: prometheus.FromLabelValue("kube_service_info", "load_balancer_ip"),
 				Type:      sdkMetric.ATTRIBUTE,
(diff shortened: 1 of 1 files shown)


roobre:
It would be great if @github decided to run the CI for this PR

gsanchezgavier:
@roobre  the change looks good !!, but I don't have context. What was the issue with this metric?

paologallinaharbur:
Same as guillermo, I would add 1 line for the future ourselves in the PR description

roobre:
Sorry about that, added some context to the PR now :)

gsanchezgavier (approved):
thanks!!

gsanchezgavier (approved):
Not sure why GHA don't want to run. Just check what happens on mergin to develop

PR #297, opened 2021-12-24
remove unused storage code

gsanchezgavier:
`storage` was used to store the cache data for components that were had cache mechanism (network and controlplane client). All cache is now handled by informers and this code is unused.


roobre (approved):
🔥

PR #303, opened 2021-12-29
kubelet/test: Improve integration tests

roobre:
Kubelet tests used to use `exclude.Optional()` to avoid failing when "optional" metrics were not found. This however can potentially hide bugs in the code that produces those metrics, which would be hidden from the test.

This PR gets rid of `exclude.Optional` and instead adds a bunch of organic ignore rules.


src/kubelet/kubelet_test.go
```diff
@@ -85,19 +79,99 @@ func TestScraper(t *testing.T) {
...
+
+		// Exclude limits/requested metrics for nodes, pods and containers
+		exclude.Exclude(
+			exclude.Groups("node", "pod", "container"),
+			func(group string, spec *definition.Spec, e *integration.Entity) bool {
+				return limitsRequestsRegex.MatchString(spec.Name)
+			},
+		),
+		// Exclude metrics that depend on limits when those limits are not set.
+		exclude.Exclude(exclude.Groups("pod", "container"), exclude.Dependent(utilizationDependencies)),
```
paologallinaharbur, line 122: why excluding whole groups with ".Dependent" and not specific metrics?

paologallinaharbur: Ahhh I got it, It exclude everything from the group having those dependencies.. smart 😄

src/kubelet/kubelet_test.go
```diff
@@ -26,19 +28,11 @@ import (
 )
 
 func TestScraper(t *testing.T) {
-
-	commonMetricsToExclude := []string{"net.rxBytesPerSecond", "net.txBytesPerSecond", "net.errorsPerSecond"}
-	nodeMetricsToExclude := append(commonMetricsToExclude, "allocatableCpuCoresUtilization", "allocatableMemoryUtilization")
-
 	// Create an asserter with the settings that are shared for all test scenarios.
 	asserter := asserter.New().
+		Silently().
```
paologallinaharbur, line 33: I am happy to see that, but I fear it is a leftover. Sure you do not want to keep it?

src/kubelet/kubelet_test.go
```diff
@@ -85,19 +79,99 @@ func TestScraper(t *testing.T) {
...
-func ExcludeMissingMetricsPendingPod(_ string, spec *definition.Spec, ent *integration.Entity) bool {
-	metricsToExcludeForPendingPods := []string{"memoryUsedBytes", "memoryWorkingSetBytes", "cpuUsedCores",
+// kubeletExclusions is a helper that returns all the exclusions needed to assert the kubelet metrics without getting
+// false negatives.
+func kubeletExclusions() []exclude.Func {
+	// Network metrics are known to be missing on some environments.
+	networkMetrics := []string{"net.rxBytesPerSecond", "net.txBytesPerSecond", "net.errorsPerSecond"}
+
+	// TODO: Unclear why we need to exclude node utilization metrics.
+	nodeUtilizationMetrics := []string{"allocatableCpuCoresUtilization", "allocatableMemoryUtilization"}
```
paologallinaharbur, line 89: It is not unclear anymore, I would open an issue and link it in the TODO

paologallinaharbur (approved):
some nits, but thanks for taking the time to investigate those

src/kubelet/kubelet_test.go
```diff
@@ -85,19 +79,99 @@ func TestScraper(t *testing.T) {
...
+		// Static pods, typically living in kube-system, do not have creation dates.
+		exclude.Exclude(
+			exclude.Groups("pod"),
+			func(_ string, _ *definition.Spec, ent *integration.Entity) bool {
+				return asserter.EntityHas(ent, "namespace", "kube-system")
+			},
+			exclude.Metrics("createdAt", "createdBy", "createdKind", "deploymentName"),
+		),
+
+		// Exclude deploymentName metric for pods not created by a deployment
```
paologallinaharbur, line 133: should we create an issue to add also DaemonSetName, JobName, etc?

paologallinaharbur: it could be really useful for entity relations

src/kubelet/kubelet_test.go
roobre: Hahah it is 100% a leftover :P

src/kubelet/kubelet_test.go
roobre: I need to doublecheck if node metrics are also affected by this, or did you check it already? I've only confirmed the bug for containers/pods.

paologallinaharbur: I guess they are since they are based as well on computed ones

roobre:
Removing `ci/skip-e2e` as we have made some changes to the data being reported in #319

PR #305, opened 2021-12-29
main: use in-memory store for integration

roobre:
Since the integration is now long-running, it no longer needs to persist metrics for delta calculations or other tasks.
This removes the need to mount emptyDir volumes on `/tmp` for the integration container.

Open questions:
- ~Is `InMemoryStore` production ready, or meant for testing purposes only?~
- Will it behave properly in the long term, or will it leak resources?

### DIFF
--- a/cmd/kubernetes-static/main.go
+++ b/cmd/kubernetes-static/main.go
@@ -57,7 +57,7 @@ func main() {
 
 	fakeK8s := fake.NewSimpleClientset(k8sData.Everything()...)
 
-	i, err := integration.New(integrationName, integrationVersion, integration.Args(&args))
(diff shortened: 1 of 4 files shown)


paologallinaharbur:
The change looks good but the open questions scares me, I am not 100% sure that wasn't merely a feature needed for tests

roobre:
> I am not 100% sure that wasn't merely a feature needed for tests

I think that part should be fine, as the on-disk storage actually uses the inMemoryStore. I'll cross that out from the PR.

paologallinaharbur (approved):

paologallinaharbur (approved):
LGTM

PR #307, opened 2021-12-29
all: replace integration-sdk logger with logrus

roobre:


paologallinaharbur (approved):
LGTM

src/sink/http_test.go
```diff
@@ -181,7 +181,7 @@ func defaultPesterClient(t *testing.T) *pester.Client {
 	c.MaxRetries = 5
 	c.Timeout = sink.DefaultRequestTimeout
 	c.LogHook = func(e pester.ErrEntry) {
-		log.NewStdErr(false)
+		log.Println(e)
```
paologallinaharbur, line 184: should this be an log.Error since e is ErrEntry??

roobre: Ane rror might be too scary I think, but it should at least be a warning. Let me change that :)

PR #309, opened 2021-12-31
main: extract logic to integration object

roobre:
The old `main` function of the integration was doing many things and handling too many errors. This PR moves that logic, mostly untouched, to an `Integration` object tasked with managing the lifecycle of the integration, including building configuration, cluster clients, and running the scrapers.


paologallinaharbur:
Closing for now since it become stale and I fear it is easier to do it from scratch rather then rebase

PR #312, opened 2022-01-03
ksm: wait for some time for endpoints to appear before erroring

roobre:
A new `EndpointsDiscoverer` implementation, `EndpointsDiscovererWithTimeout` has been added. EndpointsDiscovererWithTimeout wraps another EndpointsDiscoverer adding a simple retry logic on top.

Main motivation for this change is to have the KSM pod no crash if KSM has not been deployed, although it can also be useful to make it more tolerant for other situations like KSM being migrated.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -35,6 +35,10 @@ type KSM struct {
 	Namespace   string `mapstructure:"namespace"`
 	Distributed bool   `mapstructure:"distributed"`
 	Enabled     bool   `mapstructure:"enabled"`
+	Discovery   struct {
+		Wait    time.Duration `mapstructure:"retry"`
+		Retires int           `mapstructure:"retries"`
+	} `mapstructure:"discovery"`
 }
 
 type Kubelet struct {
@@ -92,6 +96,9 @@ func LoadConfig(filePath string, fileName string) (*Config, error) {
 	v.SetDefault("nodeName", "node")
 	v.SetDefault("nodeIP", "node")
 	v.SetDefault("httpServerPort", 0)
+
+	v.SetDefault("ksm.discovery.retry", 10*time.Second)
+	v.SetDefault("ksm.discovery.retries", 6)
 
 	v.SetEnvPrefix("NRI_KUBERNETES")
 	v.AutomaticEnv()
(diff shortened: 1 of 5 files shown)


internal/config/config.go
```diff
@@ -93,6 +97,9 @@ func LoadConfig(filePath string, fileName string) (*Config, error) {
 	v.SetDefault("nodeIP", "node")
 	v.SetDefault("httpServerPort", 0)
```
paologallinaharbur, line 99: why keeping this NL?

internal/config/config.go
```diff
@@ -93,6 +97,9 @@ func LoadConfig(filePath string, fileName string) (*Config, error) {
...
 
+	v.SetDefault("ksm.discovery.retry", 7*time.Second)
```
paologallinaharbur, line 100: I do not like that much namings, it is not clear that timeout it is the "totalTimeout" and that retry is the time slept between one try and the other

paologallinaharbur: whats about "backoffTime"?

paologallinaharbur (approved):
some nits, but looking good

internal/discovery/endpoints_discoverer.go
```diff
@@ -97,3 +98,34 @@ func (d *endpointsDiscoverer) Discover() ([]string, error) {
...
+type EndpointsDiscovererWithTimeout struct {
+	EndpointsDiscoverer
+	Retry   time.Duration
+	Timeout time.Duration
+}
+
+// Discover will call poll the inner EndpointsDiscoverer every Retry seconds up to a max of Retries times until it
+// returns an error, or a non-empty list of endpoints.
+// If the max number of Retries is exceeded, it will return ErrDiscoveryTimeout.
+func (edt *EndpointsDiscovererWithTimeout) Discover() ([]string, error) {
```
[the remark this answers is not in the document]
roobre: **E**ndpoints**Discoverer**With**Timeout**

internal/config/config.go
roobre: No reason actually, I can get rid of it :)

internal/config/config.go
roobre: LGTM, [wikipedia](https://en.wikipedia.org/wiki/Exponential_backoff) uses "delay" to refer to this so I've followed that and used `BackoffDelay`. How does it look to you?

roobre: Note to self, change the values in the `mapstructure` tag as they are still the old names.

paologallinaharbur (approved):

Issue #313, opened 2022-01-03
Configuration file is in an invalid path


kang-makes:
While I am trying to fix other issue, I am unable de run the integration because the configuration file is hardcoded.

It is not a perfect approach and I could bear with it until I realized that it is in a weird location:

```
% go run ./cmd/nri-kubernetes -c ./here.yaml
[ERR] Config File "nri-kubernetes-config" Not Found in "[/etc/newrelic-infra/integrations.d]"
exit status 3
```

That path is for the agent, not the integration itself.

I create this issue to follow all this effort as there is more than enabling or changing the folder where to read the file but also change the charts.

- [x] https://github.com/newrelic/helm-charts/blob/master/charts/newrelic-infrastructure-v3/templates/control-plane/daemonset.yaml#L85
- [x] https://github.com/newrelic/helm-charts/blob/master/charts/newrelic-infrastructure-v3/templates/ksm/deployment.yaml#L71
- [x] https://github.com/newrelic/helm-charts/blob/master/charts/newrelic-infrastructure-v3/templates/kubelet/daemonset.yaml#L83

PR #314, opened 2022-01-03
Fix config file location

kang-makes:
Solves #313

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -84,7 +84,7 @@ type MTLS struct {
 func LoadConfig(filePath string, fileName string) (*Config, error) {
 	v := viper.New()
 
-	// We need to assure that defaults have been set in order to bind env variables
+	// We need to assure that defaults have been set in order to bind env variables.
 	// https://github.com/spf13/viper/issues/584
 	v.SetDefault("clusterName", "cluster")
 	v.SetDefault("verbose", false)
@@ -102,7 +102,7 @@ func LoadConfig(filePath string, fileName string) (*Config, error) {
 	v.AddConfigPath(".")
 	v.SetConfigName(fileName)
 
-	// This could fail not only if file has not been found or has errors in the YAML/missing attributes but also with errors in environment variables
+	// This could fail not only if file has not been found or has errors in the YAML/missing attributes but also with errors in environment variables.
 	if err := v.ReadInConfig(); err != nil {
 		return nil, err
 	}
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -8,8 +8,8 @@ import (
 )
 
 const (
-	FileName = "nri-kubernetes-config"
-	FilePath = "/etc/newrelic-infra/integrations.d/"
+	DefaultFileName = "nri-kubernetes"
+	DefaultFilePath = "/etc/newrelic-infra"
 )
 
 type Config struct {
@@ -84,7 +84,7 @@ type MTLS struct {
 func LoadConfig(filePath string, fileName string) (*Config, error) {
 	v := viper.New()
 
-	// We need to assure that defaults have been set in order to bind env variables.
+	// We need to assure that defaults have been set in order to bind env variables
 	// https://github.com/spf13/viper/issues/584
 	v.SetDefault("clusterName", "cluster")
 	v.SetDefault("verbose", false)
@@ -99,9 +99,10 @@ func LoadConfig(filePath string, fileName string) (*Config, error) {
 
 	// Config File
 	v.AddConfigPath(filePath)
+	v.AddConfigPath(".")
 	v.SetConfigName(fileName)
 
-	// If error reading file or file not found, use flag/env variables
+	// This could fail not only if file has not been found or has errors in the YAML/missing attributes but also with errors in environment variables
 	if err := v.ReadInConfig(); err != nil {
 		return nil, err
 	}
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,9 @@
 .DS_Store
-deploy/local.*
-deploy/local-unprivileged.yaml
-deploy/local-openshift.yaml
 bin
 .idea
+
+nri-kubernetes.yaml
+nri-kubernetes.yml
+
 e2e/charts/e2e-resources/Chart.lock
 e2e/charts/e2e-resources/charts/
--- a/cmd/nri-kubernetes/main.go
+++ b/cmd/nri-kubernetes/main.go
@@ -55,7 +55,7 @@ type clusterClients struct {
 }
 
 func main() {
-	c, err := config.LoadConfig(config.FilePath, config.FileName)
+	c, err := config.LoadConfig(config.DefaultFilePath, config.DefaultFileName)
 	if err != nil {
 		log.Error(err.Error())
 		os.Exit(exitIntegration)


internal/config/config.go
```diff
@@ -4,12 +4,13 @@ import (
 	"strings"
 	"time"
 
+	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 )
 
 const (
-	FileName = "nri-kubernetes-config"
-	FilePath = "/etc/newrelic-infra/integrations.d/"
```
paologallinaharbur, line 12: I read the issue but I did not understand why the location is invalid and what you mean by "That path is for the agent, not the integration itself."
The charts looks like it is mounting the file exacting there and so far always worked.

internal/config/config.go
```diff
@@ -99,9 +125,10 @@ func LoadConfig(filePath string, fileName string) (*Config, error) {
 
 	// Config File
 	v.AddConfigPath(filePath)
+	v.AddConfigPath(".")
```
paologallinaharbur, line 128: This is useful when running it locally, we were missing it 👍

paologallinaharbur: Now that I think about it.. should we create a const for it?

internal/config/config_test.go
```diff
@@ -18,15 +18,15 @@ func TestLoadConfig(t *testing.T) {
...
+		c, err := config.LoadConfigWithFileLocation(fakeDataDir, workingData)
 		require.NoError(t, err)
 		require.Equal(t, "dummy_cluster", c.ClusterName)
 
 		t.Run("with_env_precedence", func(t *testing.T) {
 			_ = os.Setenv("NRI_KUBERNETES_CLUSTERNAME", "different_value")
 			_ = os.Setenv("NRI_KUBERNETES_NODENAME", "fake-node")
 
-			c, err := config.LoadConfig(fakeDataDir, workingData)
+			c, err := config.LoadConfigWithFileLocation(fakeDataDir, workingData)
```
paologallinaharbur, line 29: I would add a test for loadConfig as well

internal/config/config.go
```diff
@@ -81,7 +82,32 @@ type MTLS struct {
...
-func LoadConfig(filePath string, fileName string) (*Config, error) {
+func LoadConfig() (*Config, error) {
+	v := viper.New()
+
+	pflag.StringP("config-name", "c" , FileName, "filename of the configuration to look for")
+	pflag.StringP("config-path", "p" , FilePath, "path where the configuration file is")
+
+	// Environment variables instead of v consistent with the config file
+	v.SetEnvPrefix("NRI_KUBERNETES")
+	v.AutomaticEnv()
```
paologallinaharbur, line 93: Do we want to support env for these flags?

internal/config/config.go
```diff
@@ -81,7 +82,32 @@ type MTLS struct {
 	TLSSecretNamespace string `mapstructure:"secretNamespace"`
 }
 
-func LoadConfig(filePath string, fileName string) (*Config, error) {
+func LoadConfig() (*Config, error) {
+	v := viper.New()
```
paologallinaharbur, line 86: Can't we leverage the same viper "obj" or is it cleaner in this way?

internal/config/config.go
```diff
@@ -99,9 +125,10 @@ func LoadConfig(filePath string, fileName string) (*Config, error) {
 
 	// Config File
 	v.AddConfigPath(filePath)
+	v.AddConfigPath(".")
 	v.SetConfigName(fileName)
 
-	// If error reading file or file not found, use flag/env variables
```
paologallinaharbur, line 104: I do not understand neither the old or the new comment 😅

internal/config/config.go
kang-makes: I made a PR for the helm chart as well: https://github.com/newrelic/helm-charts/pull/638

`/etc/newrelic-infra/integrations.d/` is the path were the agent read the configurations for integrations, not the config for the integration itself.

As the agent is if forward only mode there is so much issue with that but in the case of kubelet daemonset, the agent compalins of that  file not being a valid config v3/v4 config file.

TL;DR: Agent expect [this](https://docs.newrelic.com/docs/infrastructure/host-integrations/infrastructure-integrations-sdk/specifications/host-integrations-standard-configuration-format/#config) instead of the configuration for the integration itself.

internal/config/config.go
kang-makes: I thought about it but then it would be an array/slice and slices are not constants in Go.

I accept suggestions.

internal/config/config.go
kang-makes: If could make the development easier or could change the location using envs instead of flags.

I would support everything just in case. it could be a NTH in the future to use remote configuration, for example.

internal/config/config.go
paologallinaharbur: The chart should not mount that as well in the agent sidecar, therefore  the agent should not aware of this at all.

However, it is true what you say and `/etc/newrelic-infra` looks better

internal/config/config.go
kang-makes: If I use the same Viper object (and thus I can reuse the same function), Viper will try to look for `configPath`(or was it `config-path`?) and `configName` in the YAML and fails because they are not optional yada yada...

paologallinaharbur:
Related to https://github.com/newrelic/helm-charts/pull/638/files

internal/config/config.go
kang-makes: I am rookie so I could be wrong. Correct me if I am wrong, please.

The agent does not need of this file because is a configuration for the integration itself, the agent should not have any configuration file for this integration as it should not run/execute any binary for this integration.

What the agent should have is an open port were this integration will POST the metrics that the agent should send to the backend.

AFAIK the agent do not need anything for that except forwardMode and forwardOnlyMode, right?

internal/config/config.go
kang-makes: Fixed in a nit picking commit. Would you resolve this conversation?

internal/config/config.go
paologallinaharbur: everything correct 😄 
The agent container should not be even aware of this file, therefore you were right that `integrations.d/` looks weird

internal/config/config.go
```diff
@@ -81,10 +82,35 @@ type MTLS struct {
 	TLSSecretNamespace string `mapstructure:"secretNamespace"`
 }
 
-func LoadConfig(filePath string, fileName string) (*Config, error) {
+func LoadConfig() (*Config, error) {
 	v := viper.New()
 
-	// We need to assure that defaults have been set in order to bind env variables.
+	pflag.StringP("config-name", "c", FileName, "filename of the configuration to look for")
+	pflag.StringP("config-path", "p", FilePath, "path where the configuration file is")
```
roobre, line 89: These seem kind of redundant to me. I'd expect to either look for a hardcoded name (`nri-kubernetes.yaml`) in a harcoded list of paths (`.`, `/etc/newrelic-infra`), or to just point to the full path directly.

internal/config/config.go
```diff
@@ -81,10 +82,35 @@ type MTLS struct {
...
+
+	// If error parsing flag/env variables
+	if err := v.BindPFlags(pflag.CommandLine); err != nil {
+		return nil, err
+	}
+
+	return LoadConfigWithFileLocation(
+		v.GetString("config-path"),
+		v.GetString("config-name"),
+	)
```
roobre, line 107: Could we get rid of this if we use fullpath-or-default as suggested above?

internal/config/config.go
kang-makes: Done.

internal/config/config_test.go
kang-makes: Done.

internal/config/config.go
paologallinaharbur: I mean simply to add a separate constant like ` const localPath = "."` instead of an inline variable.

The other approach would be to use a func just for that:
```golang
func getSlice() []string {
    return []string{"hello", "world"}
}
```
but maybe it is too much

internal/config/config.go
paologallinaharbur: It makes sense to have them divided since Viper is asking for those separately and therefore we would need to split it, no?

paologallinaharbur (approved):
Overall it looks ok!

Pre-approving, but there is still one nit open and a conversation from Roberto

kang-makes:
In the end, the path with `integrations.d` was not making any issues because the configuration was mounted in the integration container but not in the agent sidecar but this can lead to misunderstandings.

I have ditched all the work from the last 2 hours because tests were passing but when I tried to run the integration itself failed because the Integration SDK also manages flags its own way.

```
flag provided but not defined: -c
Usage of /private/var/folders/n0/x5b9pcyx5f3528l599g3mymh0000gn/T/GoLand/___2go_build_github_com_newrelic_nri_kubernetes:
Exiting.
```

So I simply added the local folder as a source were to load the configuration file and changed path where to load the integration configuration.

paologallinaharbur (approved):
LGTM (😢 SDK)

internal/config/config.go
```diff
@@ -84,7 +84,7 @@ type MTLS struct {
 func LoadConfig(filePath string, fileName string) (*Config, error) {
 	v := viper.New()
 
-	// We need to assure that defaults have been set in order to bind env variables.
+	// We need to assure that defaults have been set in order to bind env variables
```
roobre, line 87: Bring the trailing dot back 😠

roobre (approved):

internal/config/config.go
kang-makes: https://github.com/newrelic/newrelic-infra-operator/blob/ce13dfff70008c54ed0598bb6904b0cf8fed9808/.golangci.yml#L66

https://github.com/tetafro/godot

Ok. Omw.

paologallinaharbur: I though the same, but it is a multiline comment

PR #315, opened 2022-01-04
ci: add unsupported build to CI

roobre:


roobre:
Pinging @cristianciutea just in case he can make a very quick review and ensure this way of building the agent is not complete nonsense :)

Dockerfile.windows
```diff
@@ -2,12 +2,14 @@ ARG IMAGE_NAME=newrelic/infrastructure
 ARG IMAGE_TAG=1.10.36-windows-1809-alpha
 FROM $IMAGE_NAME:$IMAGE_TAG
 
+LABEL unsupported=true
```
paologallinaharbur, line 5: 🤔

.github/workflows/windows.yaml
```diff
@@ -0,0 +1,98 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.16'
+      - name: Get windows build
+        id: osinfo
+        shell: powershell
+        run: |
+          Get-ComputerInfo
```
paologallinaharbur, line 40: is this needed?

paologallinaharbur (approved):
everything looks good

.github/workflows/windows.yaml
roobre: Previous iterations of this script used the output of this step to determine the windows version programatically. Now it is defined by hand on the matrix, but I thought the command was still useful as it might help in debugging. It tells you the details of the windows version, including build number etc.

Dockerfile.windows
roobre: Just to set expectations on people browsing this repo :)

PR #316, opened 2022-01-04
Merge Develop into Main

paologallinaharbur:
Pending tasks:
 - [x] update canaries (different repo) -> different PR
    - nri-bundle should be updated
    - nightly should be checked
    - canary code should 
  - [x] decide what to do with commit history


roobre (approved):
Approving without having looked at the diff 🙈 
Regarding the commit history, I'd vote to do a fully-fledged merge for this, we don't want to lose all commit history. You can enable it in the settings (it will be reset at 3AM by the setting sync bot).

roobre:
RE canaries, it should be easy to change the chart to point to the v3 image, even if it is not in the bundle. In my personal cluster I have this as the values for the wrapper chart:

```yaml
global:
  cluster: Terabox # Cluster name
  licenseKey: redacted

newrelic-infrastructure-v3:
  privileged: false

nri-bundle:
  kube-state-metrics:
    prometheusScrape: false

  infrastructure:
    enabled: false

  prometheus:
    enabled: false
  nri-prometheus:
  # ...
```

gsanchezgavier:
would make sense to update the root readme.md too here ? not sure for instance if skaffold is working

paologallinaharbur:
True, the readme is broken, before merging I would fix that adding some new info.

But I would open a new (😭 ) Pr against develop

paologallinaharbur:
Currently failing due to the NRI_KUBERNETES_SINK_HTTP_PORT var that changed name in the chart

Issue #311, opened 2022-01-03
Fix local computation of utilization metrics


roobre:
As discovered during #303, memory and cpu utlization metrics are not being computed correctly.

For now, impact for users is nil as these metrics are calculated in the backend as a fallback, but we should fix this behavior locally.

This happens because we are feeding to the `toUtilization` function a metric name that is not present in the `RawGroups`, but rather in the processed metrics:
https://github.com/newrelic/nri-kubernetes/blob/99eaab95b4171277e9cd2ea4774d05665047ee14/src/metric/definition.go#L941-L941

@paologallinaharbur has proposed a fix:

> 
> The fix could be something like
> ```
> func toUtilizationFetchFunc(dividendMetric, divisorMetric definition.FetchFunc) definition.FetchFunc {
> 
> 	return func(groupLabel, entityID string, groups definition.RawGroups) (definition.FetchedValue, error) {
> 		dividend, err := dividendMetric(groupLabel, entityID, groups)
> 		if err != nil {
> 			return nil, fmt.Errorf("getting dividend metric %q: %w", divisorMetric, err)
> 		}
> 		divisor, err := divisorMetric(groupLabel, entityID, groups)
> 		if err != nil {
> 			return nil, fmt.Errorf("getting divisor metric %q: %w", divisorMetric, err)
> 		}
> 		return computePercentage(dividend.(uint64), divisor.(uint64))
> 	}
> }
> ```
> being able to pass fetch function instead of using `fromRawData`, but we need to take care of casting the data correctly since they returns interface{}

PR #319, opened 2022-01-05
Computed metrics

paologallinaharbur:
Computed metrics were affected by a bug preventing them to work at all 

Fix newrelic/nri-kubernetes#311


src/kubelet/kubelet_test.go
```diff
@@ -118,8 +110,6 @@ func kubeletExclusions() []exclude.Func {
 				return limitsRequestsRegex.MatchString(spec.Name)
 			},
 		),
-		// Exclude metrics that depend on limits when those limits are not set.
-		exclude.Exclude(exclude.Groups("pod", "container"), exclude.Dependent(utilizationDependencies)),
```
roobre, line 122: This line was intended to exclude utilization metrics only when pod does not have request/limits, should we really remove it?

src/metric/definition.go
```diff
@@ -1017,11 +1027,46 @@ func isPersistentVolume() definition.FetchFunc {
...
+func convertValue(v interface{}) (float64, error) {
+	switch v := v.(type) {
+	case uint64:
+		return float64(v), nil
+	case int:
+		return float64(v), nil
+	case prometheus.GaugeValue:
+		return float64(v), nil
+	case float64:
+		return v, nil
```
roobre, line 1059: I'd say we can shorten the switch statement like this:
```suggestion
	case uint64, int, float64, prometheus.GaugeValue:
		return float64(v), nil
```

src/metric/definition.go
```diff
@@ -1017,11 +1027,46 @@ func isPersistentVolume() definition.FetchFunc {
...
+	case definition.FetchedValues:
+		if len(v) == 1 {
+			for _, k := range v {
+				return convertValue(k)
+			}
+		}
+		return 0, fmt.Errorf("unable to convert FetchedValues")
```
roobre, line 1066: And get rid of one indentation level by erroring out earlier
```suggestion
		// If retrieved metric is a list of values, we can only handle the case where
		// the list has only one.
		if len(v) != 1 {
			return 0, fmt.Errorf("unable to convert FetchedValues")
		}
		for _, m := range v {
			return convertValue(m)
		}
```

roobre (commented):
Should we add a test case covering the multi-values thing?

src/metric/definition.go
paologallinaharbur: Screen Shot 2022-01-07 at 12 42 29 PM
That was my first implementation, but sadly go compiler is not smart enough. I found it really weird as well that was not possible

src/metric/definition.go
paologallinaharbur: makes sense!

src/kubelet/kubelet_test.go
paologallinaharbur: That is not needed anymore, I can bring it back in case it is needed in the future :)

paologallinaharbur: @roobre let me know what you think

src/metric/definition.go
roobre: fuck, you're totally right.

src/kubelet/kubelet_test.go
roobre: I'm not sure I understand, why is it not needed anymore 🤔 

If I have a pod without `requests.limits.memory` (in k8s), what would the value of `memoryUtilization` be?

paologallinaharbur: that in the current scenario it is not needed. We do not need that exclude rule to pass tests with the current dataset

roobre: That should not be the case, we should miss this metrics as many (most) of the containers have no `resources` specified.

I think there's a bug on the not running exclusion, I'm working on it now.

roobre: Done, just pushed a commit to make the test fail 😈 @paologallinaharbur

roobre:
Btw, given the excessive amount of metrics we need to exclude, I'll be happy to use `Silently()` on this one.

roobre (approved):

Issue #308, opened 2021-12-29
Increase pester retry time


roobre:
## Description

In some of the pods, remarkably Kubelet, the integration will burn its retries too fast, even before the agent starts, causing the pod to crash.

## Expected Behavior
The integration should wait at least 20s total before giving up on POSTing to the agent service.

PR #320, opened 2022-01-05
Add backoff delay from pester to the agent forwarder

kang-makes:
Solves #308

### DIFF
--- a/cmd/nri-kubernetes/main.go
+++ b/cmd/nri-kubernetes/main.go
@@ -67,7 +67,7 @@ func main() {
 		logger.SetLevel(log.DebugLevel)
 	}
 
-	i, err := createIntegrationWithHTTPSink(c.HTTPServerPort, c.MaxRetries)
+	i, err := createIntegrationWithHTTPSink(c)
 	if err != nil {
 		logger.Errorf("creating integration with http sink: %v", err)
 		os.Exit(exitIntegration)
@@ -255,16 +255,16 @@ func buildClients(c *config.Config) (*clusterClients, error) {
 	}, nil
 }
 
-func createIntegrationWithHTTPSink(httpServerPort string, maxRetries int) (*integration.Integration, error) {
+func createIntegrationWithHTTPSink(config *config.Config) (*integration.Integration, error) {
 	c := pester.New()
 	c.Backoff = pester.ExponentialBackoff
-	c.MaxRetries = maxRetries
+	c.MaxRetries = config.MaxRetries
 	c.Timeout = sink.DefaultRequestTimeout
 	c.LogHook = func(e pester.ErrEntry) {
 		logger.Debugf("sending data to httpSink: %q", e)
 	}
 
-	endpoint := net.JoinHostPort(sink.DefaultAgentForwarderhost, httpServerPort)
+	endpoint := net.JoinHostPort(sink.DefaultAgentForwarderhost, config.HTTPServerPort)
 
 	sinkOptions := sink.HTTPSinkOptions{
 		URL:        fmt.Sprintf("http://%s%s", endpoint, sink.DefaultAgentForwarderPath),
--- a/cmd/nri-kubernetes/main.go
+++ b/cmd/nri-kubernetes/main.go
@@ -258,9 +258,11 @@ func buildClients(c *config.Config) (*clusterClients, error) {
 
 func createIntegrationWithHTTPSink(config *config.Config) (*integration.Integration, error) {
 	c := pester.New()
-	c.Backoff = pester.ExponentialBackoff
-	c.MaxRetries = config.Sink.HTTP.MaxRetries
-	c.Timeout = sink.DefaultRequestTimeout
+	c.Backoff = func(retry int) time.Duration {
+		return config.Sink.HTTP.BackoffDelay
+	}
+	c.MaxRetries = int(config.Sink.HTTP.Timeout/time.Second) // As many attempts as the Context timeout allows
+	c.Timeout = config.Sink.HTTP.ConnectionTimeout
 	c.LogHook = func(e pester.ErrEntry) {
 		logger.Debugf("sending data to httpSink: %q", e)
 	}
@@ -270,7 +272,7 @@ func createIntegrationWithHTTPSink(config *config.Config) (*integration.Integrat
 	sinkOptions := sink.HTTPSinkOptions{
 		URL:        fmt.Sprintf("http://%s%s", endpoint, sink.DefaultAgentForwarderPath),
 		Client:     c,
-		CtxTimeout: sink.DefaultCtxTimeout,
+		CtxTimeout: config.Sink.HTTP.Timeout,
 		Ctx:        context.Background(),
 	}
 
--- a/cmd/nri-kubernetes/main.go
+++ b/cmd/nri-kubernetes/main.go
@@ -67,7 +67,7 @@ func main() {
 		logger.SetLevel(log.DebugLevel)
 	}
 
-	i, err := createIntegrationWithHTTPSink(c.HTTPServerPort)
+	i, err := createIntegrationWithHTTPSink(c.HTTPServerPort, c.MaxRetries)
 	if err != nil {
 		logger.Errorf("creating integration with http sink: %v", err)
 		os.Exit(exitIntegration)
@@ -255,10 +255,10 @@ func buildClients(c *config.Config) (*clusterClients, error) {
 	}, nil
 }
 
-func createIntegrationWithHTTPSink(httpServerPort string) (*integration.Integration, error) {
+func createIntegrationWithHTTPSink(httpServerPort string, maxRetries int) (*integration.Integration, error) {
 	c := pester.New()
-	c.Backoff = pester.LinearBackoff
-	c.MaxRetries = 5
+	c.Backoff = pester.ExponentialBackoff
+	c.MaxRetries = maxRetries
 	c.Timeout = sink.DefaultRequestTimeout
 	c.LogHook = func(e pester.ErrEntry) {
 		logger.Debugf("sending data to httpSink: %q", e)
--- a/cmd/nri-kubernetes/main.go
+++ b/cmd/nri-kubernetes/main.go
@@ -261,7 +261,7 @@ func createIntegrationWithHTTPSink(config *config.Config) (*integration.Integrat
 	c.Backoff = func(retry int) time.Duration {
 		return config.Sink.HTTP.BackoffDelay
 	}
-	c.MaxRetries = int(config.Sink.HTTP.Timeout/time.Second) // As many attempts as the Context timeout allows
+	c.MaxRetries = int(config.Sink.HTTP.Timeout / time.Second) // As many attempts as the Context timeout allows
 	c.Timeout = config.Sink.HTTP.ConnectionTimeout
 	c.LogHook = func(e pester.ErrEntry) {
 		logger.Debugf("sending data to httpSink: %q", e)
--- a/cmd/nri-kubernetes/main.go
+++ b/cmd/nri-kubernetes/main.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path"
 	"runtime"
+	"strconv"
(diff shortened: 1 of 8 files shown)


cmd/nri-kubernetes/main.go
```diff
@@ -67,7 +67,7 @@ func main() {
 		logger.SetLevel(log.DebugLevel)
 	}
 
-	i, err := createIntegrationWithHTTPSink(c.HTTPServerPort)
+	i, err := createIntegrationWithHTTPSink(c.HTTPServerPort, c.MaxRetries)
```
roobre, line 70: Could we pass the whole config here? It will make easier to rebase #309

src/sink/http_test.go
```diff
@@ -177,8 +177,8 @@ func defaultPesterClient(t *testing.T) *pester.Client {
 	t.Helper()
 
 	c := pester.New()
-	c.Backoff = pester.LinearBackoff
-	c.MaxRetries = 5
+	c.Backoff = pester.ExponentialBackoff
+	c.MaxRetries = 6
 	c.Timeout = sink.DefaultRequestTimeout
```
roobre, line 182: Is `c.Timeout` enforced as the "global timeout" for all the retries? If so, does it make sense to have both a max amount of retries and a global timeout?

cmd/nri-kubernetes/main.go
kang-makes: Sure.

src/sink/http_test.go
kang-makes: `c.Timeout` is an enforced timeout for the connection itself. Meaning the amount of time it is opened a connection.

The global timeout is [here](https://github.com/newrelic/nri-kubernetes/blob/feature/exponential-backoff-for-pester/src/sink/http.go#L16) and is [context who closes the connection](https://github.com/newrelic/nri-kubernetes/blob/feature/exponential-backoff-for-pester/src/sink/http.go#L73).

We could retry until the context is closed instead of having the max retry counted. So it will close when the amount of time is be equal to the exponential backoff plus the theoretically maximum of 15 seconds is more than the context timeout.

The errors are also different:
On context timeout is:
```
DEBU[0068] sending data to httpSink: {"2022-01-05 15:17:22.534716 +0100 CET m=+68.025432250" "Do" "http://localhost:8084/v1/data" "POST" '\x00' '\a' '\x06' "Post \"http://localhost:8084/v1/data\": context deadline exceeded"}
ERRO[0068] publishing integration: performing HTTP request: Post "http://localhost:8084/v1/data": context deadline exceeded
exit status 4
```
While on running without agent is
```
DEBU[0007] sending data to httpSink: {"2022-01-05 15:18:14.276618 +0100 CET m=+7.569047590" "Do" "http://localhost:8084/v1/data" "POST" '\x00' '\x03' '\x02' "Post \"http://localhost:8084/v1/data\": dial tcp 127.0.0.1:8084: connect: connection refused"}
ERRO[0007] publishing integration: performing HTTP request: Post "http://localhost:8084/v1/data": dial tcp 127.0.0.1:8084: connect: connection refused
exit status 4
```

I would say that we almost always will have the `context deadline exceeded` error so we can go directly to retry until we hit the Default Context Timeout. Would you like to add a variable `DefaultCtxTimeout` instead of the `MaxRetries`?

internal/config/config.go
```diff
@@ -20,7 +20,8 @@ type Config struct {
 	NodeName       string        `mapstructure:"nodeName"`
 	HTTPServerPort string        `mapstructure:"httpServerPort"`
 	Interval       time.Duration `mapstructure:"interval"`
-	Timeout        time.Duration `mapstructure:"timeout"`
+	Timeout        time.Duration `mapstructure:"timeout"` // TODO: Unimplemented/unused
```
paologallinaharbur, line 23: would make sense to link it to the context of pester? 

If so in the config I would add a section like:
```
sink:
  HTTPServerPort:
  MaxRetries:
  Timeout:
```

kang-makes: This is coupled to the issue Roberto pointed out. Maybe get rid of `maxRetries` and simply put a variable for the Context timeout directly and retry until we hit the context timeout (backoff time + connection time)

Do I reuse `timeout` for this? [In the comments in the chart](https://github.com/newrelic/helm-charts/blob/master/charts/newrelic-infrastructure-v3/values.yaml#L41-L42) this was meant to be the timeout to connect to the metrics we are scrapping.

I have no problem reusing `timeout` it but I have not enough context of this integration to see if there are any case where we would need two different timeouts.

paologallinaharbur: we should check in the old code what timeout was doing 😕

paologallinaharbur: Before it was honored. I'll create an issue to track that we lost that functionality

I do not like that much Timeout there, I would prefer a Timeout per component (ETCD, Kublect, KSM) and one for the Sink. 

Just one Timeout in the Top level is not that meaningful

kang-makes: Good point.

What about this?
```yaml
sink:
  http:
    port:
    maxRetries:
    timeout:
```

There is no sight of using other sink than HTTP but this allows us to develop other sinks in the future (at config level).

paologallinaharbur: I like it 🆙

src/sink/http_test.go
roobre: I'm still not understanding the difference between `DefaultCtxTimeout` and `c.Timeout`. Which one applies to each connection individually, and which to all attempts globally?

kang-makes: `c.Timeout` is the timeout for a single request to happen from the integration to the agent.
`DefaultCtxTimeout` is the timeout for the whole process of connecting once: `DefaultCtxTimeout = backoff + c.Timeout`

kang-makes: Neither both of them are used globally for all the attempts because (the context is being close each connection attempt](https://github.com/newrelic/nri-kubernetes/blob/feature/exponential-backoff-for-pester/src/sink/http.go#L74)

roobre: Ok thanks, it's crystal clear now! It also makes sense why retries is used. Could we add small comments on top of these values explaining this? So we don't get confused when we look at this again :)

paologallinaharbur: As far as I remembered was the overall timeout for all requests. 

Notice that the context is closed at the end of each `.Write`. That function use the pester `.Do` that manages all the retries using the same context. This is why there are two:
```
// DefaultCtxTimeout is the default IO timeout for the context of the client.
DefaultCtxTimeout = 15 * time.Second
// DefaultRequestTimeout is the default IO timeout for each request.
DefaultRequestTimeout = 15 * time.Second
```

Moreover, [pester itself](//If the request has been cancelled, skip retries) is request.contex aware stating "//If the request has been cancelled, skip retries" 
If that is not the behaviour than there is a bug since having the two was intended :)

internal/config/config.go
kang-makes: Last commit added `sink` option tree.

internal/config/config.go
```diff
@@ -18,15 +18,24 @@ type Config struct {
 	KubeconfigPath string        `mapstructure:"kubeconfigPath"`
 	NodeIP         string        `mapstructure:"nodeIP"`
 	NodeName       string        `mapstructure:"nodeName"`
-	HTTPServerPort string        `mapstructure:"httpServerPort"`
 	Interval       time.Duration `mapstructure:"interval"`
-	Timeout        time.Duration `mapstructure:"timeout"`
+	Timeout        time.Duration `mapstructure:"timeout"` // TODO: Unimplemented/unused (issue #322)
+
+	Sink struct {
```
paologallinaharbur, line 24: I would avoid inline struct (merely because the file is not using them)

paologallinaharbur: mmm but it is true that it has only an attribute... I am not sure, I still prefer them not inlined, but I understand why you place it there, up to you if you want to change it :)

internal/config/config.go
```diff
@@ -97,7 +106,9 @@ func LoadConfig(filePath string, fileName string) (*Config, error) {
 	v.SetDefault("kubelet.networkRouteFile", "/proc/net/route")
 	v.SetDefault("nodeName", "node")
 	v.SetDefault("nodeIP", "node")
-	v.SetDefault("httpServerPort", 0)
```
paologallinaharbur, line 100: This is set from an env var in the chart and due to 
```
	// We need to assure that defaults have been set in order to bind env variables.
	// https://github.com/spf13/viper/issues/584
```
we still need to have a default for that and change their  name if we want to be able to customize it from envs.

(however it is true that currently we do not strictly need different ports since only control plane has hostNetwork=true)

paologallinaharbur: we can sync offline about this

paologallinaharbur (commented):
Only a single doubt remaining, everything else looking good!

internal/config/config.go
kang-makes: I completely forgot. I am bringing this line back.

internal/config/config.go
kang-makes: I just copy and pasted what @roobre had before me for the KSM as it is a single attribute. KSM has 2:
```go
	Discovery   struct {
		// Wait BackoffDelay between discovery attempts.
		BackoffDelay time.Duration `mapstructure:"backoffDelay"`
		// Give up discovery and fail if Timeout has passed since first attempt.
		Timeout time.Duration `mapstructure:"timeout"`
	} `mapstructure:"discovery"`
```

I prefer to not complicate this more as there is only one sink (yet and hoping to not have more than one)

cmd/nri-kubernetes/main.go
```diff
@@ -255,21 +256,23 @@ func buildClients(c *config.Config) (*clusterClients, error) {
...
 
-func createIntegrationWithHTTPSink(httpServerPort string) (*integration.Integration, error) {
+func createIntegrationWithHTTPSink(config *config.Config) (*integration.Integration, error) {
 	c := pester.New()
-	c.Backoff = pester.LinearBackoff
-	c.MaxRetries = 5
-	c.Timeout = sink.DefaultRequestTimeout
+	c.Backoff = func(retry int) time.Duration {
+		return config.Sink.HTTP.BackoffDelay
+	}
```
roobre, line 263: Smart, I like it.

internal/config/config.go
```diff
@@ -18,15 +18,25 @@ type Config struct {
...
+	} `mapstructure:"sink"`
 
 	ControlPlane `mapstructure:"controlPlane"`
 	Kubelet      `mapstructure:"kubelet"`
 	KSM          `mapstructure:"ksm"`
 }
 
+type HTTPSink struct {
+	Port              int           `mapstructure:"port"`
+	ConnectionTimeout time.Duration `mapstructure:"connectionTimeout"` // Give up ConnectionTimeout each connection attempt to the agent.
```
roobre, line 35:
```suggestion
	ConnectionTimeout time.Duration `mapstructure:"connectionTimeout"` // Give up on a connection if it takes more than ConnectionTimeout to complete.
```

cmd/nri-kubernetes/main.go
```diff
@@ -255,21 +256,23 @@ func buildClients(c *config.Config) (*clusterClients, error) {
...
+	c.MaxRetries = int(config.Sink.HTTP.Timeout / time.Second) // As many attempts as the Context timeout allows
```
roobre, line 264:
```suggestion
	// Allow as many attempts as seconds are in the global timeout.
	// This is a rough and conservative approximation as we don't actually need to enforce a number of retries: 
	// We simply rely on the global timeout instead. However, pester requires this value to be set regardless.
	c.MaxRetries = int(config.Sink.HTTP.Timeout / time.Second)
```

roobre (approved):
Left some nits, otherwise looking good!

cmd/nri-kubernetes/main.go
kang-makes: I hesitated to write such a big comment. Good to know that I should have done it.

PR #324, opened 2022-01-13
Readme and structure update

paologallinaharbur:
-> No content has been added, I simply deleted broken or totally outdated stuff
 -> Skaffold was broken and still pointing to old manifests
 -> No reason to keep windows docs in this branch


roobre (approved):
LGTM!

My only nit would be that we're now mixing hard-wrapped and non-wrapped lines in the README.md, I'd suggest to unwrap them to make it more uniform.

paologallinaharbur:
I really hate super long lines, I would wrap them all, but the repolinter complains, this is why they are mixed 😠 
I would not block this in the meanwhile for a discussion since we are going to revisit it as part of the doc effort:)

Issue #325, opened 2022-01-13
Add retry mechanism in kubelet component


paologallinaharbur:
The kubelet component currently restart if kubelet is not able to answer.

Therefore pod restarts from time to time due to network issues (in canary ~4 times in 2 months)

It would be nice to add something like pester to deal with this sort of issues

Issue #322, opened 2022-01-05
Timeout config is not honored


paologallinaharbur:
Currently the Timeout option to scrape data is not honored, I would like to have one per component es:

 - `ControlPlane.Timeout` (or even one per component)
 - `Kubelet.Timeout`
 - `KSM.Timeout`

This should not be confused with pester timeout

Old behaviour/docs:
```
Timeout       int    `default:"5000" help:"timeout in milliseconds for calling metrics sources"`
```

That used to be passed to each http client

paologallinaharbur:
Keep in mind that if pester is used then there is the timeout per request and for all retries

PR #326, opened 2022-01-13
feat(kubelet): adding pester as client wrapper

paologallinaharbur:
Fix https://github.com/newrelic/nri-kubernetes/issues/325

After checking memory leaks I noticed that kubelet fails from time to time due to kubelet not answering

### DIFF
--- a/src/kubelet/client/client.go
+++ b/src/kubelet/client/client.go
@@ -7,6 +7,7 @@ import (
 	"path"
 	"time"
 
+	"github.com/sethgrid/pester"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/newrelic/nri-kubernetes/v2/internal/logutil"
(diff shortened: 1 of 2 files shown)


src/kubelet/client/connector.go
```diff
@@ -228,6 +230,18 @@ func checkConnection(conn connParams) error {
...
 }
 
+func (dp *defaultConnector) pesterClientWithTimeout(timeout time.Duration) *pester.Client {
+	httpPester := pester.New()
+	httpPester.Backoff = pester.LinearBackoff
+	httpPester.MaxRetries = 3
+	httpPester.Timeout = timeout
+	httpPester.LogHook = func(e pester.ErrEntry) {
+		dp.logger.Debugf("getting data from kubelet: %v", e)
+	}
```
kang-makes, line 240: Is this going to have issues because it is not able to connect to `kubelet` like with the agent?

Should we have this configurable in some way?

kang-makes: I mean, if Kubelet has an issue and start rejecting this will backoff almost instantly as it happened to the agent.

kang-makes (commented):
Left a comment.

src/kubelet/client/connector.go
paologallinaharbur: yes, but I would say that is linked to an other issue https://github.com/newrelic/nri-kubernetes/issues/322

I configured it in that way to mimic [KSM](https://github.com/newrelic/nri-kubernetes/blob/d67f7401273153b3dcc842f80323c7655d8a64f2/src/ksm/client/client.go#L39), but in the future I would make at least the timeout configurable per component

src/kubelet/client/connector.go
```diff
@@ -228,6 +230,18 @@ func checkConnection(conn connParams) error {
 	return nil
 }
 
+func (dp *defaultConnector) pesterClientWithTimeout(timeout time.Duration) *pester.Client {
```
roobre, line 233: I have some concerns on this method being part of the `defaultConnector`. I think the connector should do as few things as possible, namely returning `connParams` on `Connect()`.

Since pester supports wrapping an existing client (i.e. `pester.NewExtendedClient()`), do you think it might be better to use that?

Something like

```go
func (cp connParams) WithPester() connParams {
	cp.client = pester.NewExtendedClient(cp.client) // Ugly cast ellided for brevity
	return cp
}
```

We can also put that functionality outside of `connParams` if you think it's cleaner, I'm not fully in for baking it in.

paologallinaharbur: you are right, changing completely the approach

paologallinaharbur:
adding as KSM https://github.com/newrelic/nri-kubernetes/blob/d67f7401273153b3dcc842f80323c7655d8a64f2/src/ksm/client/client.go#L39

roobre (approved):
Looking superb, thanks a lot @paologallinaharbur! ✨

PR #331, opened 2022-01-17
feat(leak): custom cache cleaning itself periodically

paologallinaharbur:
The implementation is almost the same of the SDK one. The main difference is the goroutine cleaning each interval all entries too old

### DIFF
--- a/cmd/nri-kubernetes/main.go
+++ b/cmd/nri-kubernetes/main.go
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/newrelic/infra-integrations-sdk/integration"
+	"github.com/newrelic/infra-integrations-sdk/persist"
 	"github.com/sethgrid/pester"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/client-go/kubernetes"
@@ -20,6 +21,7 @@ import (
 	"k8s.io/client-go/util/homedir"
 
 	"github.com/newrelic/nri-kubernetes/v2/internal/config"
+	"github.com/newrelic/nri-kubernetes/v2/internal/storer"
 	"github.com/newrelic/nri-kubernetes/v2/src/client"
 	"github.com/newrelic/nri-kubernetes/v2/src/controlplane"
 	"github.com/newrelic/nri-kubernetes/v2/src/ksm"
@@ -68,7 +70,9 @@ func main() {
 		logger.SetLevel(log.DebugLevel)
 	}
 
-	i, err := createIntegrationWithHTTPSink(c)
+	cache := storer.NewInMemoryStore(storer.DefaultTTL, storer.DefaultInterval, logger)
+
+	i, err := createIntegrationWithHTTPSink(c, cache)
 	if err != nil {
 		logger.Errorf("creating integration with http sink: %v", err)
 		os.Exit(exitIntegration)
@@ -256,7 +260,7 @@ func buildClients(c *config.Config) (*clusterClients, error) {
 	}, nil
 }
 
-func createIntegrationWithHTTPSink(config *config.Config) (*integration.Integration, error) {
+func createIntegrationWithHTTPSink(config *config.Config, cache persist.Storer) (*integration.Integration, error) {
 	c := pester.New()
 	c.Backoff = func(retry int) time.Duration {
 		return config.Sink.HTTP.BackoffDelay
@@ -284,7 +288,7 @@ func createIntegrationWithHTTPSink(config *config.Config) (*integration.Integrat
 		return nil, fmt.Errorf("creating HTTPSink: %w", err)
 	}
 
-	return integration.New(integrationName, integrationVersion, integration.Writer(h), integration.InMemoryStore())
+	return integration.New(integrationName, integrationVersion, integration.Writer(h), integration.Storer(cache))
 }
 
 func getK8sConfig(c *config.Config) (*rest.Config, error) {
--- a/cmd/nri-kubernetes/main.go
+++ b/cmd/nri-kubernetes/main.go
@@ -12,7 +12,6 @@ import (
 	"time"
 
 	"github.com/newrelic/infra-integrations-sdk/integration"
-	"github.com/newrelic/infra-integrations-sdk/persist"
 	"github.com/sethgrid/pester"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/client-go/kubernetes"
@@ -70,9 +69,7 @@ func main() {
 		logger.SetLevel(log.DebugLevel)
 	}
 
-	cache := storer.NewInMemoryStore(storer.DefaultTTL, storer.DefaultInterval, logger)
-
-	i, err := createIntegrationWithHTTPSink(c, cache)
+	i, err := createIntegrationWithHTTPSink(c)
 	if err != nil {
 		logger.Errorf("creating integration with http sink: %v", err)
 		os.Exit(exitIntegration)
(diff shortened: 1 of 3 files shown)


cmd/nri-kubernetes/main.go
```diff
@@ -11,6 +11,10 @@ import (
 	"strings"
 	"time"
 
+	"github.com/newrelic/infra-integrations-sdk/persist"
+
```
paologallinaharbur, line 15: nit, cleaning before merging if approved

internal/storer/storer.go
```diff
@@ -0,0 +1,117 @@
...
+package storer
+
+import (
+	"errors"
+	"reflect"
+	"sync"
+	"time"
+
+	"github.com/newrelic/infra-integrations-sdk/persist"
+
```
paologallinaharbur, line 11: same nit to clean

internal/storer/storer.go
```diff
@@ -0,0 +1,116 @@
...
+	ims := &InMemoryStore{
+		cachedData: make(map[string]jsonEntry),
+		locker:     &sync.Mutex{},
+		ttl:        ttl,
+		logger:     logger,
+	}
+
+	go func() {
+		select {
+		case <-time.Tick(interval):
```
paologallinaharbur, line 47: missing loop

paologallinaharbur:
related to https://github.com/newrelic/nri-kubernetes/issues/332

internal/storer/storer.go
```diff
@@ -0,0 +1,118 @@
...
+		logger:     logger,
+	}
+
+	tk := time.NewTicker(interval)
+
+	go func() {
+		for {
+			<-tk.C
+			ims.cleanCache()
+		}
```
alvarocabanas, line 51: I miss Context propagation  in all the integration and  <-ctx.Done(): in this specific point to properly handle clean shutdown of the application. I just comment and you can resolve it because I think we are not using context at all in the integration and it's a more wider decission.

alvarocabanas (commented):
LGTM

internal/storer/storer.go
paologallinaharbur: Yes, it is a "known limitation", do you think that a //Todo comment will be helpful or since the whole integration is affected by that is not needed?

I the ctx.Done() we should also stop the ticker

cmd/nri-kubernetes/main.go
```diff
@@ -68,7 +70,9 @@ func main() {
 		logger.SetLevel(log.DebugLevel)
 	}
 
-	i, err := createIntegrationWithHTTPSink(c)
+	cache := storer.NewInMemoryStore(storer.DefaultTTL, storer.DefaultInterval, logger)
+
+	i, err := createIntegrationWithHTTPSink(c, cache)
```
roobre, line 75: Could we create the custom storer inside `createIntegrationWithHTTPSink`, rather than passing it? We could then save the extra parameter to `createIntegrationWithHTTPSink`.

internal/storer/storer.go
roobre: +1 to Álvaro's comment, I'd love if we could somehow give the caller a means to stop the background loop. Perhaps a simple `func (ims *InMemoryStore) Stop()` could do the trick.

internal/storer/storer.go
```diff
@@ -0,0 +1,118 @@
...
+	//notice this is the timestamp of the creation of the jsonEntry, we do not keep track of the last-access timestamp.
+	timestamp time.Time
+	value     interface{}
+}
+
+// NewInMemoryStore will create and initialize an InMemoryStore.
+func NewInMemoryStore(ttl time.Duration, interval time.Duration, logger *logrus.Logger) *InMemoryStore {
+	ims := &InMemoryStore{
+		cachedData: make(map[string]jsonEntry),
+		locker:     &sync.Mutex{},
```
roobre, line 40: Probably overkill, but changing this to an RWMutex it's pretty much a free improvment.

```suggestion
		locker:     &sync.RWMutex{},
```

internal/storer/storer.go
```diff
@@ -0,0 +1,118 @@
...
+		value:     value,
+	}
+	return ts.Unix()
+}
+
+// Get gets the value associated to a given key and stores it in the value referenced by the pointer passed as
+// second argument.
+func (j InMemoryStore) Get(key string, valuePtr interface{}) (int64, error) {
+	j.locker.Lock()
+	defer j.locker.Unlock()
```
roobre, line 74:
```suggestion
	j.locker.RLock()
	defer j.locker.RUnlock()
```

internal/storer/storer.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	// Using reflection to indirectly set the value passed as reference
+	reflect.Indirect(rv).Set(reflect.Indirect(reflect.ValueOf(entry.value)))
+
+	return entry.timestamp.Unix(), nil
+}
+
+// cleanCache removes the cached data entries if older than TTL.
+func (j InMemoryStore) cleanCache() {
```
roobre, line 95: I'm not a huge fan of the naming here, mainly because of the `cache` part as we usually refer to this object as "storer" everywhere else.

I'd suggest `Vacuum`, or `Cleanup`. The comment above should clarify to readers that we are removing expired entries.

```suggestion
func (j InMemoryStore) vacuum() {
```

internal/storer/storer.go
roobre: Another alternative would be to not use a background loop and force a cleanup on `Save()`. This is a bit obscure but the SDK would in the end call `Save()` once every run, so seems quite tempting to abuse it.

paologallinaharbur: yes, I had it like this in the POC I did not like it that much since it was being executed too often wasting cpu since the whole cache needed to be inspected.

roobre: You're totally right, I didn't think about how big the map gets for the kubelet 🤦🏻‍♀️

paologallinaharbur: I would use a goroutine for that, is there any drawback I miss? it is a bit more complex, but it is "standard".

We should also close it, but the code doe not allows contexts

paologallinaharbur: I added a way to stop it, but there is no one currently calling it

internal/storer/storer.go
```diff
@@ -0,0 +1,131 @@
...
+
+// Holder for any entry in the JSON storage.
+type jsonEntry struct {
+	//notice this is the timestamp of the creation of the jsonEntry, we do not keep track of the last-access timestamp.
+	timestamp time.Time
+	value     interface{}
+}
+
+// NewInMemoryStore will create and initialize an InMemoryStore.
+func NewInMemoryStore(ttl time.Duration, interval time.Duration, logger *logrus.Logger) *InMemoryStore {
```
gsanchezgavier, line 39: new packages are using the option or/and config pattern , would make sense to implement it here as well?

internal/storer/storer_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+		cache.Set(testKey, testValue)
+		time.Sleep(time.Second * 3)
+
+		_, err := cache.Get(testKey, &val)
+		assert.Equal(t, testValue, val)
+		assert.NoError(t, err)
+	})
+
+	t.Run("fails_if_value_is_nil_or_not_a_pointer", func(t *testing.T) {
```
gsanchezgavier, line 62: what would happen if the pointer type doesn't match with the type of the cache value ? could you add a test for that?

internal/storer/storer.go
paologallinaharbur: I wanted to keep the implementation as similar as possible to the sdk one, it is basically a copy paste with the cleaning.

internal/storer/storer.go
roobre: Looks alright to me!

internal/storer/storer.go
paologallinaharbur: but it could make sense to update that to be similar and have better structure

roobre (approved):

internal/storer/storer_test.go
paologallinaharbur: I updated adding a test and making the code less cryptic failing instead of panicking

gsanchezgavier (approved):
Looks good thanks!

PR #337, opened 2022-01-21
feat(tests): testing retries

paologallinaharbur:


alvarocabanas:
I added what you proposed

PR #338, opened 2022-01-21
sink: implement mTLS-enabled http client

roobre:
In sync with https://github.com/newrelic/infrastructure-agent/pull/913, this PR allows the HTTP sink to send metrics to the infrastructure-agent using HTTPS and mTLS.


src/integration/sink/tls.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/newrelic/nri-kubernetes/v3/internal/config"
+)
+
+var ErrCAAppend = errors.New("appending certs to pool")
+
+func NewTLSClient(conf config.TLSConfig) (*http.Client, error) {
```
alvarocabanas, line 16: I don't see as a bad approach to create a NewTLSClient_test that ensures that it can make calls to a mockedServer :
` srv := NewServer(httpsPort)
  go srv.ListenAndServeTLS("cert.pem", "key.pem")
  defer srv.Close()`

alvarocabanas: Because at the end, you are not testing http.Client but you are ensuring that your constructor creates a valid httpClient.

internal/config/config.go
```diff
@@ -64,6 +64,19 @@ type HTTPSink struct {
...
 	Retries int `mapstructure:"retries"`
+	// TLS allows to configure TLS encryption and authentication for the HTTP sink.
+	TLS TLSConfig `mapstructure:"tls"`
+}
+
+type TLSConfig struct {
+	// Enabled dictates whether TLS is used to connect to the HTTP sink.
+	Enabled bool `mapstructure:"enabled"`
+	// CertPath is the path to a PEM-encoded TLS certificate that will be presented to the server.
+	CertPath string `mapstructure:"certPath"`
```
paologallinaharbur, line 75: could we mount in a similar way of the ETCD mtls?

Not saying that I do not prefer this one, just that I would use the same approach

src/integration/prober/prober.go
```diff
@@ -15,19 +16,47 @@ import (
...
+	client  *http.Client
 }
 
 var ErrProbeTimeout = errors.New("probe timed out")
 var errProbeNotOk = errors.New("probe did not return 200 Ok")
 
+type OptionFunc func(p *Prober) error
+
+// WithLogger returns an OptionFunc which tells the Prober to use the specified logger.
+func WithLogger(logger *log.Logger) OptionFunc {
```
paologallinaharbur, line 29: This is not used

roobre: Should be, I forgot to make the change 😅 
Good catch!

src/integration/sink/tls.go
roobre: This does make a lot of sense, and it's perfectly testable. I'll add a test for it :)

roobre: The main problem with this is how incredibly painful it is to create an HTTPs server for testing. `httptest` uses hardcoded certificates, and to create an "standard" one you need to add a _lot_ of (potentially buggy) boilerplate to set it up.

I'll give it a try but I think it will be awfully complex.

internal/config/config.go
paologallinaharbur: mmmm true, if we can mount it directly it is better, the use case is different therefore we can use a different approach 👍

src/integration/sink/tls_test.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	server.TLS = &tlsConfig
+	server.StartTLS()
+
+	return server, nil
+}
+
+func TestTlsClient(t *testing.T) {
+	server, err := testServer()
```
sigilioso, line 47: For test helpers it maybe useful to include `*testing.T` as a parameter and drop the returning error, if it actually fails it a bit easier to find where it failed. Besides, in case the helper is widely used, you avoid checking the error each time it is used.

Something like this:
```go
func testServer(t *testing.T) *httptest.Server {
    // ...
    serverCert, err := tls.LoadX509KeyPair("testdata/server+3.pem", "testdata/server+3-key.pem")
    if err != nil {
         return t.Fatalf("error loading server certificate: %w", err)
    }
    // ...
}

// ...
server := testServer(t)
```

paologallinaharbur (approved):

src/integration/sink/tls_test.go
roobre: lgtm, I gave that a try on https://github.com/newrelic/nri-kubernetes/pull/338/commits/e9faf94dcb3bdad67e9cdd7e5aeadd6578f0a7d6

PR #341, opened 2022-01-24
Update dependabot.yml to skip e2e

paologallinaharbur:
Since external PRs do not have access to secrets they fail.

https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#labels

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,13 +7,22 @@ updates:
     open-pull-requests-limit: 3
     ignore:
       - dependency-name: "k8s.io/*"
+    labels:
+      - dependencies
+      - ci/skip-e2e 
   - package-ecosystem: docker
     directory: "/"
     schedule:
       interval: daily
     open-pull-requests-limit: 2
+    labels: 
+      - ci/skip-e2e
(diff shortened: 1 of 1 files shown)


roobre:
It would be sad to lose e2e tests on k8s dependency bumps :(

Is there any other way we can work around this?
E.g. perhaps we can give write access to the dependabot user, in a somewhat restricted way.

paologallinaharbur:
🤔  Actually dependabot already pushes stuff to the repo, therefore it has some access 

https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/managing-encrypted-secrets-for-dependabot

@roobre

paologallinaharbur:
Not possible https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/

roobre (approved):
Let's hope https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-940182533 is figured out soon :(

PR #344, opened 2022-01-25
controlplane/authenticator: allow to use `kubernetes.io/tls` secrets

### LINKED ISSUE #342: Control plane secret fetcher does not use `type: kubernetes.io/tls` secrets
While fetching secrets for mTLS authentication in the control plane authenticator, we retrieve the certificate data like this:

https://github.com/newrelic/nri-kubernetes/blob/f6d1980085e98d434d2ce48e16729179bcbce3db/src/controlplane/client/authenticator/authenticator.go#L142

However, the convention for TLS secrets in K8s says that the names for those fields should be `tls.crt`, `tls.key`, and `ca.crt`. Users will likely expect TLS secrets to work, so we should handle that scenario.

Proposal: We could check if the secret we have fetched as `type: kubernetes.io/tls`. If it does, we look for `tls.crt`, `tls.key`, and `ca.crt`. If it doesn't, we assume it's for the old convention and look for the "old" ones.

Moreover, we should doublecheck that we are base64-decoding the contents of the certificate, either manually or go-client doing it for us.

https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets

roobre:
Fixes #342


src/controlplane/client/authenticator/authenticator.go
```diff
@@ -115,6 +116,29 @@ type certificatesData struct {
...
+type certificateSecretKeys struct {
+	cert string
+	key  string
+	ca   string
+}
+
+// opaqueCertKey is the key for the secret data where the PEM-encoded certificate is located.
+const opaqueCertKey = "cert"
+
+// opaqueKeyKey is the key for the secret data where the PEM-encoded private key is located.
```
paologallinaharbur, line 130: do we want to say somehow that this is legacy? (if we want that of course)

src/controlplane/client/authenticator/authenticator.go
```diff
@@ -137,19 +161,35 @@ func (a K8sClientAuthenticator) getTLSCertificatesFromSecret(mTLSConfig *config.
...
 	}
 
-	var cert, key []byte
+	keynames := certificateSecretKeys{
+		cert: opaqueCertKey,
+		key:  opaqueKeyKey,
+		ca:   opaqueCacertKey,
+	}
+
+	if secret.Type == corev1.SecretTypeTLS {
```
paologallinaharbur, line 170: I would add a comment saying why we had different names and we now change them

src/controlplane/client/authenticator/authenticator_mtls_test.go
```diff
@@ -141,6 +142,68 @@ func Test_Authenticator_with_mTLS(t *testing.T) {
...
+			name: "from_tls_secret_with_k8s_keys",
+			listerer: secretListerer(t, secretName, secretNamespace, corev1.SecretTypeTLS, map[string][]byte{
+				corev1.TLSCertKey:       clientCert,
+				corev1.TLSPrivateKeyKey: clientKey,
+				"ca.crt":                clientCACert,
+			}),
+		},
+		{
+			// This test case is the same as above, but key names are hardcoded.
+			// This test case should fail if Kubernetes constants change unexpectedly.
```
paologallinaharbur, line 160: galaxy

src/controlplane/client/authenticator/authenticator.go
```diff
@@ -137,19 +161,35 @@ func (a K8sClientAuthenticator) getTLSCertificatesFromSecret(mTLSConfig *config.
...
+		a.logger.Debugf("Secret %q has type %q, using standard key names", secret.Name, secret.Type)
+
+		keynames = certificateSecretKeys{
+			cert: corev1.TLSCertKey,
```
paologallinaharbur, line 174: the API denies having a SecretTypeTLS with different keys right? even in past versions I imagine. 
To avoid someone with SecretTypeTLS and our keys (for any weird reason)

paologallinaharbur (approved):
Some nits, and corner case to doublecheck, everything else looks good!

src/controlplane/client/authenticator/authenticator.go
gsanchezgavier: agree, perhaps a general comment to the `certificateSecretKeys` struct mentioning the legacy NR naming for this secrets keys.

gsanchezgavier (approved):
Looks good to me ! 
I totally miss this when refactored the CP.
The reason we are keeping the old names is to avoid the friction on the migration and save the renaming for existing users that have this configured ?

roobre:
>  The reason we are keeping the old names is to avoid the friction on the migration and save the renaming for existing users that have this configured ?

Yes, the idea is to avoid making a big breaking change if there is a clean way of providing backwards compat.

Or, in the words of Helm maintainers:

src/controlplane/client/authenticator/authenticator.go
roobre: Not 100% sure if it you can submit a tls secret without the required keys, but even if you could, you shouldn't :P

In any case, the error (that should eventually get logged) explicitly says the key that we were not able to find:

https://github.com/newrelic/nri-kubernetes/blob/controlplane-tls-certs/src/controlplane/client/authenticator/authenticator.go#L187

paologallinaharbur (approved):

src/controlplane/client/authenticator/authenticator.go
```diff
@@ -116,6 +117,32 @@ type certificatesData struct {
 	ca   []byte
 }
 
+// certificateSecretKeys contains the name of the keys inside the secret where the certificate, private key, and CA
+// certificates are stored.
+// Earlier versions of the integration allowed to define a secret of type corev1.SecretTypeOpaque, with the required
+// certs and keys stored in keys named as the constants below.
+// New versions can also consume a secret of type corev1.SecretTypeTLS, using standard names for certificate and key.
```
gsanchezgavier, line 124: 👍

PR #345, opened 2022-01-25
ci/e2e: add 1.23.3 to test matrix, upgrade patches for other versions

roobre:
Pretty much as the title says.


paologallinaharbur:
The "continue-on-error for 1.23" did something weird, e2e tests are not marked as failed, but are not visible anymore 😭

roobre:
No idea why that happens 🤦🏻‍♀️ 
I'll check the workflow spec, maybe I need to put the flag on a `step` rather than the job.

roobre:
Closing as this is not working as it should, we should re-take this effort later.

PR #346, opened 2022-01-26
sink: wait for infra-agent to be ready before starting

roobre:
As the title says, after upgrading the `nri-bundle`  on my local environment, the `nrk8s-ksm` pod restarted 3 times while waiting for the agent to become ready. `nrk8s-kubelet` restarted twice. This simple trick should make a more graceful wait without alerting the user if the agent does not come up immediately.


paologallinaharbur:
yes, I noticed that as well, thanks for taking care of it!

src/sink/probe.go
```diff
@@ -0,0 +1,39 @@
...
+	for {
+		time.Sleep(backoff)
+
+		if time.Since(start) > timeout {
+			return fmt.Errorf("%w after %v", ErrProbeTimeout, timeout)
+		}
+
+		resp, err := client.Get(endpoint)
+		if err != nil {
+			continue
```
paologallinaharbur, line 28: I like the approach, but I would like to have some tests 😄

src/sink/probe.go
```diff
@@ -0,0 +1,39 @@
...
+	"fmt"
+	"net/http"
+	"time"
+)
+
+const backoff = 3 * time.Second
+
+var ErrProbeTimeout = errors.New("probe did not succeed")
+
+func WaitForEndpoint(endpoint string, timeout time.Duration) error {
```
paologallinaharbur, line 14: could we use the pester client already available with an increased number of retries for this instead of custom code?

cmd/nri-kubernetes/main.go
```diff
@@ -264,7 +264,15 @@ func buildClients(c *config.Config) (*clusterClients, error) {
 	}, nil
 }
 
-func createIntegrationWithHTTPSink(config *config.Config) (*integration.Integration, error) {
+func createIntegrationWithHTTPSink(config *config.Config, logger *log.Logger) (*integration.Integration, error) {
+	endpoint := net.JoinHostPort(sink.DefaultAgentForwarderhost, strconv.Itoa(config.Sink.HTTP.Port))
+
+	logger.Info("Waiting for agent container to be ready...")
+	err := sink.WaitForEndpoint(fmt.Sprintf("http://%s%s", endpoint, sink.DefaultAgentReadyPath), 2*time.Minute)
```
gsanchezgavier, line 271: is 2m an empiric value? could we reduced to 1m? I just have in mind that using a wrong license key could pop up an error faster

src/sink/probe.go
gsanchezgavier: +1

cmd/nri-kubernetes/main.go
roobre: Reduced to 1m, plus it's now configurable :)

src/sink/probe.go
roobre: I tried to do so, but faced two blockers:
1. Using pester directly (e.g. not probing) does not fit very well with the problem we're trying to solve: Agent might take a few to start, but after that should respond promptly and with minimal retries
2. Pester does not easily allow to configure a global timeout, which we do want to do here.

The `for` loop is a bit custom, but I think it's pretty simple and (now) it's tested :)

src/integration/prober/prober.go
```diff
@@ -0,0 +1,60 @@
...
+func (p *Prober) attempt(url string) error {
+	resp, err := http.Get(url)
+	if err != nil {
+		return fmt.Errorf("probe attempt to %s failed, retrying in %s: %w", url, p.backoff, err)
+	}
+
+	_ = resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("%w: %d", errProbeNotOk, resp.StatusCode)
```
gsanchezgavier, line 62: perhaps you can get rid of the `errProbeNotOk` var since is just for loggin

src/integration/prober/prober.go
```diff
@@ -0,0 +1,66 @@
...
+		backoff: backoff,
+		Logger:  logutil.Discard,
+	}
+}
+
+// Probe repeatedly hits the specified url with a GET request every Prober.backoff, and blocks until a request returns
+// 200, or Prober.timeout passes.
+func (p *Prober) Probe(url string) error {
+	start := time.Now()
+	for {
```
gsanchezgavier, line 37: backoff is missing

gsanchezgavier (commented):
LGTM

gsanchezgavier (approved):
Looks good! thanks for adding the exported Func comments

internal/config/config.go
```diff
@@ -10,8 +10,8 @@ import (
 const (
 	DefaultFileName = "nri-kubernetes"
 	DefaultFilePath = "/etc/newrelic-infra"
-	DefaultTimeout  = 10000 * time.Millisecond
-	DefaultRetries  = 4
+	DefaultTimeout  = 3 * time.Second
```
paologallinaharbur, line 13: I would keep it 10 seconds. The charts use that for contacting kubelet and ksm and I do not see a drawback in having it "longer". I believe it can save random restart of the pod not really needed

src/integration/sink/sink.go
```diff
@@ -15,13 +15,13 @@ const (
...
-// Doer is the interface that httpSink client should satisfy.
+// Doer is the interface that HTTPSink client should satisfy.
 type Doer interface {
 	Do(req *http.Request) (*http.Response, error)
 }
 
-// httpSink holds the configuration of the HTTP sink used by the integration.
-type httpSink struct {
+// HTTPSink holds the configuration of the HTTP sink used by the integration.
+type HTTPSink struct {
```
paologallinaharbur, line 24: why now it is exported? Should we return this instead of the interface from .New()?

src/integration/wrapper.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+}
+
+// WithProbeBackoff configures the time the internal prober waits between HTTP endpoint checks.
+func WithProbeBackoff(backoff time.Duration) OptionFunc {
+	return func(i *Wrapper) error {
+		i.probeBackoff = backoff
+		return nil
+	}
+}
```
paologallinaharbur, line 57: I am not a fan of adding unused code just in case it is going to be useful in the future

src/integration/prober/prober.go
roobre: I planned to do so but I'm not sure if the linter would complain in this case. Since I already had the var created from a previous iteration where this error was somewhere returned) I left it there, but we can remove it if you want!

internal/config/config.go
roobre: You're right, for a second I though I was changing only the agent timeout. I'll revert this

src/integration/sink/sink.go
roobre: I think this was due to some refactor I did not fully complete. I prefer exported types so I'll change the return type for `New` :)

src/integration/wrapper.go
roobre: When I added it I planned to add the relevant config options, but in the end I never did.

Do you think there is some value in adding these to the config?

internal/config/config.go
roobre: Done in https://github.com/newrelic/nri-kubernetes/pull/346/commits/bbe936630551e15c66ff7efa6c1e049963b85645.
I kept the retries to 3 though, 4 with 10s mean waiting 40s which seems a bit excessive in my view

src/integration/wrapper.go
paologallinaharbur: mmm not really, at least I cannot imagine a scenario in which a customer want to set those up

roobre: Done, got rid of them :)

paologallinaharbur (approved):

PR #347, opened 2022-01-27
scrape/job: log recoverable errors as Trace

roobre:
So-called recoverable "errors" are expected to happen, as it is nearly impossible for a cluster to have all possible metrics available.

This PR changes log level of these errors to `Trace`.


paologallinaharbur (approved):
Thanks for this, logs will improve a lot

PR #351, opened 2022-02-01
feat(ci/cd): main build

paologallinaharbur:
Using bash if-else since github action is not supporting if-else statements.

### DIFF
--- a/.github/workflows/nightlyAndMainImage.yml
+++ b/.github/workflows/nightlyAndMainImage.yml
@@ -1,14 +1,11 @@
-name: Nightly build
+name: Nightly and Main build
 
 on:
   schedule:
     - cron: "0 4 * * *"
   push:
     branches:
-      - master
       - main
-# TODO once develop is merged this will be not needed anymore
-      - develop
 
 jobs:
   build:
@@ -50,11 +47,13 @@ jobs:
     runs-on: ubuntu-latest
     env:
       DOCKER_IMAGE_NAME: newrelic/nri-kubernetes
-      DOCKER_IMAGE_TAG: nightly
+      DOCKER_IMAGE_TAG: main
       DOCKER_PLATFORMS: "linux/amd64,linux/arm64,linux/arm" # Must be consistent with the matrix from the job above
     steps:
       - uses: actions/checkout@v2
-
+      - name: Setting nightly tag if schedule
+        if: ${{ github.event.schedule }}
+        run: echo "DOCKER_IMAGE_TAG=nightly" >> $GITHUB_ENV
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v1
       - name: Set up Docker Buildx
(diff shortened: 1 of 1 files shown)


.github/workflows/nightlyAndMainImage.yml
```diff
@@ -50,11 +47,17 @@ jobs:
...
     env:
       DOCKER_IMAGE_NAME: newrelic/nri-kubernetes
-      DOCKER_IMAGE_TAG: nightly
       DOCKER_PLATFORMS: "linux/amd64,linux/arm64,linux/arm" # Must be consistent with the matrix from the job above
     steps:
       - uses: actions/checkout@v2
-
+      - name: Setting Docker image tag
+        run: |
+          if [[ "${{ github.ref_name }}" == "main" ]]
```
roobre, line 55: What is `ref_name` and how is it used to distinguish cron/merge?

.github/workflows/nightlyAndMainImage.yml
```diff
@@ -50,11 +47,17 @@ jobs:
...
+          then
+          echo "DOCKER_IMAGE_TAG=main" >> $GITHUB_ENV
+          else
+          echo "DOCKER_IMAGE_TAG=nightly" >> $GITHUB_ENV
+          fi
```
roobre, line 60: I would use a more compact syntax and indentation

```suggestion
          if [[ "${{ github.ref_name }}" == "main" ]]; then
            echo "DOCKER_IMAGE_TAG=main" >> $GITHUB_ENV
          else
            echo "DOCKER_IMAGE_TAG=nightly" >> $GITHUB_ENV
          fi
```

.github/workflows/nightlyAndMainImage.yml
paologallinaharbur: somehow I assumed without even thinking about that, that the ref_name was going to be different. The assomption should be completely wrong. 🙇 

I will check that

roobre: In other workflows I've used `github.event`, as will have an object with the name of the event type.

For example, `${{ if github.event.schedule }}` should evaluate to true if the event runs on schedule:

https://github.com/newrelic/nri-kubernetes/blob/a0ed2ac4810af825bd8b9f69f3fdd1692ad7c697/.github/workflows/security.yaml#L23

roobre: I'm not sure how that could be put in a bash `if`... but I guess we can always have two steps setting the variable, and `skip` based on that :hack:

paologallinaharbur: I added a default and a set that can be skipped

paologallinaharbur: @roobre It feels like a bit an hack, let me know what you think

.github/workflows/nightlyAndMainImage.yml
```diff
@@ -1,14 +1,11 @@
-name: Nightly build
+name: Nightly and Main build
```
roobre, line 1:
```suggestion
name: Nightly and main build
```

roobre (approved):

PR #349, opened 2022-01-28
e2e: also run tests with containerd

roobre:
It's free real estate!

PR #353, opened 2022-02-01
ci: increase e2e parallelism to 5

roobre:
With the addition of containrd scenarios in #349 , the amount of parallel jobs has fallen behind half of the total number of jobs. This commit fixes that.


gsanchezgavier (approved):
LGTM

PR #357, opened 2022-02-08
ci: use reusable workflows

nadiamoe:
This PR attempts to deduplicate the logic present in multiple workflow files using [reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows), which are sort of included into other workflows.

Here I introduce two reusable workflows, one for building the integration, and another one for bulding the docker image. The push/pr workflow has been modified to use them.

### Future steps:

- Reuse both for security scanning jobs
- Reuse `container.reusable.yml` for the release workflow (outscoped from this PR due to requiring more testing)
- Push per-branch image to ghcr.io using `container.reusable.yml`


paologallinaharbur:
I really like the idea, the only issue if the line count that I would expect to be the opposite:
Screen Shot 2022-02-09 at 3 27 14 PM

roobre:
> I really like the idea, the only issue if the line count that I would expect to be the opposite:

Yes, definitely. There are three reasons for this:
1. That now all inputs to the jobs are explicitly declared and described, rather than just being env vars
2. That in this PR I also add a job to build the docker image in the push/pr flow, which wasn't there before
3. More importantly, that I still haven't repalaced other instances of these, like in the release workflow.

I though it would be better to start introducing these in the low-risk workflows and then hop to the release one in a different one :)

paologallinaharbur:
The I like the approach! But tests are still failing 😄

.github/workflows/container.reusable.yml
```diff
@@ -0,0 +1,91 @@
...
+        run: |
+          docker buildx build --push --platform="${{ inputs.platforms }}" \
+            -t "${{ inputs.name }}:${{ steps.tag.outputs.name }}" \
+            .
+      - name: Push :latest images
+        if: ${{ inputs.push && inputs.latest }}
+        run: |
+          docker buildx build --push --platform=${{ inputs.platforms }} \
+            -t ${{ inputs.name }}:latest \
+            .
```
kang-makes, line 91:
```suggestion
      - name: Push versioned images
        if: ${{ inputs.push && ! inputs.latest }}
        run: |
          docker buildx build --push --platform="${{ inputs.platforms }}" \
            -t "${{ inputs.name }}:${{ steps.tag.outputs.name }}" \
            .

      - name: Push versioned images :latest images
        if: ${{ inputs.push && inputs.latest }}
        run: |
          docker buildx build --push --platform=${{ inputs.platforms }} \
            -t "${{ inputs.name }}:${{ steps.tag.outputs.name }}" \
            -t ${{ inputs.name }}:latest \
            .
```

It should be faster and does not test if layers are cached or not.

.github/workflows/container.reusable.yml
```diff
@@ -0,0 +1,91 @@
...
+
+          docker buildx build $load --platform=${{ inputs.platforms }} \
+            -t ${{ inputs.name }}:${{ steps.tag.outputs.name }} \
+            .
+
+      - uses: docker/login-action@v1
+        if: ${{ inputs.push }}
+        with:
+          username: ${{ secrets.FSI_DOCKERHUB_USERNAME }}
+          password: ${{ secrets.FSI_DOCKERHUB_TOKEN }}
```
kang-makes, line 78: Would we add these two as an `input` too?

.github/workflows/container.reusable.yml
```diff
@@ -0,0 +1,91 @@
...
+      - name: Build container images
+        run: |
+          load=""
+          if [[ "${{ inputs.load}}" == "true" ]]; then
+            load="--load"
+          fi
+
+          docker buildx build $load --platform=${{ inputs.platforms }} \
+            -t ${{ inputs.name }}:${{ steps.tag.outputs.name }} \
+            .
```
kang-makes, line 72: Which is the point of this step? Also misleading step name.

.github/workflows/push_pr.yaml
```diff
@@ -4,36 +4,30 @@ on:
...
-      - name: Build integration
-        env:
-          GOOS: ${{ matrix.goos }}
-          GOARCH: ${{ matrix.goarch }}
-        run: |
-          make compile
+    # Should be just:
+    # uses: ./.github/workflows/build.reusable.yml
+    # but https://github.com/github/feedback/discussions/10679
+    uses: newrelic/nri-kubernetes/.github/workflows/build.reusable.yml@ci/reusable-workflows
```
kang-makes, line 13: This is already solved. You can remove it if you wish.
```suggestion
    uses: ./.github/workflows/build.reusable.yml
```

.github/workflows/push_pr.yaml
```diff
@@ -4,36 +4,30 @@ on:
...
+    with:
+      upload: true
+      artifact: nri-kubernetes-${{ github.sha }}
+
+  build-container:
+    # Should be just:
+    # uses: ./.github/workflows/build.reusable.yml
+    # but https://github.com/github/feedback/discussions/10679
+    uses: newrelic/nri-kubernetes/.github/workflows/container.reusable.yml@ci/reusable-workflows
```
kang-makes, line 22: 🔝 
```suggestion
    uses: ./.github/workflows/build.reusable.yml
```

kilokang (commented):
I left some comments. I would like to have this merge and don't stale because the reusability of this is huge and we could have more readable pipelines.

Have you thought about moving the reusable snippets out of `workflows`? We could have pipelines in `.github/workflows` and reusable snippets (for example) in `.github/reusable` so they are not mixed.

.github/workflows/container.reusable.yml
roobre: Is the only change here the `if` line? Github is not highlighting anything and my eyes are old.

If it is, `inputs.latest` is intended to mean "push `:latest` _too_", not "push `:latest` only".

.github/workflows/container.reusable.yml
roobre: Definitely, good point!

.github/workflows/container.reusable.yml
roobre: It is basically ported from the previous pipeline, but the idea is to have an step in which you can load the image out from the `buildx` container. This is required to run a security scan on it, for example.

That being said, it is perhaps better to just have the `--load` argument always present and just condition the job with `if: ${{ inputs.load }}`. WDYT?

.github/workflows/push_pr.yaml
roobre: It is not :(
https://github.com/newrelic/nri-kubernetes/actions/runs/1916515758

roobre:
Marking this as draft as it would seem the reusable workflows features is broken again

.github/workflows/container.reusable.yml
kang-makes: The `if` line and that it is adding two `--target`/`-t` in the latest.

Having them separated into two commands can lead to cache miss and rebuilding containers twice, having a `latest` container virtually different to the one with a `vX.Y.Z` tag.

.github/workflows/container.reusable.yml
kang-makes: I leave it to you because I do not see the point of having to load other containers (without specifying the container or providing a `.tar.gz`) on the CI pipeline.

If you see it useful, go for it. But I don't see it used anywhere so I don't see a point in having a feature we are not using.

.github/workflows/push_pr.yaml
kang-makes: This is not a thing on the relative path but with the lack of `@main` which is even worse.

svetlanabrennan:
Stale pr: closing

PR #362, opened 2022-02-16
integration: allow to specify a custom sink

roobre:
When debugging, it is often helpful to manually inspect the payload of the integration. This PR adds a config flag that will cause the integration to use stdout as a sink, rather than the infra agent.

This will also skip the agent pod readiness check.

### DIFF
--- a/cmd/nri-kubernetes/main.go
+++ b/cmd/nri-kubernetes/main.go
@@ -31,6 +31,7 @@ const (
 
 	_ = iota
 	exitClients
+	exitConfig
 	exitIntegration
 	exitLoop
 	exitSetup
@@ -54,7 +55,7 @@ type clusterClients struct {
 func main() {
 	logger = log.StandardLogger()
 
-	c, err := config.LoadConfig(config.DefaultFilePath, config.DefaultFileName)
+	c, err := config.LoadConfig(config.DefaultConfigFolderName, config.DefaultConfigFileName)
 	if err != nil {
 		log.Error(err.Error())
 		os.Exit(exitIntegration)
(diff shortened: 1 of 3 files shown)


cmd/nri-kubernetes/main.go
```diff
@@ -73,19 +73,28 @@ func main() {
...
 		integration.WithLogger(logger),
 		integration.WithMetadata(integration.Metadata{
 			Name:    integrationName,
 			Version: integrationVersion,
 		}),
-	)
+	}
+
+	fileInfo, err := os.Stdout.Stat()
+	if err == nil && fileInfo.Mode()&os.ModeCharDevice != 0 {
```
paologallinaharbur, line 85: a bit too hacky, no? 😅  could we add a parameter like HTTPsink.enable?

roobre: Fair enough, I think a config flag + env var would be more easier to understand.

roobre: Done!

cmd/nri-kubernetes/main.go
```diff
@@ -73,19 +74,31 @@ func main() {
...
 			Version: integrationVersion,
 		}),
-	)
+	}
+
+	switch c.Sink.Type {
+	case config.SinkTypeHTTP:
+		integrationOptions = append(integrationOptions, integration.WithHTTPSink(c.Sink.HTTP))
+	case config.SinkTypeStdout:
+		logger.Warn("Sinking metrics to stdout")
```
paologallinaharbur, line 89: should we specify that this is no-op since the dafault is alreadty stdOUT?

paologallinaharbur (approved):

cmd/nri-kubernetes/main.go
roobre: Fair enough :)

PR #363, opened 2022-02-16
config: document options

roobre:


internal/config/config.go
```diff
@@ -16,90 +16,183 @@ const (
...
+	// integration to exit with a non-zero code.
+	StaticEndpoint *Endpoint `mapstructure:"staticEndpoint"`
+	// Autodiscover contains one or more criteria for discovering control plane endpoints. The first criteria that
+	// matches an endpoint will be attempted.
+	// If an endpoint matching the criteria is discovered but connection fails, the integration will skip to the next
+	// endpoint.
+	// The list of discovered endpoints is exhausted, either because of failure or because it is empty, the integration
+	// will understand there is no control plane component discoverable _at the current point in time_, and will keep
+	// probing for endpoints periodically.
+	Autodiscover []AutodiscoverControlPlane `mapstructure:"autodiscover"`
```
roobre, line 149: @gsanchezgavier Can you take a look at this statements and ensure they are correct? I wrote this from the top of my head and I do remember that we changed the approach a couple times here.

paologallinaharbur:
👏 👏 👏 I'll check it out tomorrow! But it looks awesome!

internal/config/config.go
gsanchezgavier: this is correct but i would clarify the difference between the Pod discovery and the endpoint probing. I made a suggestion below , feel free to take something or modify it

> The component will iterate over the Autodiscover entries looking for the first Pod that matches the criteria  (selector, namespace and matchNode). Then it probes the listed endpoints for that Autodiscover until the first one successfully responding.
If no Pod is discovered or non of the listed endpoints probes succeeds the he integration will understand there is no control plane component discoverable at the current point in time, and will keep probing.

roobre: I'm still missing some bits, I think. Let me write a series of use cases here and we can check if they are correct:
| Scenario | Result |
| ---------|---------|
| The first autodiscovery entry does not find a matching pod | The integration will try the next autodiscovery entry |
| None of the autodiscovery entries find a matching pod | The integration will keep probing for pods in case they appear later |
| All the endpoints configured for the first autodiscovery entry, which did find a pod, fail | ?? |

paologallinaharbur (approved):
Thanks for taking are of this, apart from what guillermo is pointing out everything else looked good

internal/config/config.go
gsanchezgavier: for that last one it fails https://github.com/newrelic/nri-kubernetes/blob/main/src/controlplane/scraper.go#L255

gsanchezgavier: so my explanation was wrong I will edit that comment

gsanchezgavier: 🤦 i'm wrong , i was right at the begging , this is where that error is handled . sorry for the mess 
https://github.com/newrelic/nri-kubernetes/blob/main/src/controlplane/scraper.go#L264

roobre: I'm still a bit confused here 😅 
One more relevant point: https://github.com/newrelic/nri-kubernetes/blob/1e3231333f92f311afc7d1bcd2fbbb3855670293/src/controlplane/discoverer/discoverer.go#L92-L92

roobre: @gsanchezgavier So what would be the behavior for the third scenario in the table?

gsanchezgavier: in that case it will not fail but it will not try with the rest of autodiscovery entries.

roobre: That's a bit weird, we might want to revisit that. I'll document it for now until we change it.

gsanchezgavier: sure we can review it, not sure if there is a case where multiple entires could match in a cluster

roobre: Updated the comment accordingly, please LMK your thoughts @gsanchezgavier :)

internal/config/config.go
```diff
@@ -8,98 +8,201 @@ import (
...
+	// will contact this endpoint directly instead.
+	// Please note that failure to connect to a StaticEndpoint is considered a fatal error and will cause the
+	// integration to exit with a non-zero code.
+	StaticEndpoint *Endpoint `mapstructure:"staticEndpoint"`
+	// Autodiscover contains one or more criteria for discovering control plane endpoints. Entries will be iterated in
+	// order, with the following rules:
+	// 1. If an entry's criteria (Selector, Namespace, MatchNode) does not match any pod, the next entry will be tried.
+	// 2. If none of the entries matches any pod, the integration will not error but keep probing in case matching pods appear.
+	// 3. If an entry's criteria more than one pod, only the first match will be considered.
+	// matches an endpoint will be attempted.
```
gsanchezgavier, line 156: looks like this line was from an old version.

internal/config/config.go
```diff
@@ -8,98 +8,201 @@ import (
...
+	// 4. Endpoints are tried in order for a matching pod, until metrics can be read successfully from one of them.
+	// 5. If all endpoints for a matching fail, no more entries will be processed, and the integration will error.
```
gsanchezgavier, line 158: the integration will not error in this case, what would happen is the same as point 2, the integration will start a new autodiscover cicle on the next iteration.

gsanchezgavier (approved):

PR #364, opened 2022-02-17
Dockerfile: simplify binary path, ensure 755 permissions

roobre:
Since the integration is now a standalone binary, there is no real reason for the binary to be placed in such a complicated path.

As for the chmod, `ADD --chmod` is supported only when using BuildKit. Since we already need to perform a `RUN mv`, using a `chmod` command instead removes the BuildKit requirement while not adding any extra layer to he image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,16 +3,16 @@ FROM alpine:3.15.0
 ARG TARGETOS
 ARG TARGETARCH
 
-RUN apk add --no-cache --upgrade && apk add --no-cache tini=0.19.0-r0 curl bind-tools
+RUN apk add --no-cache --upgrade && apk add --no-cache tini curl bind-tools
 
-ADD --chmod=755 bin/nri-kubernetes-${TARGETOS}-${TARGETARCH} /var/db/newrelic-infra/newrelic-integrations/bin/
+ADD bin/nri-kubernetes-${TARGETOS}-${TARGETARCH} /bin/
 
-RUN mv /var/db/newrelic-infra/newrelic-integrations/bin/nri-kubernetes-${TARGETOS}-${TARGETARCH} \
-       /var/db/newrelic-infra/newrelic-integrations/bin/nri-kubernetes
+RUN mv /bin/nri-kubernetes-${TARGETOS}-${TARGETARCH} /bin/nri-kubernetes && \
+    chmod 755 /bin/nri-kubernetes
 
 # creating the nri-agent user used only in unprivileged mode
 RUN addgroup -g 2000 nri-agent && adduser -D -u 1000 -G nri-agent nri-agent
 
 USER nri-agent
 
-ENTRYPOINT ["/sbin/tini", "--", "/var/db/newrelic-infra/newrelic-integrations/bin/nri-kubernetes"]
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -3,16 +3,16 @@ FROM alpine:3.15.0
 ARG TARGETOS
 ARG TARGETARCH
 
-RUN apk add --no-cache --upgrade && apk add --no-cache tini=0.19.0-r0 curl bind-tools
+RUN apk add --no-cache --upgrade && apk add --no-cache tini curl bind-tools
 
-ADD --chmod=755 bin/nri-kubernetes-${TARGETOS}-${TARGETARCH} /var/db/newrelic-infra/newrelic-integrations/bin/
+ADD bin/nri-kubernetes-${TARGETOS}-${TARGETARCH} /bin/
```
kang-makes, line 8: Why do we not add the file, rename it and change its mode in the same layer/step?
```suggestion
ADD --chmod=755 bin/nri-kubernetes-${TARGETOS}-${TARGETARCH} /bin/nri-kubernetes
```

roobre: To clarify on the cryptic comment: Last time we checked it, doing what you suggest caused the file to end up as:
```
/bin/nri-kubernetes/nri-kubernetes-linux-amd64
```
And it happened only on the CI environment, IIRC. But please do a PR, this behavior might have changed.

roobre:
> Are we sure we want to change the path? This will make the integration be "not standard" to the rest of the integrations.

I think this makes sense as this is not anymore a standard integration. It does not behave like the rest, and the agent does not need to be able to find it by any means.

Dockerfile
```diff
@@ -3,16 +3,16 @@ FROM alpine:3.15.0
...
+RUN mv /bin/nri-kubernetes-${TARGETOS}-${TARGETARCH} /bin/nri-kubernetes && \
+    chmod 755 /bin/nri-kubernetes
 
 # creating the nri-agent user used only in unprivileged mode
 RUN addgroup -g 2000 nri-agent && adduser -D -u 1000 -G nri-agent nri-agent
 
 USER nri-agent
 
-ENTRYPOINT ["/sbin/tini", "--", "/var/db/newrelic-infra/newrelic-integrations/bin/nri-kubernetes"]
+ENTRYPOINT ["/sbin/tini", "--", "/bin/nri-kubernetes"]
```
gsanchezgavier, line 18: would just `nri-kubernetes` work since bin is in the PATH ?

paologallinaharbur: I would prefer absolute path anyway. It just makes me feel safe knowing exactly which binary is executed

Dockerfile
kang-makes: `DOCKER_BUILDKIT=1` do really weird things... I asked out of curiosity because I thought that should be the intended behavior.

Maybe there are even more differences between `ADD` and `COPY` than I expected.

Dockerfile
roobre: IIRC tini was not very happy with relative paths, but might be mistaken.

paologallinaharbur (approved):

PR #376, opened 2022-03-16
chore(goVersion): test 1.18

paologallinaharbur:
Testing 1.18 verifying cpu improvements


paologallinaharbur:
Comparing cpu consumption in e2e is totally useless, we should do it properly if we are interested into that info

roobre:
@paologallinaharbur if you can send me an image with this I can deploy it side-to-side in my "canaries" to check it :)

PR #377, opened 2022-03-16
[CI/CD] lint, test, release chart

paologallinaharbur:
- [x] Moving e2e charts to /.charts
 - [x] adapt pipelines
 - [x] Test implementation in fork


.github/workflows/release-integration.yml
```diff
@@ -91,18 +77,21 @@ jobs:
...
-          exit-code: '0'
+          exit-code: '1'
           ignore-unfixed: true
           severity: 'CRITICAL,HIGH'
 
-      - name: Push versioned images
+      - name: Push pre-release images
+        if: ${{ github.event.release.prerelease }}
         run: |
+          echo "DOCKER_IMAGE_TAG=${DOCKER_IMAGE_TAG}-pre" >> $GITHUB_ENV
```
kang-makes, line 87: I know that you have tested in a fork but are we sure that every time you "touch" github's envfile the environment is instantaneously updated?

I had the feeling that is not happening and that is the reason I [used the environment variable directly.](https://github.com/newrelic/newrelic-k8s-metrics-adapter/pull/40/files#diff-2c3b5a007a3ca5391253871b413bf6d78192fe551b28bdfe115df67aa07528b1R78) and it is working:
 - From prod repo: https://hub.docker.com/layers/newrelic-k8s-metrics-adapter/newrelic/newrelic-k8s-metrics-adapter/0.3.0-pre/images/sha256-3436a0523cf0394b9e1f373976fb80436aa3f053d8b42eab4e9a39bac3411b2d?context=explore
 - From fork repo: https://hub.docker.com/r/jperezflorez123/newrelic-k8s-metrics-adapter/tags

.gitignore
```diff
@@ -5,5 +5,5 @@ bin
 nri-kubernetes.yaml
 nri-kubernetes.yml
 
-e2e/charts/e2e-resources/Chart.lock
-e2e/charts/e2e-resources/charts/
+charts/e2e-resources/Chart.lock
+charts/e2e-resources/charts/
```
kang-makes, line 9: It would be awesome that this change had its own commit. Reviewing this would be easier ;)

e2e/charts/e2e-resources/Chart.yaml
```diff
@@ -1,10 +0,0 @@
-apiVersion: v2
-version: 1.2.0
-appVersion: "1.0.0"
-description: This chart creates e2e resources.
-name: e2e-resources
-
-dependencies:
-  - name: kube-state-metrics
-    version: 2.13.2
-    repository: https://kubernetes.github.io/kube-state-metrics
```
kang-makes, line 10: Why is this deleted? Is this not going to make this chart uninstallable?

kang-makes: Ah. It has so many changes that git thinks is a new file: https://github.com/newrelic/nri-kubernetes/pull/377/files#diff-a8dbaa85f24d99fa37816cd13be16708b3eab7f9b9fd58fb2908d1ff073f5b6aR1-R24

Ok.

charts/e2e-resources/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v2
+version: 1.2.0
+appVersion: "1.0.0"
```
kang-makes, line 3: In case we release it by mistake. `appVersion` is not mandatory.
```suggestion
version: 1.2.0-devel
```

.github/workflows/release-chart.yml
```diff
@@ -0,0 +1,48 @@
...
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      # As of today, there is no way to filter/exclude charts to be released. Config does not support it:
+      # https://github.com/helm/chart-releaser/blob/4eb598f96aa53d1c0b8234af2087834e1f2275e1/pkg/config/config.go#L40-L61
+      # So we delete the Chart.yaml to filter the chart: https://github.com/helm/chart-releaser-action/pull/18
+      - name: remove the chart template as it should not be deployed
+        run: |
+          rm ./charts/e2e-resources/Chart.yaml
```
kang-makes, line 38: Would we want to release this with a `-devel` or something? It would be useful to have something like this in the canaries someday?

I would leave this as you have it right now until we need it, just asking myself.

.github/workflows/release-integration.yml
```diff
@@ -39,46 +39,32 @@ jobs:
...
-      - name: Generate docker image version from git tag
-        run: |
-          echo "${{ github.event.release.tag_name }}" | grep -E '^v[0-9.]*[0-9]$'
-          DOCKER_IMAGE_TAG=$(echo "${{ github.event.release.tag_name }}" | sed 's/^v//')
-          echo "DOCKER_IMAGE_TAG=$DOCKER_IMAGE_TAG" >> $GITHUB_ENV
-
-      - if: ${{ github.event.release.prerelease }}
-        run: |
-          echo "DOCKER_IMAGE_TAG=${DOCKER_IMAGE_TAG}-pre" >> $GITHUB_ENV
-          echo "S3_PATH=$S3_PATH/test" >> $GITHUB_ENV
```
kang-makes, line 59: Docker `-pre` suffix has been moved to a step below. What about this `S3_PATH`?

kang-makes (commented):
I left a few comments and doubts here and there. It is a pretty good job overall.

.gitignore
paologallinaharbur: image

.github/workflows/release-integration.yml
paologallinaharbur: S3_PATH is not useful anymore, since we are not publishing things in S3

.github/workflows/release-integration.yml
paologallinaharbur: I think it does, a race condition like that would be horrible, but you are right, it is better to use classical Env Var since we are in the same step!

.github/workflows/release-chart.yml
paologallinaharbur: Yes we could do it, but since it is not currently needed I will keep it like that.

paologallinaharbur:
I rebased and moved around things, but the changes with respect to the old pipeline are just:
```git
index 872250d..c2a4585 100644
--- a/.github/workflows/release-integration.yml
+++ b/.github/workflows/release-integration.yml
@@ -84,7 +84,7 @@ jobs:
       - name: Push pre-release images
         if: ${{ github.event.release.prerelease }}
         run: |
-          echo "DOCKER_IMAGE_TAG=${DOCKER_IMAGE_TAG}-pre" >> $GITHUB_ENV
+          DOCKER_IMAGE_TAG=${DOCKER_IMAGE_TAG}-pre"
           docker buildx build --push --platform=$DOCKER_PLATFORMS \
             -t $DOCKER_IMAGE_NAME:$DOCKER_IMAGE_TAG \
             .
diff --git a/charts/e2e-resources/Chart.yaml b/charts/e2e-resources/Chart.yaml
index ef96b38..49e879c 100755
--- a/charts/e2e-resources/Chart.yaml
+++ b/charts/e2e-resources/Chart.yaml
@@ -1,6 +1,5 @@
 apiVersion: v2
-version: 1.2.0
-appVersion: "1.0.0"
+version: 1.2.0-devel
 name: e2e-resources
```

.github/workflows/push_pr.yaml
```diff
@@ -31,9 +32,43 @@ jobs:
...
+          if [[ -n "$changed" ]]; then
+            echo "::set-output name=changed::true"
+          fi
+      - name: Set up helm-unittest
+        if: steps.list-changed.outputs.changed == 'true'
+        run: |
+          helm plugin install https://github.com/quintush/helm-unittest
+      - name: Run unit tests
+        if: steps.list-changed.outputs.changed == 'true'
+        run: |
```
kang-makes, line 62: Could we do it in the same step?

Like in the "original" one: https://github.com/newrelic/newrelic-k8s-metrics-adapter/pull/40/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR78-R87

.github/workflows/release-chart.yml
```diff
@@ -0,0 +1,48 @@
...
+        uses: actions/checkout@v2
+        with:
+          ref: gh-pages
+      - name: Download remote index file and check equality
+        run: |
+          curl -vsSL https://newrelic.github.io/nri-kubernetes/index.yaml > index.yaml.remote
+          LOCAL="$(md5sum < index.yaml)"
+          REMOTE="$(md5sum < index.yaml.remote)"
+          echo "$LOCAL" = "$REMOTE"
+          test "$LOCAL" = "$REMOTE"
```
kang-makes, line 23: I'd like to pint that this step can fail if there is no `index.yaml` created in the `gh-pages` branch.

But I swa you put [an index.yaml in that branch](https://github.com/newrelic/nri-kubernetes/commit/1ae507d0c2572740219614886f53ca28948c3289). Are we completely sure that this is the bare minimum for `cr` to work? Because I put a completely empty `index.yaml` (`touch index.yaml && git add -A && git push`) and it failed.

I would change my PR on the metrics adapter if you confirm me that this works :)

kang-makes (approved):
A couple of nits but LGTM.

Separating the commit about changing the e2e chart location made the re-review way easier. Good job!

.github/workflows/release-chart.yml
paologallinaharbur: I run 
```
git checkout --orphan gh-pages
 - Deleted everything but the gitignore
helm repo index .
```

In the forked it worked!

.github/workflows/push_pr.yaml
paologallinaharbur: sure!

.github/workflows/release-chart.yml
kang-makes: Really cool. I did not think about using `helm` to create an empty index. Now I feel dumb 😅

.github/workflows/release-chart.yml
```diff
@@ -0,0 +1,48 @@
...
+    runs-on: ubuntu-latest
+    needs: [ validate-gh-pages-index ]
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      # As of today, there is no way to filter/exclude charts to be released. Config does not support it:
+      # https://github.com/helm/chart-releaser/blob/4eb598f96aa53d1c0b8234af2087834e1f2275e1/pkg/config/config.go#L40-L61
+      # So we delete the Chart.yaml to filter the chart: https://github.com/helm/chart-releaser-action/pull/18
```
roobre, line 35: Alternatively, could we move the e2e chart to a different dir? Like `charts-internal` or something along those lines.

I'm trying to avoid footgunning ourselves if we need to do a release manually and we forget that we need to remove this (which we will).

.github/workflows/release-integration.yml
```diff
@@ -91,18 +77,21 @@ jobs:
...
           docker buildx build --push --platform=$DOCKER_PLATFORMS \
             -t $DOCKER_IMAGE_NAME:$DOCKER_IMAGE_TAG \
             .
-      - name: Push :latest images
+      - name: Push release images
         if: ${{ ! github.event.release.prerelease }}
         run: |
           docker buildx build --push --platform=$DOCKER_PLATFORMS \
+            -t $DOCKER_IMAGE_NAME:$DOCKER_IMAGE_TAG \
             -t $DOCKER_IMAGE_NAME:latest \
```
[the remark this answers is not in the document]
paologallinaharbur: this is @kang-makes black magic

.github/workflows/push_pr.yaml
```diff
@@ -31,9 +32,41 @@ jobs:
...
+          for chart in $(ct --config .github/ct.yaml list-changed); do
+            if [ -d "$chart/tests/" ]; then
+              helm unittest -3 $chart
+            fi
+          done
+
   test:
     name: Run tests
-    needs: [ build ]
+    needs: [ build, chart-lint ]
```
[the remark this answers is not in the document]
paologallinaharbur: not at all

.github/workflows/release-chart.yml
```diff
@@ -0,0 +1,48 @@
...
+  validate-gh-pages-index:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          ref: gh-pages
+      - name: Download remote index file and check equality
+        run: |
+          curl -vsSL https://newrelic.github.io/nri-kubernetes/index.yaml > index.yaml.remote
```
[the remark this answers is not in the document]
paologallinaharbur: I would not change it since every-other pipeline has it hardcoded. Moreover, it is impossible not to notice it in case you mistype it since the pipeline would fail

paologallinaharbur:
waiting for the ci to kick in, to merge this

.github/workflows/release-integration.yml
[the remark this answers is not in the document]
kang-makes: These are tags. You can [get variables](https://github.com/txqueuelen/pdns-stateless/blob/main/.github/workflows/periodic-upgrade.yaml#L30-L33) and use it [as many times as you need](https://github.com/txqueuelen/pdns-stateless/blob/main/.github/workflows/periodic-upgrade.yaml#L53-L58)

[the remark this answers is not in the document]
paologallinaharbur: yeah, I actually think it is quite standard to use them 😅  I was not sure if Roberto was kidding or not, thus I accused you of black magic 😄

PR #382, opened 2022-03-21
Feature/restart count delta

### LINKED ISSUE #379: Add `restartCountDelta` metric

Right now, we report pod restarts as a cumulative number:
https://github.com/newrelic/nri-kubernetes/blob/ac95cda44565da012f15ef9249de5089e35a9406/src/metric/definition.go#L938-L938

However, the NR platform traditionally has some problems to work with cumulative counters, and detecting if a pod has been "restarted recently" becomes hard to do with a cumulative metrics.

## Feature Description

It would be nice to have a delta version of the metric, where on each report interval the integration automatically reports the difference in number of restarts w.r.t. the previous interval. This functionality is already built in the SDK and should be enough to add a metric with the `sdkMetric.PDELTA` type.

---
- [x] add new metric in nri-kubernetes
- [x] (internal link) [spec-files PR](https://source.datanerd.us/infrastructure/infra-integration-specs/pull/289) merged
- [x] (internal link) [dirac PR](https://source.datanerd.us/dirac/dirac/pull/6876) merged
- [x] create [issue to include mappings in infra-metrics-utils repo](https://github.com/newrelic/newrelic-coreint/issues/311).

sigilioso:
Closes #379

### DIFF
--- a/src/scrape/scrape_job_test.go
+++ b/src/scrape/scrape_job_test.go
@@ -146,48 +146,49 @@ var kubeletSpecs = definition.SpecGroups{
 	"container": metric.KubeletSpecs["container"],
 }
 
-type testGrouper struct {
-	// PodValuesInTime allows to overwrite some metrics value in each Group execution. Ex: {"isReady": {0, 1, 2, 3}}
-	PodValuesInTime map[string][]interface{}
-	// ContainerValuesInTime allows to overwrite some metrics value in each Group execution. Ex: {"restartCount": {0, 1, 2, 3}}
-	ContainerValuesInTime map[string][]interface{}
+// grouperMock implements Grouper interface returning mocked metrics which might change over subsequent calls.
+type grouperMock struct {
+	// ValuesInGroupCalls overwrite some metrics in subsequent calls using {"<path>": {subsequent-values} Examples:
+	// {"pod.kube-system_newrelic-infra-rz225.isReady": {0, 1, 2, 3}},
+	// {"container.kube-system_newrelic-infra-rz225.restartCount": {0, 1, 2, 3}}
+	ValuesInGroupCalls map[string][]interface{}
 
 	groupCallsCount int
 }
 
-func (tg *testGrouper) Group(definition.SpecGroups) (definition.RawGroups, *data.ErrorGroup) {
-	podData := tg.buildMetrics(
-		testdata.ExpectedGroupData["pod"]["kube-system_newrelic-infra-rz225"],
-		tg.PodValuesInTime,
-	)
-	containerData := tg.buildMetrics(
-		testdata.ExpectedGroupData["container"]["kube-system_newrelic-infra-rz225_newrelic-infra"],
-		tg.ContainerValuesInTime,
-	)
+func (g *grouperMock) Group(definition.SpecGroups) (definition.RawGroups, *data.ErrorGroup) {
 	// We reduce the test fixtures in order to simplify testing.
-	groups := definition.RawGroups{
-		"pod": {
-			"kube-system_newrelic-infra-rz225": podData,
-		},
-		"container": {
-			"kube-system_newrelic-infra-rz225_newrelic-infra": containerData,
-		},
+	groupsDefinition := map[string]string{
+		"pod":       "kube-system_newrelic-infra-rz225",
+		"container": "kube-system_newrelic-infra-rz225_newrelic-infra",
 	}
-	tg.groupCallsCount++
+	groups := definition.RawGroups{}
+	for entityType, entityName := range groupsDefinition {
+		if groups[entityType] == nil {
+			groups[entityType] = map[string]definition.RawMetrics{}
+		}
+		groups[entityType][entityName] = g.metricsForCurrentCall(entityType, entityName)
+	}
+	g.groupCallsCount++
 	return groups, nil
 }
 
-func (tg *testGrouper) buildMetrics(
-	baseData definition.RawMetrics, valuesInTime map[string][]interface{},
-) definition.RawMetrics {
+// metricsForCurrentCall returns a copy of metrics from `testdata.ExpectedGroupData` corresponding to the provided
+// parameters with values changes as configured for the current group call.
+func (g *grouperMock) metricsForCurrentCall(entityType, entityName string) definition.RawMetrics {
 	metrics := definition.RawMetrics{}
-	for k, v := range baseData {
+	for k, v := range testdata.ExpectedGroupData[entityType][entityName] {
 		metrics[k] = v
 	}
-	for k, values := range valuesInTime {
-		if lenValues := len(values); lenValues > 0 {
-			metrics[k] = values[tg.groupCallsCount%lenValues]
+	for rawPath, values := range g.ValuesInGroupCalls {
+		path := strings.Split(rawPath, ".")
+		lenValues := len(values)
+		if len(path) != 3 || path[0] != entityType || path[1] != entityName || lenValues < 1 {
+			continue
 		}
+		metricName := path[2]
+		valueToUse := g.groupCallsCount % lenValues
+		metrics[metricName] = values[valueToUse]
 	}
 	return metrics
 }
@@ -196,7 +197,7 @@ func TestPopulateK8s(t *testing.T) {
 	t.Parallel()
 	intgr := testutil.NewIntegration(t)
 
-	testJob := NewScrapeJob("test", &testGrouper{}, kubeletSpecs)
+	testJob := NewScrapeJob("test", &grouperMock{}, kubeletSpecs)
 
 	k8sVersion := &version.Info{GitVersion: "v1.15.42"}
 	errPopulate := testJob.Populate(intgr, "test-cluster", logutil.Debug, k8sVersion)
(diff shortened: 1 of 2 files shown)


src/scrape/scrape_job_test.go
```diff
@@ -200,6 +233,40 @@ func TestPopulateK8s(t *testing.T) {
 	}
 }
 
+func TestRestartCountDeltaValues(t *testing.T) {
+	intgr, err := integration.New("test", "test", integration.InMemoryStore())
+	assert.NoError(t, err)
+	intgr.Clear()
```
paologallinaharbur, line 239: why `intgr.Clear()` is needed?

src/scrape/scrape_job_test.go
```diff
@@ -200,6 +233,40 @@ func TestPopulateK8s(t *testing.T) {
 	}
 }
 
+func TestRestartCountDeltaValues(t *testing.T) {
+	intgr, err := integration.New("test", "test", integration.InMemoryStore())
```
paologallinaharbur, line 237: If I am not wrong two subsequent calls to integration.New() cause a panic (😢 )
If not really needed I would try to avoid using the integration in tests

src/scrape/scrape_job_test.go
```diff
@@ -144,18 +145,50 @@ var kubeletSpecs = definition.SpecGroups{
...
+func (tg *testGrouper) buildMetrics(
+	baseData definition.RawMetrics, valuesInTime map[string][]interface{},
+) definition.RawMetrics {
+	metrics := definition.RawMetrics{}
+	for k, v := range baseData {
+		metrics[k] = v
+	}
+	for k, values := range valuesInTime {
+		if lenValues := len(values); lenValues > 0 {
+			metrics[k] = values[tg.groupCallsCount%lenValues]
```
paologallinaharbur, line 188: [nit] I would add a comment explaining this, maybe also moving the index calculation to a different line

src/scrape/scrape_job_test.go
paologallinaharbur, line 237: can we add t.parallel()?

paologallinaharbur (commented):
I like the approach, I added just some nits. 
By the way after adding a new metric we need to update the spec file 😢  (I would really prefer the nrql being able to deal with cumulative counts in a better way)

src/scrape/scrape_job_test.go
roobre: If you use `t.Parallel` you'll need to wrap this with a mutex to avoid a race in the SDK. It's not particularly good, but the problem is already solved :)
https://github.com/newrelic/nri-kubernetes/blob/main/internal/testutil/integration.go#L18

src/scrape/scrape_job_test.go
```diff
@@ -144,18 +145,50 @@ var kubeletSpecs = definition.SpecGroups{
 	"container": metric.KubeletSpecs["container"],
 }
 
-type testGrouper struct{}
+type testGrouper struct {
```
roobre, line 148: I'm probably complaining about code that was already here (sorry!), but I'm finding hard to understand what `testGrouper` is. Is it a helper? A mock? What does it do? I'd love to see a godoc comment explaining it :)

src/scrape/scrape_job_test.go
```diff
@@ -144,18 +145,50 @@ var kubeletSpecs = definition.SpecGroups{
...
+	// PodValuesInTime allows to overwrite some metrics value in each Group execution. Ex: {"isReady": {0, 1, 2, 3}}
+	PodValuesInTime map[string][]interface{}
+	// ContainerValuesInTime allows to overwrite some metrics value in each Group execution. Ex: {"restartCount": {0, 1, 2, 3}}
+	ContainerValuesInTime map[string][]interface{}
```
roobre, line 152: Is there a reason to have two identical fields for pod and container specifically? I wonder if we could do this differently so if we need to add values over time for another entity, e.g. `Volume`, we don't need to add another `VolumeValuesInTime` field.

src/scrape/scrape_job_test.go
```diff
@@ -144,18 +145,50 @@ var kubeletSpecs = definition.SpecGroups{
...
-			"kube-system_newrelic-infra-rz225_newrelic-infra": testdata.ExpectedGroupData["container"]["kube-system_newrelic-infra-rz225_newrelic-infra"],
+			"kube-system_newrelic-infra-rz225_newrelic-infra": containerData,
 		},
-	}, nil
+	}
+	tg.groupCallsCount++
+	return groups, nil
+}
+
+func (tg *testGrouper) buildMetrics(
```
roobre, line 179: What does `buildMetrics` do?

src/scrape/scrape_job_test.go
roobre: As integrations are globally, it would be advisable to have unique names for them. Otherwise, other test using `"test", "test"` might cause race/concurrency issues. The Test integration helper does this for you :)

https://github.com/newrelic/nri-kubernetes/blob/main/internal/testutil/integration.go#L23

src/scrape/scrape_job_test.go
```diff
@@ -200,6 +233,40 @@ func TestPopulateK8s(t *testing.T) {
...
+		assert.Empty(t, err.(data.PopulateResult).Errors)
+		time.Sleep(time.Second)
+	}
+
+	for _, entity := range intgr.Entities {
+		if entity.Metadata.Name == "newrelic-infra" {
+			lenExpectedDeltas := len(expectedRestartCountDeltas)
+			require.Equal(t, lenExpectedDeltas, len(entity.Metrics))
+			for i := 0; i < lenExpectedDeltas; i++ {
+				assert.Equal(t, expectedRestartCountDeltas[i], entity.Metrics[i].Metrics["restartCountDelta"])
```
roobre, line 264: I do not completely get how (why?) this is working. `entity.Metrics` is a `MetricSet`. Why do we have `n` `MetricSet`s, where `n` is the number of times we called `Populate`? Is this intended behavior of the code?

src/scrape/scrape_job_test.go
sigilioso: I just extended the mock we had there... Anyway You are right it was not obvious and now which it does more stuff it may confusing. I'll improve naming and maybe include some godocs

src/scrape/scrape_job_test.go
sigilioso: I'm improving docs/naming because if you had to ask... it is not obvious enough ;)

src/scrape/scrape_job_test.go
sigilioso: Thanks! I didn't know. I will fix it [here](https://github.com/newrelic/nri-kubernetes/blob/bc62e845386de80d99d1d5b928fc25d5b72ecd98/src/scrape/scrape_job_test.go#L195) too.

src/scrape/scrape_job_test.go
sigilioso: As it is (or it should be) a brand new integration, I guess it is not needed. I'm digging into it in this and the populate-test we already had. Thanks!

src/scrape/scrape_job_test.go
roobre: Yep, I don't think it needs to be exhaustive since it's unexported and for tests. Oneliners such as `// testGrouper is an implementation of the Grouper interface that returns mocked metrics that might change in subsequent calls.` will be more than enough :)

src/scrape/scrape_job_test.go
sigilioso: Removed here: [bc62e84](https://github.com/newrelic/nri-kubernetes/pull/382/commits/bc62e845386de80d99d1d5b928fc25d5b72ecd98)

src/scrape/scrape_job_test.go
sigilioso: Fixed here: [bc62e84](https://github.com/newrelic/nri-kubernetes/pull/382/commits/bc62e845386de80d99d1d5b928fc25d5b72ecd98)

sigilioso: Added here: [bc62e84](https://github.com/newrelic/nri-kubernetes/pull/382/commits/bc62e845386de80d99d1d5b928fc25d5b72ecd98)

src/scrape/scrape_job_test.go
sigilioso: Done: [50c8a5a](https://github.com/newrelic/nri-kubernetes/pull/382/commits/50c8a5a2a9cebd84bf71f7861bc835b414a5a448)

src/scrape/scrape_job_test.go
sigilioso: After this change [50c8a5a](https://github.com/newrelic/nri-kubernetes/pull/382/commits/50c8a5a2a9cebd84bf71f7861bc835b414a5a448) it is a bit more abstract.

We could go further and extract this mock and the corresponding specs to a _testutils file_ and give them some parameters to make it easier to reduce fixtures as needed in different tests... I would include that change in a different PR.

roobre (approved):
Love the new approach, thanks for addressing the comments! ❤️

PR #383, opened 2022-03-23
Change helm chart to use common-library

kang-makes:
There is a commit per change of the templates from the actual behavior to the one that comes from common-library.

I would recommend any reviewer to go commit by commit so you can progressively see the changes. For example, something like "I change here the `privileged` but not the `securityContext` that I changed in the next commit".

Don't hesitate to leave comments that could be solved in further commits so you don't forget them and see if they survive all the refactoring.

Also you'll see that there is some inconsistency in the naming of helpers. This is intended. The library is prefixed with `common.` and old helpers are prefixed with `newrelic.` and `newrelic.compatibility.` while new ones are prefixed with `nriKuberentes.` to make the follow up easier between existing helpers and new helpers that are been introduced in this PR.

We could tackle this inconsistency after EOLing the compatibility layer if you would or in a last commit changing that naming issue.

To don't have to make people re-review constantly I will tackle all the NIT in the same commit at the end of the review when everybody is happy with it.

Have fun :)


roobre:
If you can include the changes I proposed here a while back it would be amazing :)
https://github.com/newrelic/helm-charts/pull/717

.gitignore
```diff
@@ -1,9 +1,32 @@
...
+*~
+\#*\#
+.\#*
+
+# Vim-related files
+[._]*.s[a-w][a-z]
+[._]s[a-w][a-z]
+*.un~
+Session.vim
+.netrwhist
```
gsanchezgavier, line 32: nit: something I learn a few month back, there is a way to locally ignore this for all projects. https://stackoverflow.com/questions/7335420/global-git-ignore

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
```diff
@@ -24,9 +24,7 @@ spec:
     metadata:
       annotations:
         checksum/nri-kubernetes: {{ include (print $.Template.BasePath "/controlplane/configmap.yaml") . | sha256sum }}
-        {{- if not (include "newrelic.licenseCustomSecretName" .) }}
-        checksum/secret: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
-        {{- end }}
+        checksum/license-secret: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
```
gsanchezgavier, line 28: is this ok? How this will work if there is not custom secret ? I saw previous one not rendering this annotation.

charts/newrelic-infrastructure/templates/controlplane/_tolerations_helper.tpl
```diff
@@ -0,0 +1,25 @@
+{{- /*
+Defaults for controlPlane while keeping then overridable.
```
gsanchezgavier, line 2: 'them" ?

charts/newrelic-infrastructure/templates/controlplane/_tolerations_helper.tpl
```diff
@@ -0,0 +1,25 @@
...
+  effect: "NoExecute"
+{{- end -}}
+
+
+
+{{- /*
+As this chart deploys what it should be three charts to maintain the transition to v3 as smooth as possible.
+This means that this chart has 3 tolerations so a helper should be done per scraper.
+*/ -}}
+{{- define "nriKubernetes.controlPlane.tolerations" -}}
```
gsanchezgavier, line 17: oh this is tricky. So precedence is like CP-toletartion -> chart toleration -> global toleration -> default toleration. 
NIT: Perhaps adding something like this to the comment could help to understand it faster

gsanchezgavier: cool there are test for this !! 👏

gsanchezgavier (commented):
I reach until toleration commit, I will keep looking it later . so far looks great !

charts/newrelic-infrastructure/values.yaml
```diff
@@ -12,23 +12,20 @@ images:
     tag: 1.23.0
     # -- Image for the agent sidecar.
     repository: newrelic/k8s-events-forwarder
-    registry: docker.io
```
paologallinaharbur, line 15: shouldn't we keep `registry:` to show the user they can change that?

charts/newrelic-infrastructure/templates/kubelet/daemonset.yaml
```diff
@@ -99,8 +99,8 @@ spec:
           resources: {{ toYaml . | nindent 12 }}
           {{- end }}
         - name: agent
-          image: "{{ .Values.images.agent.registry }}/{{ .Values.images.agent.repository }}:{{ .Values.images.agent.tag }}"
-          command: [ "/sbin/tini", "--", "/usr/bin/newrelic-infra" ]
```
paologallinaharbur, line 93: why removing tini?

charts/newrelic-infrastructure/templates/ksm/deployment.yaml
```diff
@@ -33,6 +33,11 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with include "common.dnsConfig" . }}
+      dnsConfig:
```
paologallinaharbur, line 37: anything to add into the values for this?

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
```diff
@@ -65,6 +65,9 @@ spec:
               value: {{ get (fromYaml (include "nriKubernetes.controlPlane.agentConfig" .)) "http_server_port" | quote }}
             - name: "NRI_KUBERNETES_CLUSTERNAME"
               value: {{ include "common.cluster" . }}
+            - name: "NRI_KUBERNETES_VERBOSE"
```
paologallinaharbur, line 75: would make sense to have an if statement and add that merely if it is true?

charts/newrelic-infrastructure/templates/_helpers_compatibility.tpl
```diff
@@ -175,6 +175,20 @@ Please set:
 ------
 {{- end -}}
 
+{{- define "newrelic.compatibility.message.customAttributes" -}}
+We still support using custom attributes but we support it as a map and dropped it as a string.
+customAttributes: {{ .Values.customAttributes | quote }}
+
+You should change your values to something like this:
+
+customAttributes:
```
paologallinaharbur, line 184: nice!

charts/newrelic-infrastructure/templates/controlplane/agent-configmap.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.controlPlane.enabled -}}
+{{- if .Values.customAttributes | kindIs "string" }}
+{{- fail ( include "newrelic.compatibility.message.customAttributes" . ) -}}
```
paologallinaharbur, line 3: why having the fail here instead of having it unified in the notes.txt?

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
```diff
@@ -79,10 +79,10 @@ spec:
             - name: NRI_KUBERNETES_VERBOSE
               value: "1"
             {{- end }}
-            {{- with .Values.controlPlane.extraEnv }}
```
paologallinaharbur, line 79: I guess you discussed the name already, *if not* are we sure that it worth to change this?

charts/newrelic-infrastructure/templates/ksm/_agent-config_helper.tpl
```diff
@@ -0,0 +1,20 @@
+{{- /*
+Defaults for ksm's agent config
+*/ -}}
+{{- define "nriKubernetes.ksm.agentConfig.defaults" -}}
+is_forward_only: true
+is_secure_forward_only: true
```
paologallinaharbur, line 6: ditto

charts/newrelic-infrastructure/templates/ksm/_agent-config_helper.tpl
```diff
@@ -0,0 +1,20 @@
...
+overide_host_root: ""  # Typo from here: https://github.com/newrelic/infrastructure-agent/blob/master/pkg/config/config.go#L267
```
paologallinaharbur, line 7: Can you explain further?

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
```diff
@@ -72,14 +85,11 @@ spec:
...
-            - name: NRI_KUBERNETES_VERBOSE
-              value: "1"
-            {{- end }}
-            {{- with .Values.controlPlane.extraEnv }}
+
+            {{- with .Values.controlPlane.env }}
             {{- toYaml . | nindent 12 }}
             {{- end }}
-          {{- with .Values.controlPlane.extraEnvFrom }}
+          {{- with .Values.controlPlane.envFrom }}
```
roobre, line 92: I'm not a big fan of removing the `extra` prefix. Do we have a reason for doing that? Users might get the impression they are fully overriding the `env` section of the pod, while that's not the case: We're appending.

charts/newrelic-infrastructure/templates/secret.yaml
```diff
@@ -1,14 +1,2 @@
...
-  labels:
-    {{- include "newrelic.labels" . | nindent 4 }}
-  name: {{ template "newrelic.fullname" . }}-license
-type: Opaque
-data:
-  licenseKey: {{ include "newrelic.licenseKey" . | b64enc }}
-{{- end }}
-{{- end }}
+{{- /* Common library will take care of creating the secret or not. */}}
+{{- include "common.license.secret" . }}
```
roobre, line 2: Missing newline.

charts/newrelic-infrastructure/values.yaml
roobre: I think what @kang-makes had in mind is to support advanced use cases where users can configure their CRI so unqualified images are redirected to registries that are not `docker.io`. But I think we should leave the entry:
```yaml
registry: 
```

charts/newrelic-infrastructure/values.yaml
```diff
@@ -62,10 +59,9 @@ kubelet:
     - operator: "Exists"
       effect: "NoExecute"
   nodeSelector: {}
-  affinity:
-    nodeAffinity: {}
-  extraEnv: []
-  extraEnvFrom: []
+  affinity: {}
+  env: []
+  envFrom: []
```
roobre, line 64: As said above, I'm not fully in with this change.

charts/newrelic-infrastructure/values.yaml
```diff
@@ -90,24 +86,11 @@ ksm:
   # Advanced users only. Setting this to `false` is not supported and will break the New Relic experience.
   enabled: true
   annotations: {}
-  tolerations:
-    - operator: "Exists"
-      effect: "NoSchedule"
-    - operator: "Exists"
-      effect: "NoExecute"
```
roobre, line 97: This change is definitely not user friendly. Users will think the default is `[]` (no tolerations), while it's not. I'd keep the tolerations here if possible.

charts/newrelic-infrastructure/values.yaml
```diff
@@ -90,24 +86,11 @@ ksm:
...
   nodeSelector: {}
-  affinity:
-    podAffinity:
-      preferredDuringSchedulingIgnoredDuringExecution:
-        - podAffinityTerm:
-            topologyKey: kubernetes.io/hostname
-            labelSelector:
-              matchLabels:
-                app.kubernetes.io/name: kube-state-metrics
-          weight: 100
```
roobre, line 107: As above: I do not like this change, values should be explicit unless we really really need them not to be.

charts/newrelic-infrastructure/values.yaml
roobre: Inb4 "but the global then..."
I strongly believe users using such advanced features should manually go here and set `tolerations` to `null` if they want their global to be honored.

charts/newrelic-infrastructure/values.yaml
```diff
@@ -142,39 +125,20 @@ ksm:
 controlPlane:
   # -- Deploy control plane monitoring component.
   enabled: true
-  # -- Run Control Plane scraper with `hostNetwork` even if `privileged` is set to false.
-  # `hostNetwork` is required for most control plane configurations, as they only accept connections from localhost.
-  unprivilegedHostNetwork: false
```
roobre, line 147: I agree with getting rid of this flag, but we really need to decide what is going to happen with the controlPlane pods if `privileged: false`. Failing is not an option as this is something that can come from uninterative flows (guided install).

roobre: I would even suggest leaving everything related to `privileged` out of this PR and discuss on a different one.

charts/newrelic-infrastructure/values.yaml
```diff
@@ -146,22 +136,7 @@ controlPlane:
...
-          - matchExpressions:
-              - key: node-role.kubernetes.io/controlplane
-                operator: Exists
-          - matchExpressions:
-              - key: node-role.kubernetes.io/etcd
-                operator: Exists
-          - matchExpressions:
-              - key: node-role.kubernetes.io/master
-                operator: Exists
+  affinity: {}
```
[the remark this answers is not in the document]
roobre: Agreed, I definitely dislike this change. If globals are not honored, so be it. Users can fix that by manually setting these to `null`.

charts/newrelic-infrastructure/values.yaml
```diff
@@ -412,22 +376,12 @@ podLabels: {}
...
-
-# -- Security context used in all the containers of the pods
-# When `privileged == true`, the Kubelet scraper will run as root and ignore these settings.
-# @default -- See `values.yaml`
-securityContext:
-  # OpenShift users might want to change the default UID:GID to be inside the allowed range.
-  runAsUser: 1000
-  runAsGroup: 2000
-  allowPrivilegeEscalation: false
-  readOnlyRootFilesystem: true
```
roobre, line 425: `securityContext` is probably the only default worth hiding behind a template, as we want it to change depending on `privileged`.

Regardless, I think we should leave the blank entry:

```yaml
// Very nice coment explaining why these are blank and how defaults change.
podSecurityContext:
containerSecurityContext:
```

roobre (changes requested):

charts/newrelic-infrastructure/templates/controlplane/_tolerations_helper.tpl
kang-makes: I leave this to the nit-picking commit.

charts/newrelic-infrastructure/values.yaml
kang-makes: @roobre hit the point: https://github.com/containers/image/blob/main/docs/containers-registries.conf.5.md#normalization-of-dockerio-references

I don't like that `docker.io` is the de-facto standard but it is that way. You can set in the CRI what registry you want to point to. So the sysadmin could be able to point an `unqualified search registries` to wherever they want.

Solved in [#249f236](https://github.com/newrelic/nri-kubernetes/pull/383/commits/249f236fc5b7dcbed0dd346ff38793e28f14be3f)

charts/newrelic-infrastructure/templates/ksm/deployment.yaml
kang-makes: I would like to add it to the documentation of the common library instead of having to repeat everything in all the charts. WDYT?

charts/newrelic-infrastructure/templates/controlplane/agent-configmap.yaml
kang-makes: I was desperately trying to make [this test work](https://github.com/newrelic/nri-kubernetes/blob/helm/adapt-helm-chart/charts/newrelic-infrastructure/tests/configmap_cp_agent_test.yaml#L110-L140) but I couldn't.

The thing here is that `helm unittest` thinks that the manifest you are trying to test is not in the testing suite if you fail in other manifests/files. So I wanted this very manifest to fail and properly test `failedTemplate` but I couldn't.

I would leave this as it is now until we find out how to test that the manifest fails correctly. WDYT?

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
```diff
@@ -133,10 +133,9 @@ spec:
                   fieldPath: "spec.nodeName"
             {{- end }}
             {{- end }}
-            {{- if (include "newrelic.customAttributes" .) }}
+            {{- with fromYaml ( include "common.customAttributes" . ) }}
             - name: "NRIA_CUSTOM_ATTRIBUTES"
-              value: >-
-                {{- include "newrelic.customAttributes" . | nindent 16 }}
+              value: {{ . | toJson | quote }}
```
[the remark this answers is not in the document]
kang-makes: It is changed a commit later because this is to make progressive changes.

`customAttributes` are handled by the agent config and is removed here: https://github.com/newrelic/nri-kubernetes/pull/383/commits/1c85539e2722fd33671d5fae4becb42fbfbc2b8d#diff-d16718b05aef34f809eb79603c6d3f59c281721e380661893e9b6c95c525e27aL136-L151

I am resolving this thread :)

charts/newrelic-infrastructure/templates/controlplane/_agent-config_helper.tpl
```diff
@@ -0,0 +1,20 @@
+{{- /*
+Defaults for controlPlane's agent config
+*/ -}}
+{{- define "nriKubernetes.controlPlane.agentConfig.defaults" -}}
+is_forward_only: true
+is_secure_forward_only: true
```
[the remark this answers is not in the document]
kang-makes: It worked in the end-to-end because the agent ends adding plugins when the `is_forward_only` is set to true and it does not care about if it is secure or not. [See this snippet.](https://github.com/newrelic/infrastructure-agent/blob/master/pkg/plugins/plugins_linux.go#L42-L59)

So if `is_forward_only` is set to true there is no need to ser `is_secure_forward`.

I have to say that I added using [the `Dockerfile` from the forwarder as an example](https://github.com/newrelic/infrastructure-agent/blob/master/build/container/Dockerfile#L47-L66) to set these environment variables.

As I said in other thread:
> Is also a step to remove the forwarder image and use agent one directly if we want to remove spurious hard-to-follow dependencies.

I can clean this a bit. Fixed in [#1cfa609](https://github.com/newrelic/nri-kubernetes/pull/383/commits/1cfa609a444fa9ffffec3c83b67469243ef5ab56)

I am leaving the thread open in case you want to blame something more :)

charts/newrelic-infrastructure/templates/ksm/deployment.yaml
paologallinaharbur: I fear that users will not check that, I would prefer to have them in the values/readme of the chart itself

charts/newrelic-infrastructure/templates/ksm/_agent-config_helper.tpl
kang-makes: This conversation is being done [here](https://github.com/newrelic/nri-kubernetes/pull/383#discussion_r844038246).

Resolving.

charts/newrelic-infrastructure/templates/ksm/_agent-config_helper.tpl
kang-makes: As I said in other thread:

> Is also a step to remove the forwarder image and use agent one directly if we want to remove spurious hard-to-follow dependencies.

I used [the Dockerfile from the forwarder as an example](https://github.com/newrelic/infrastructure-agent/blob/master/build/container/Dockerfile#L47-L66) to set these environment variables (even if I am still using the forwarder image).

I am closing this thread and we can do the follow-up [here](https://github.com/newrelic/nri-kubernetes/pull/383#discussion_r845220611) if you would.

charts/newrelic-infrastructure/templates/_helpers_compatibility.tpl
```diff
@@ -176,6 +175,20 @@ Please set:
...
+{{- range $k, $v := fromJson .Values.customAttributes -}}
+  {{- $k | nindent 2 }}: {{ $v | quote }}
```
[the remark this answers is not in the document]
kang-makes: I am talking about this [here](https://github.com/newrelic/nri-kubernetes/pull/383#discussion_r844157393).

Resolving.

charts/newrelic-infrastructure/templates/_helpers_compatibility.tpl
```diff
@@ -176,6 +175,20 @@ Please set:
...
+{{- end }}
+
+**NOTE**: If you read above errors like "invalid character ':' after top-level value" or "json: cannot unmarshal string into Go value of type map[string]interface {}" means that the string you have in your values is not a valid JSON, Helm is not able to parse it and we could not show you how you should change it. Sorry.
```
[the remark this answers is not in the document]
kang-makes: I tried to use the method [you put above](https://github.com/newrelic/nri-kubernetes/pull/383#discussion_r844155720) but it failed.

It is really honorable for you to trust the documentation but it does not work that way. Here you have an example:

`% kind create cluster`
```
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.23.4) 🖼
 ✓ Preparing nodes 📦
 ✓ Writing configuration 📜
 ✓ Starting control-plane 🕹️
 ✓ Installing CNI 🔌
 ✓ Installing StorageClass 💾
Set kubectl context to "kind-kind"
You can now use your cluster with:

kubectl cluster-info --context kind-kind

Thanks for using kind! 😊
```
`helm upgrade --install --set cluster=test --set licenseKey=test --set customAttributes="test:" test .`
```
Release "test" does not exist. Installing it now.
Error: execution error at (newrelic-infrastructure/templates/kubelet/daemonset.yaml:25:34): We still support using custom attributes but we support it as a map and dropped it as a string.
customAttributes: "test:"

You should change your values to something like this:

customAttributes:
  Error: "invalid character 'e' in literal true (expecting 'r')"

**NOTE**: If you read above errors like "invalid character ':' after top-level value" or "json: cannot unmarshal string into Go value of type map[string]interface {}" means that the string you have in your values is not a valid JSON, Helm is not able to parse it and we could not show you how you should change it. Sorry.
```

Bug in sprig? In Helm? I don't know. But I also don't want to fight this in this PR.

I leave this open for you to comment your thoughts, feel free to give any other point of view :)

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
```diff
@@ -93,61 +103,52 @@ spec:
...
           imagePullPolicy: {{ .Values.images.forwarder.pullPolicy }}
           securityContext:
-            {{- mustMergeOverwrite .Values.securityContext  (include "newrelic.compatibility.securityContext" . | fromYaml) | toYaml | nindent 12 }}
+            {{- mustMergeOverwrite
+                    (include "common.securityContext.container" . | fromYaml )
+                    (include "newrelic.compatibility.securityContext" . | fromYaml )
+                    | toYaml | nindent 12 }}
           ports:
-            - containerPort: 8001
+            - containerPort: {{ get (fromYaml (include "nriKubernetes.controlPlane.agentConfig" .)) "http_server_port" }}
```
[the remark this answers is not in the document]
kang-makes: You choose. I can roll back if you wish.

charts/newrelic-infrastructure/templates/ksm/_tolerations_helper.tpl
```diff
@@ -0,0 +1,25 @@
+{{- /*
+Defaults for ksm while keeping then overridable.
+*/ -}}
+{{- define "nriKubernetes.ksm.tolerations.defaults" -}}
+- operator: "Exists"
+  effect: "NoSchedule"
+- operator: "Exists"
+  effect: "NoExecute"
+{{- end -}}
```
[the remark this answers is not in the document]
kang-makes: Done in [#13cd6e0](https://github.com/newrelic/nri-kubernetes/pull/383/commits/13cd6e04fdbf96d153b5fcf8f2e6bb49f8f023a5)

charts/newrelic-infrastructure/templates/ksm/_tolerations_helper.tpl
```diff
@@ -0,0 +1,25 @@
...
+
+
+
+{{- /*
+As this chart deploys what it should be three charts to maintain the transition to v3 as smooth as possible.
+This means that this chart has 3 tolerations so a helper should be done per scraper.
+*/ -}}
+{{- define "nriKubernetes.ksm.tolerations" -}}
+{{- if gt (len .Values.ksm.tolerations) 0  -}}
```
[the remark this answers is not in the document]
kang-makes: Because it is an array and I wanted to test that it is an array nobody knows why now and make Helm to yell an error. 

I have tested it and it is not a nice message... Fixed in [#c917114](https://github.com/newrelic/nri-kubernetes/pull/383/commits/c91711476f8b554deeb65fb1c9b628ffb59a4770)

charts/newrelic-infrastructure/templates/kubelet/_agent-config_helper.tpl
```diff
@@ -0,0 +1,26 @@
...
+http_server_port: 8003
+features:
+  docker_enabled: false
+{{- if include "common.privileged" . }}
+is_secure_forward_only: true
+overide_host_root: ""  # Typo from here: https://github.com/newrelic/infrastructure-agent/blob/master/pkg/config/config.go#L267
+{{- end }}
+{{- if .Values.enableProcessMetrics }}
+enable_process_metrics: {{ .Values.enableProcessMetrics }}
+{{- end }}
```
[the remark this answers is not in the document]
kang-makes: You are right. I fixed it and added a test so we do not forget about it. Solved in [#552d23c](https://github.com/newrelic/nri-kubernetes/pull/383/commits/552d23c0fdc6bf9cce606e2d141a7722d4b7c856)

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
kang-makes: Paolo stated this [above](https://github.com/newrelic/nri-kubernetes/pull/383#discussion_r845194876) and this was solved in [#64e304d](https://github.com/newrelic/nri-kubernetes/pull/383/commits/64e304d72084e4070dc15553499680bc4f1ba85a).

charts/newrelic-infrastructure/templates/secret.yaml
kang-makes: I am leaving this for the nit-picking commit.

charts/newrelic-infrastructure/values.yaml
kang-makes: Paolo stated this https://github.com/newrelic/nri-kubernetes/pull/383#discussion_r845194876 and this was solved in [#64e304d](https://github.com/newrelic/nri-kubernetes/pull/383/commits/64e304d72084e4070dc15553499680bc4f1ba85a).

charts/newrelic-infrastructure/templates/kubelet/daemonset.yaml
kang-makes: CAOS has [`tini` in the entrypoint and `/usr/bin/newrelic-infra-service` in the command](https://github.com/newrelic/infrastructure-agent/blob/master/build/container/Dockerfile#L44-L45)

I don't want to override `tini` so CAOS can change PID 1 freely but only change the command from `newrelic-infra-service` to `newrelic-infra`.

```diff
- command: [ "/sbin/tini", "--", "/usr/bin/newrelic-infra" ]
+ args: [ "newrelic-infra" ]
```

Do you see any issues?

charts/newrelic-infrastructure/templates/NOTES.txt
```diff
@@ -20,17 +20,18 @@ You have specified ksm or kubelet integration components as not enabled.
 Those components are needed to have the full experience on NROne kubernetes explorer.
 {{- end }}
 
-{{- if and .Values.controlPlane.enabled (not (include "newrelic.controlPlane.hostNetwork" .)) }}
+{{- if and .Values.controlPlane.enabled (not (include "common.hostNetwork" . | default .Values.controlPlane.hostNetwork)) }}
```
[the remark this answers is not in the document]
kang-makes: Screenshot 2022-04-07 at 18 48 37

Well... This is kind of tricky because `default` reads falsiness so from global and local toggles are going to be overridden by the `values.yaml` where `.Values.controlPlane.hostNetwork` is true.

So virtually is the same as putting directly this:
```suggestion
{{- if and .Values.controlPlane.enabled .Values.controlPlane.hostNetwork }}
```

WDYT?

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
```diff
@@ -18,50 +18,63 @@ spec:
...
-          {{- toYaml . | nindent 8 }}
+        {{- . | nindent 8 }}
       {{- end }}
-      serviceAccountName: {{ template "newrelic.serviceAccountName" . }}-controlplane
-      {{- if include "newrelic.controlPlane.hostNetwork" . }}
+      {{- with include "common.dnsConfig" . }}
+      dnsConfig:
+        {{- . | nindent 8 }}
+      {{- end }}
+      {{- if ( include "common.hostNetwork" . ) | default .Values.controlPlane.hostNetwork }}
```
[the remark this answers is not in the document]
kang-makes: The same I said [up here](https://github.com/newrelic/nri-kubernetes/pull/383#discussion_r845368587):

Screenshot 2022-04-07 at 18 48 37

Virtually is the same as putting directly this:
```suggestion
{{- .Values.controlPlane.hostNetwork }}
```

Because of the way `default` threats falsiness... it is just to keep the precedence.

charts/newrelic-infrastructure/values.yaml
```diff
@@ -381,20 +381,10 @@ podLabels: {}
...
-
 # -- Run the integration with full access to the host filesystem and network.
 # Running in this mode allows reporting fine-grained cpu, memory, process and network metrics for your nodes.
 # Additionally, it allows control plane monitoring, which requires hostNetwork to work.
-privileged: true
+privileged:
```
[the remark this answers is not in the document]
kang-makes: Solved in [#aa2aa46](https://github.com/newrelic/nri-kubernetes/pull/383/commits/aa2aa46ad9b26aa3e9ef1948c515845302918f77)

charts/newrelic-infrastructure/templates/kubelet/daemonset.yaml
```diff
@@ -60,10 +60,9 @@ spec:
               value: {{ get (fromYaml (include "nriKubernetes.kubelet.agentConfig" .)) "http_server_port" | quote }}
             - name: "NRI_KUBERNETES_CLUSTERNAME"
               value: {{ include "common.cluster" . }}
-            {{- if .Values.verboseLog }}
-            - name: NRI_KUBERNETES_VERBOSE
-              value: "1"
-            {{- end }}
+            - name: "NRI_KUBERNETES_VERBOSE"
+              value: {{ include "common.verboseLog.value" . | quote }}
```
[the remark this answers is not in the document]
kang-makes: This is not for the agent but for the integration which is a boolean.

The agent uses a config map for the agent config which [uses a helper that sources the common library](https://github.com/newrelic/nri-kubernetes/pull/383/commits/1c85539e2722fd33671d5fae4becb42fbfbc2b8d#diff-6c80a80d922a33ec1af302d794c90725a9c4540cc41bcf1b8afef46af1a2fda0R21).

The agent uses an integer for this: https://github.com/newrelic/helm-charts/blob/master/library/common-library/templates/_agent-config.tpl#L4-L7

charts/newrelic-infrastructure/templates/NOTES.txt
```diff
@@ -20,17 +20,18 @@ You have specified ksm or kubelet integration components as not enabled.
...
 Warning:
 ========
 
 Most Control Plane components listen in the loopback address only, which is not reachable without `hostNetwork: true`.
 Control plane autodiscovery might not work as expected.
-You can enable hostNetwork for control plane scraper pods only by setting `controlPlane.unprivilegedHostNetwork: true`,
-or alternative disable control plane monitoring altogether with `controlPlane.enabled: false`.
+You can enable hostNetwork for control plane scraper pods only by setting `global.hotNetwork`, `hostNetwork` or
+`controlPlane.hostNetwork: true` (in order of granularity for all the helm charts, this chart or only the control plane metrics
+discovery or alternative disable control plane monitoring altogether with `controlPlane.enabled: false`.
```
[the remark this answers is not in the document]
kang-makes: This seems better for the `privileged` PR to make this less messy.

But I am leaving this thread open as a reminder.

charts/newrelic-infrastructure/templates/kubelet/_security_context_helper.tpl
```diff
@@ -0,0 +1,21 @@
...
+{{- define "nriKubernetes.kubelet.securityContext.agentContainer" -}}
+{{- $privileged := dict -}}
+{{- if include "common.privileged" . -}}
+{{- $privileged = fromYaml ( include "nriKubernetes.kubelet.securityContext.privileged" . ) -}}
+{{- end -}}
+{{- $privileged
+        | mustMergeOverwrite (include "newrelic.compatibility.securityContext" . | fromYaml )
+        | mustMergeOverwrite (include "common.securityContext.container" . | fromYaml )
+        | toYaml
+-}}
```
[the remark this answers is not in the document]
kang-makes: Can we leave this for the privileged PR?

There are too many things that need to change.

charts/newrelic-infrastructure/templates/kubelet/daemonset.yaml
```diff
@@ -16,51 +16,61 @@ spec:
...
+        {{- . | nindent 8 }}
+      {{- end }}
+      {{- with include "common.priorityClassName" . }}
+      priorityClassName: {{ . }}
       {{- end }}
-      serviceAccountName: {{ template "newrelic.serviceAccountName" . }}
+      {{- with include "common.securityContext.pod" . }}
+      securityContext:
+        {{- . | nindent 8 }}
+      {{- end }}
```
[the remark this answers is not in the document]
kang-makes: It should but I am not completely sure. Privileged PR is going to be big too so we can separate it if you would.

charts/newrelic-infrastructure/templates/controlplane/_agent-config_helper.tpl
```diff
@@ -0,0 +1,20 @@
+{{- /*
+Defaults for controlPlane's agent config
+*/ -}}
+{{- define "nriKubernetes.controlPlane.agentConfig.defaults" -}}
```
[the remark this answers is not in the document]
gsanchezgavier: I see, what i'm not convinced is about having the default inside the template, I don't know if there are good practices or standards related but I like more the idea to put defaults on Values or the templates rather than the helplers.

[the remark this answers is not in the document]
kang-makes: I thought it was a good idea and have this in a more declarative way here in a helper instead of having it configured in three different layers (environment variables in Dockerfile, environment variables at container runtime and a couple of lines in the manifest)

The point of this is that these defaults are not "defaults" but are needed for the integration to work. They were hardcoded as environment variables and inside the forwarder image. That is the reason I removed them from being "hidden" between many layers and moved here.

The map for agent config is only in `.Values.common` so there is no possibility to make it go to the values unless we add config per scraper but I would be happy to do that change if that makes things easy to follow, document and reuse possible values.

Let me know what you think :)

charts/newrelic-infrastructure/values.yaml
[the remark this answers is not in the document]
kang-makes: I got affinity and tolerations back to the values and changed tests accordingly.

`tolerations` roll back solved in [#6dc65d5](https://github.com/newrelic/nri-kubernetes/pull/383/commits/6dc65d59737a550ce3b82aca406fb2891dd848b0)
`affinity` roll back solved in [#583772b](https://github.com/newrelic/nri-kubernetes/pull/383/commits/583772be24ee6debf2db48717295fd8b68ec9970)

charts/newrelic-infrastructure/values.yaml
kang-makes: I got affinity and tolerations back to the values and changed tests accordingly.

`tolerations` roll back solved in [#6dc65d5](https://github.com/newrelic/nri-kubernetes/pull/383/commits/6dc65d59737a550ce3b82aca406fb2891dd848b0)
`affinity` roll back solved in [#583772b](https://github.com/newrelic/nri-kubernetes/pull/383/commits/583772be24ee6debf2db48717295fd8b68ec9970)

kang-makes: This one in particular solved in [#cc3aee0](https://github.com/newrelic/nri-kubernetes/pull/383/commits/cc3aee07d53624c7ea805f80fbf4e7dfe2c32aad)

charts/newrelic-infrastructure/values.yaml
kang-makes: I leave them for the last nit-picking commit.

charts/newrelic-infrastructure/templates/controlplane/_tolerations_helper.tpl
kang-makes: This helper does not exist anymore. Resolving.

charts/newrelic-infrastructure/templates/controlplane/_tolerations_helper.tpl
kang-makes: The latest version is so much simpler than the one you reviewed here.

I think no further explanation is needed. WDYT?

charts/newrelic-infrastructure/templates/controlplane/rolebinding.yaml
```diff
@@ -6,16 +6,16 @@ apiVersion: rbac.authorization.k8s.io/v1
...
   namespace: {{ . }}
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: {{ template "newrelic.fullname" $ }}-secrets
+  name: {{ include "common.naming.fullname" $ }}-secrets
 subjects:
 - kind: ServiceAccount
-  name: {{ template "newrelic.serviceAccountName" $ }}-controlplane
+  name: {{ include "common.serviceAccount.name" $ }}-controlplane
```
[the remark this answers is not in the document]
kang-makes: I did not notice that we cannot because there is a `range` a couple of lines above. Solving this.

charts/newrelic-infrastructure/templates/ksm/deployment.yaml
kang-makes: Solved in [#881654c](https://github.com/newrelic/nri-kubernetes/pull/383/commits/881654c7b9e3b6a104b4a9a29652cf7526e4c4f8)

charts/newrelic-infrastructure/templates/secret.yaml
kang-makes: Solved in [#881654c](https://github.com/newrelic/nri-kubernetes/pull/383/commits/881654c7b9e3b6a104b4a9a29652cf7526e4c4f8)

charts/newrelic-infrastructure/values.yaml
kang-makes: Solved in [#881654c](https://github.com/newrelic/nri-kubernetes/pull/383/commits/881654c7b9e3b6a104b4a9a29652cf7526e4c4f8)

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
gsanchezgavier: > How this will work if there is not custom secret ? I saw previous one not rendering this annotation.

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
```diff
@@ -93,61 +102,52 @@ spec:
...
                 fieldRef:
                   apiVersion: "v1"
                   fieldPath: "spec.nodeName"
-            {{- if .Values.verboseLog }}
-            - name: NRIA_VERBOSE
-              value: "1"
-            {{- end }}
+
             {{- if .Values.useNodeNameAsDisplayName }}
             - name: "NRIA_DISPLAY_NAME"
```
gsanchezgavier, line 134: shouldn't this config also moved to the helper?

gsanchezgavier (approved):
LGTM but there are some open important conversations. We probably need more ✅   before merge it

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
kang-makes: Oh, sorry.

Example `values.yaml`:
```yaml
global:
  cluster: test
  customSecretName: test
  customSecretLicenseKey: test
```

`% helm template test . -f values-custom.yaml -s templates/kubelet/daemonset.yaml`
```yaml
apiVersion: apps/v1
kind: DaemonSet
spec:
  template:
    metadata:
      annotations:
        checksum/license-secret: 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b
```
Empty secret, empty file, a hash of nothing. But hashed. Ready to redeploy if this changes.

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
kang-makes: You are right. But I still have to change this from this common library to 1.0 and didn't want to spend more time on this and make this PR bigger.

I leave this comment open as a reminder of doing it in the follow-up PR.

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
gsanchezgavier: could you add a comment somewere? to me it could be confusing to have this in the annotations without having a license-secret.
I would prefer not to render it if the secret not exist. I guess that the redeploy will happen anyway if the license is changed since is an envar IRRC

kang-makes: Fair point. [From #7bd684d](https://github.com/newrelic/nri-kubernetes/pull/383/commits/7bd684da4abd9bd632f5650288ccb4367b31b8b2) it only renders when a secret is rendered.

I am keeping this thread open as @roobre pointed out the security issue. I forgot to add that the hash is for the whole secret so it changes also when the chart updates because it includes the labels. Brute forcing a whole YAML, even if you know that it only changes a couple of things like version or release name is madness from my point of view.

gsanchezgavier: would make senes to add a random(like `{{ randAlphaNum 5 | quote }}`) label/annotation or extra secret to the secret  to make brut force even harder? 

But anyway I think the auto rolling feature is worth to be implemented. Unprivileged users should not have the permissions to see this annotations right?

kang-makes: Even if they can we were using this before this refactor without hesitation. The brute force needed for this would be massive. All the labels are salt enough to keep the secret secret.

charts/newrelic-infrastructure/templates/NOTES.txt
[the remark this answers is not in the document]
kang-makes: Think that this is for the `NOTES.txt`, not for the logic of any manifest.

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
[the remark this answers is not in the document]
kang-makes: You were completely right, I don't know what I was thinking when I wrote that (and it seems that I was completely convinced of that).

Solved on [#a1b6f77](https://github.com/newrelic/nri-kubernetes/pull/383/commits/a1b6f77ff8c7384c6873cd52c42aa9ca2068523c)

charts/newrelic-infrastructure/templates/NOTES.txt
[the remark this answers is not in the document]
kang-makes: You were completely right, I don't know what I was thinking when I wrote that (and it seems that I was completely convinced of that).

Solved on [#a1b6f77](https://github.com/newrelic/nri-kubernetes/pull/383/commits/a1b6f77ff8c7384c6873cd52c42aa9ca2068523c)

charts/newrelic-infrastructure/templates/controlplane/serviceaccount.yaml
```diff
@@ -1,13 +1,13 @@
...
+{{- if include "common.serviceAccount.create" . -}}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  namespace: {{ .Release.Namespace }}
-  name: {{ include "newrelic.serviceAccountName" . }}-controlplane
-  labels:
-    {{- include "newrelic.labels" . | nindent 4 }}
-  {{- with .Values.serviceAccount.annotations }}
+  {{- with (include "common.serviceAccount.annotations" .) }}-controlplane
```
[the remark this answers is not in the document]
kang-makes: Solved in [#55a9745](https://github.com/newrelic/nri-kubernetes/pull/383/commits/55a9745e64086d80b207c6a9366f7008bcda9947)

kang-makes:
I am merging before getting 2 approvals because the PR was getting stale. All the issues that have not been marked as solved are still open. Feel free to review them and open issues in the repo to discuss them separately.

davidgit:
Thanks @kang-makes, good job!

davidgit:
@roobre feel free to open further issues

PR #384, opened 2022-03-23
readme updates

sigilioso:


e2e/README.md
```diff
@@ -17,12 +17,18 @@ and `TARGETOS` args to be populated.
...
 export  DOCKER_BUILDKIT=1
-docker build -t e2e/newrelic-infrastructure:e2e -f Dockerfile-infra .
 docker build -t e2e/nri-kubernetes:e2e  .
-minikube image load e2e/newrelic-infrastructure:e2e
 minikube image load e2e/nri-kubernetes:e2e
 ```
 
+Then, include helm needed repositories.
+```shell
+helm repo add stable https://charts.helm.sh/stable
```
roobre, line 26: Is the stable repo needed? If it is that should be something we want to get rid of, the stable repo is not being maintained anymore.

sigilioso: Actually, I think it is not needed... I got the hint [from here](https://github.com/newrelic/nri-kubernetes/blob/main/.github/workflows/e2e.yaml#L59), so maybe we should take a look at that too.

roobre: I don't think it is either, I render the charts quite often and I don't have it added. I think perhaps an older iteration of the e2e pipeline had it and we kept it accidentally when we reworked it. Perhaps @paologallinaharbur knows?

paologallinaharbur: Maybe it was initially needed for the KSM, now as far as I remember we do not use anything from that..

sigilioso: Removed here: [a8f62c3](https://github.com/newrelic/nri-kubernetes/pull/384/commits/a8f62c3fc039d686170fbf39a41d26c08f67b6e5)

roobre (approved):
Thanks for taking care of the Readme! ❤️

PR #393, opened 2022-04-06
refactor: ksm round tripper

gsanchezgavier:


src/ksm/client/discovery.go
```diff
@@ -118,12 +117,20 @@ func (sd *discoverer) Discover(timeout time.Duration) (client.HTTPClient, error)
...
+	}
+
+	return KSMClient, nil
 }
 
-func newKSMClient(timeout time.Duration, nodeIP string, endpoint url.URL, logger *logrus.Logger, k8s client.Kubernetes) *ksm {
-	bearer := k8s.Config().BearerToken
-	rt := newBearerRoundTripper(bearer)
+func newKSMClient(timeout time.Duration, nodeIP string, endpoint url.URL, logger *logrus.Logger) (*ksm, error) {
+	rt, err := transport.New(&transport.Config{TLS: transport.TLSConfig{Insecure: true}})
```
roobre, line 130: I'm getting a bit lost with the imports here. To which object or package does `transport.` refer to here?

gsanchezgavier: `"k8s.io/client-go/transport"`

src/ksm/client/discovery.go
```diff
@@ -118,12 +117,20 @@ func (sd *discoverer) Discover(timeout time.Duration) (client.HTTPClient, error)
 	}
 
 	sd.logger.Debugf("KSM client created with endpoint=%v and nodeIP=%v", endpoint, nodeIP)
-	return newKSMClient(timeout, nodeIP, endpoint, sd.logger, sd.k8sClient), nil
+
+	KSMClient, err := newKSMClient(timeout, nodeIP, endpoint, sd.logger)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create KSM http client: %w", err)
```
paologallinaharbur, line 123: [NIT] I prefer to wrap errors with different sentences to understand where they are coming from. In this case we have the same sentence in 3 funcitons

paologallinaharbur: (I am not even sure it is a best practice to do so)

src/ksm/client/discovery_test.go
```diff
@@ -450,32 +421,44 @@ func TestNodeIP(t *testing.T) {
...
-	// Accept Header was added to the request
-	assert.Equal(t, prometheus.AcceptHeader, resp.Request.Header.Get("Accept"))
-	// Correct http method was used
-	assert.Equal(t, "GET", resp.Request.Method)
+			// When retrieving http response
+			resp, err := c.Do("GET", "foo")
+
+			// The call works correctly
+			assert.NoError(t, err)
+			// The request was created with updated path for URL
```
paologallinaharbur, line 454: should we test that no token is there or it would be reduntant?

paologallinaharbur (approved):
thanks for fixing this!

src/ksm/client/discovery.go
roobre: Got it, from the docs I can also see that it should not include any extra headers in the request unless explicitly told to (which we aren't). LGTM.

roobre (approved):

PR #402, opened 2022-04-25
Upgrade common library

kang-makes:
We knew that the common library was flawed but I didn't want to make the PR even bigger than it was.

So the upgrade of the common library with all the breaking changes and the missing fine-grain issues of the previous PR (#383) should be solved in this PR.

There are a few issues that are still missing to be solved remaining in the helm-charts repository but this finishes the refactor of this chart.

This time I went (again) commit by commit changing things one by one to make the read easier as there are functions that are going to be removed and added and I suspected that the `git diff` was going to become hard to read... And I was right.


charts/newrelic-infrastructure/templates/kubelet/_security_context_helper.tpl
```diff
@@ -8,14 +9,24 @@ readOnlyRootFilesystem: true
...
+{{- $commonLibrary := include "newrelic.common.securityContext.container" . | fromYaml -}}
+
+{{- $finalSecurityContext := dict -}}
+{{- if $commonLibrary -}}
+    {{- $finalSecurityContext = mustMergeOverwrite $commonLibrary $compatibilityLayer -}}
+{{- else -}}
+    {{- $finalSecurityContext = mustMergeOverwrite $defaults $compatibilityLayer -}}
+{{- end -}}
+
+{{- toYaml $finalSecurityContext -}}
```
sigilioso, line 31: The securityContext helpers make the chart way more cleaner, 🧑‍🎨 

I'm not sure if it would worth, it but maybe we could include an additional helper to avoid repeating the common part between `nriKubernetes.kubelet.securityContext.agentContainer`, and `nriKubernetes.securityContext.container`. Something like this:

```
{{-  define "nriKubernetes.securityContext.mergeCommonAndCompatibility" -}}
{{- $compatibilityLayer := include "newrelic.compatibility.securityContext" .context | fromYaml -}}
{{- $commonLibrary := include "newrelic.common.securityContext.container" .context | fromYaml -}}
{{- $finalSecurityContext := dict -}}
{{- if $commonLibrary -}}
    {{- $finalSecurityContext = mustMergeOverwrite $commonLibrary $compatibilityLayer -}}
{{- else -}}
    {{- $finalSecurityContext = mustMergeOverwrite .defaults $compatibilityLayer -}}
{{- end -}}
{{ - end - }}
```
WDYT?

sigilioso (approved):
👍  Thanks for addressing it, great work! This is almost done :)

I've left a _nit_ and I leave another here: `_common_library_overrides.tpl` filename does still make sense? As overriding philosophy changed, maybe those helpers could be moved to `_helpers.tpl` WDYT?

charts/newrelic-infrastructure/templates/kubelet/daemonset.yaml
```diff
@@ -16,41 +17,42 @@ spec:
...
+      {{- with include "newrelic.common.dnsConfig" . }}
       dnsConfig:
         {{- . | nindent 8 }}
       {{- end }}
-      {{- with include "common.priorityClassName" . }}
+      {{- with include "newrelic.common.priorityClassName" . }}
       priorityClassName: {{ . }}
       {{- end }}
-      {{- with include "common.securityContext.pod" . }}
+      {{- with include "newrelic.common.securityContext.pod" . }}
```
kang-makes, line 50: This had an open issue in the previous PR that may need to be taken into consideration again: https://github.com/newrelic/nri-kubernetes/pull/383#discussion_r844178708

kang-makes:
>  leave another here: _common_library_overrides.tpl filename does still make sense? As overriding philosophy changed, maybe those helpers could be moved to _helpers.tpl WDYT?

You are right. With the [refactor where the helpers are placed (#388ccda)](https://github.com/newrelic/nri-kubernetes/pull/402/commits/388ccdaf8c5610a290f6e40b562a1f3843d3315b) commit I moved helpers around. Now the RBAC rules that only apply to the control plane are in a helper file in that folder instead of in the `_helpers.tpl` at root level.

Let me know your thoughts on the last commit @sigilioso.

charts/newrelic-infrastructure/templates/kubelet/_security_context_helper.tpl
kang-makes: The thing is that reading those helpers is right now enough hard to add more complexity of going back and forth jumping between helpers.

The logic of the security context is the same for all the containers **except** for the agent in the kubelet that the default behavior changes depending on the `privileged` toggle. That is the reason I prefer to have duplicated code. If in the future I have to change the behavior I would only have to change for that container which is the odd one.

If I convinced you you can resolve this thread but if not just leave it open to know if somebody has concerns about this too :)

sigilioso:
> > leave another here: _common_library_overrides.tpl filename does still make sense? As overriding philosophy changed, maybe those helpers could be moved to _helpers.tpl WDYT?
> 
> You are right. With the [refactor where the helpers are placed (#388ccda)](https://github.com/newrelic/nri-kubernetes/pull/402/commits/388ccdaf8c5610a290f6e40b562a1f3843d3315b) commit I moved helpers around. Now the RBAC rules that only apply to the control plane are in a helper file in that folder instead of in the `_helpers.tpl` at root level.
> 
> Let me know your thoughts on the last commit @sigilioso.

🧹 Now helpers are tidy and have a better naming, awesome!

Thanks for taking care of it.

sigilioso (approved):
👍 LGTM

charts/newrelic-infrastructure/templates/NOTES.txt
```diff
@@ -26,9 +26,9 @@ Warning:
 
 Most Control Plane components listen in the loopback address only, which is not reachable without `hostNetwork: true`.
 Control plane autodiscovery might not work as expected.
-You can enable hostNetwork for control plane scraper pods only by setting `global.hotNetwork`, `hostNetwork` or
-`controlPlane.hostNetwork: true` (in order of granularity for all the helm charts, this chart or only the control plane metrics
-discovery or alternative disable control plane monitoring altogether with `controlPlane.enabled: false`.
+You can enable hostNetwork for control plane scraper pods by setting `global.hotNetwork`, `hostNetwork` or only for the control
```
roobre, line 29:
```suggestion
You can enable hostNetwork for all pods by setting `global.hotNetwork`, `hostNetwork`, or only for the control
```

charts/newrelic-infrastructure/templates/_helpers.tpl
```diff
@@ -1,3 +1,32 @@
+{{/*
+This is a copy and paste from the common-library's name helper.
```
roobre, line 2: Why?

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
```diff
@@ -4,8 +4,9 @@ kind: {{ .Values.controlPlane.kind }}
 metadata:
   namespace: {{ .Release.Namespace }}
   labels:
-    {{- include "common.labels" . | nindent 4 }}
-  name: {{ include "common.naming.fullname" . }}-controlplane
+    {{- include "newrelic.common.labels" . | nindent 4 }}
+    mode: {{ if not ( include "newrelic.common.privileged" . ) }}un{{ end }}privileged
```
roobre, line 8: Should we make a teeny tiny helper for this? As we use it in several places and so.

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
```diff
@@ -18,63 +19,63 @@ spec:
...
-      {{- include "common.labels.selectorLabels" . | nindent 6 }}
+      {{- include "newrelic.common.labels.selectorLabels" . | nindent 6 }}
       app.kubernetes.io/component: controlplane
   template:
     metadata:
       annotations:
         checksum/nri-kubernetes: {{ include (print $.Template.BasePath "/controlplane/scraper-configmap.yaml") . | sha256sum }}
         checksum/agent-config: {{ include (print $.Template.BasePath "/controlplane/agent-configmap.yaml") . | sha256sum }}
-        {{- if include "common.license.secret" . }}{{- /* If the is secret to template */}}
+        {{- if include "newrelic.common.license.secret" . }}{{- /* If the is secret to template */}}
```
roobre, line 29:
```suggestion
        {{- if include "newrelic.common.license.secret" . }}
        {{- /* Add checksum only if we are not using an external secret */}}
```

roobre: I'm still not happy about checksumming a secret though 🤔

charts/newrelic-infrastructure/templates/ksm/_naming.tpl
```diff
@@ -0,0 +1,8 @@
+{{- /* Naming helpers*/ -}}
+{{- define "nriKubernetes.ksm.fullname" -}}
+{{- include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" (include "nriKubernetes.naming.fullname" .) "suffix" "ksm") -}}
+{{- end -}}
+
+{{- define "nriKubernetes.ksm.fullname.agent" -}}
+{{- include "newrelic.common.naming.truncateToDNSWithSuffix" (dict "name" (include "nriKubernetes.naming.fullname" .) "suffix" "agent-ksm") -}}
+{{- end -}}
```
roobre, line 8: I like to see these here, it makes it way more readable than if we had the `truncateToDNSWithSuffix` directly on the yamls!

roobre (approved):
Mostly nits here, looking great!

charts/newrelic-infrastructure/templates/NOTES.txt
kang-makes: Solved in [#78d6879](https://github.com/newrelic/nri-kubernetes/pull/402/commits/78d6879bf6523dee1de549e27bb17a7d7bb1fdd5)

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
kang-makes: Better [now](https://github.com/newrelic/nri-kubernetes/pull/402/commits/78d6879bf6523dee1de549e27bb17a7d7bb1fdd5)?

charts/newrelic-infrastructure/templates/_helpers.tpl
kang-makes: Better [now](https://github.com/newrelic/nri-kubernetes/pull/402/commits/78d6879bf6523dee1de549e27bb17a7d7bb1fdd5)?

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
kang-makes: Solved on [#67f2496](https://github.com/newrelic/nri-kubernetes/pull/402/commits/67f2496a0bacfc8edb07e7200a6325fb4fc9b3ef)

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
kang-makes: The secret has also the labels: https://github.com/newrelic/helm-charts/blob/master/library/common-library/templates/_license_secret.yaml.tpl#L17

So it is salted. So I see these issues:
 * If we use a random string that uses a seed, we still have the same issue that it can be brute-forced.
 * If we put a pure random string, any change on the helm chart will redeploy everything.
 
 I see no solution that could make a horrible UX.

roobre: Me neither, the only solution I see is just not force a redeploy if LK changes (which on the other hand I don't see as a very bad thing).

charts/newrelic-infrastructure/templates/kubelet/daemonset.yaml
kang-makes: It was before but the overriding system left it a bit hidden.

PR #403, opened 2022-04-26
Update dependencies to solve security issues pointed by trivy

kang-makes:
https://github.com/newrelic/nri-kubernetes/runs/6171027124?check_suite_focus=true


go.mod
```diff
@@ -1,65 +1,74 @@
...
+	k8s.io/api v0.23.6
+	k8s.io/apimachinery v0.23.6
+	k8s.io/client-go v0.23.6
+	k8s.io/kubelet v0.23.6
+)
+
+replace (
+	// To avoid CVE-2018-1098 triggering a security scan
+	go.etcd.io/etcd v3.3.27+incompatible => go.etcd.io/etcd/v3 v3.5.4
+	// This is a dependency of golint, trivy is mixing up dependencies
```
sigilioso, line 26: I think it is a really-nested dependency from `k8s.io/apimachinery` too... maybe trivy is not mixing that up

sigilioso (approved):
LGTM, awesome you took the chance to upgrade go version and some other dependencies too, thanks!

I've just left a message to update the trivy-related comment if you consider it is right

go.mod
kang-makes: To be honest, I still not understand the whole go toolset so I would leave it as this until trivy raise another issue and I have more experience.

Issue #244, opened 2021-10-21
`K8sPodSample.isReady` is reported as `NULL` for pending pods


roobre:
For pending (not scheduled) pods, we retrieve pod metrics from KSM rather than the Kubelet: https://github.com/newrelic/nri-kubernetes/blob/0a78aa518374cc8d250122d26268ba5e1858a754/src/metric/definition.go#L690-L690
This causes that if KSM does not have a `kube_pod_status_ready` metric for the pod, then that particular metric will not be set and appear as `NULL` in the NR platform.

For pending pods, `isReady` should be reported explicitly as `false`.

PR #404, opened 2022-04-27
fix(pendingPods): some metrics are missing from KSM

paologallinaharbur:
Fix https://github.com/newrelic/nri-kubernetes/issues/244

Notice that the pod metrics are fetched **only** for pending pods thanks to `IDGenerator: prometheus.FromLabelsValueEntityIDGeneratorForPendingPods(),`

A different approach would be to completely remove the fetch function and to hardcode that to `false`, let me know if you have any preference.  

I removed along the way also startTime since it was meaningless for pending pods

### DIFF
--- a/src/metric/definition.go
+++ b/src/metric/definition.go
@@ -689,7 +689,6 @@ var KSMSpecs = definition.SpecGroups{
 		TypeGenerator: prometheus.FromLabelValueEntityTypeGenerator("kube_pod_status_phase"),
 		Specs: []definition.Spec{
 			{Name: "createdAt", ValueFunc: prometheus.FromValue("kube_pod_created"), Type: sdkMetric.GAUGE, Optional: true},
-			{Name: "startTime", ValueFunc: prometheus.FromValue("kube_pod_start_time"), Type: sdkMetric.GAUGE},
 			{Name: "createdKind", ValueFunc: prometheus.FromLabelValue("kube_pod_info", "created_by_kind"), Type: sdkMetric.ATTRIBUTE, Optional: true},
 			{Name: "createdBy", ValueFunc: prometheus.FromLabelValue("kube_pod_info", "created_by_name"), Type: sdkMetric.ATTRIBUTE, Optional: true},
 			{Name: "nodeIP", ValueFunc: prometheus.FromLabelValue("kube_pod_info", "host_ip"), Type: sdkMetric.ATTRIBUTE},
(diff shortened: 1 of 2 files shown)


src/metric/definition.go
```diff
@@ -689,15 +689,15 @@ var KSMSpecs = definition.SpecGroups{
...
 			{Name: "createdBy", ValueFunc: prometheus.FromLabelValue("kube_pod_info", "created_by_name"), Type: sdkMetric.ATTRIBUTE, Optional: true},
 			{Name: "nodeIP", ValueFunc: prometheus.FromLabelValue("kube_pod_info", "host_ip"), Type: sdkMetric.ATTRIBUTE},
 			{Name: "namespace", ValueFunc: prometheus.FromLabelValue("kube_pod_info", "namespace"), Type: sdkMetric.ATTRIBUTE},
 			{Name: "namespaceName", ValueFunc: prometheus.FromLabelValue("kube_pod_info", "namespace"), Type: sdkMetric.ATTRIBUTE},
 			{Name: "nodeName", ValueFunc: prometheus.FromLabelValue("kube_pod_info", "node"), Type: sdkMetric.ATTRIBUTE},
 			{Name: "podName", ValueFunc: prometheus.FromLabelValue("kube_pod_info", "pod"), Type: sdkMetric.ATTRIBUTE},
-			{Name: "isReady", ValueFunc: definition.Transform(prometheus.FromLabelValue("kube_pod_status_ready", "condition"), toNumericBoolean), Type: sdkMetric.GAUGE},
+			// we are adding as default `false` since all ksm pods used refers to pending pods due to the IDGenerator.
+			{Name: "isReady", ValueFunc: definition.Transform(fetchWithDefault(prometheus.FromLabelValue("kube_pod_status_ready", "condition"), "false"), toNumericBoolean), Type: sdkMetric.GAUGE},
```
paologallinaharbur, line 700: This could be replace by something like:
```go
{Name: "isReady", ValueFunc: func(_, _ string, _ definition.RawGroups) (definition.FetchedValue, error) { return 0, nil }, Type: sdkMetric.GAUGE},
```

I opted for the actual one just to follow the same approach, but I am doubting among the two options

roobre (approved):
LGTM, happy to get rid of a missing metric :D

I think though that we should mention this in the release notes, just to make visible that queries mentioning `WHERE isReady IS NULL` will no longer be a thing.

src/metric/definition.go
sigilioso: I think I'm missing something. The alternative you proposed would not do the same, wold it? It always returns regardless of the `"condition"` label value which is taken into account in the solution you chose... Sorry for may possible lack of context here.

sigilioso: I think I'm missing something. The alternative you proposed would not do the same, wold it? It always returns regardless of the `"condition"` label value which is taken into account in the solution you chose... Sorry for may possible lack of context here.

paologallinaharbur: The code is quite confusing.
The metric should be never present since we are fetching pod data from KSM merely if the pod is `Pending`. You can check the `IDGenerator` few lines above

sigilioso: Now I get the reason why both are equivalent, thanks for the explanation!

PR #408, opened 2022-05-02
fix(ci/cd): extra char in pipline command

paologallinaharbur:
A small bug was introduced in the ci/cd pipeline, it affect for sure also [nri-metadata-injection](https://github.com/newrelic/k8s-metadata-injection/blob/805f1c8993ee497dc57fb6621b34f594fdf218cf/.github/workflows/release-integration.yml#L76) and [metrics-adapter](https://github.com/newrelic/newrelic-k8s-metrics-adapter/blob/24fd202c0b0699ebf2bc326e0df20876f3f24885/.github/workflows/release-integration.yml#L78)

### DIFF
--- a/.github/workflows/release-integration.yml
+++ b/.github/workflows/release-integration.yml
@@ -84,7 +84,7 @@ jobs:
       - name: Push pre-release images
         if: ${{ github.event.release.prerelease }}
         run: |
-          DOCKER_IMAGE_TAG=${DOCKER_IMAGE_TAG}-pre"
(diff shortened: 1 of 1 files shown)


paologallinaharbur:
On the other hand infra-operator, kube-events, nri-prometheus are fine

kang-makes (approved):
I don't know why there is no :face_palm: emoji in Github.

I saw emails every day but I thought that were only in the `forks` repo and I didn't find time to fix them.

Thank you :)

PR #411, opened 2022-05-04
v2: upgrade base image to newrelic/infrastructure-bundle:2.8.10

roobre:


paologallinaharbur (approved):
Would make sense to add in this PR the changelog change as well?

roobre:
Just did it @paologallinaharbur :)

PR #412, opened 2022-05-04
fix: round up CPU allocatable and capacity metrics

### LINKED ISSUE #75: Fix rounding issue for allocatableCPUCores
## Description

It appears we are rounding the `allocatableCPUCores` attribute and we should not be.  I've attached the results in Query Builder and the raw JSON returned by NRDB.  It's probably worth sanity checking related attributes to ensure we're not rounding others as well.

image

image

## Expected Behavior

I would expect the `allocatedCPUCores` attirbute to reflect what is returned by the `kubectl` output.

## Your Environment
EKS v1.17.12-eks-7684af
New Relic Infra: `newrelic/infrastructure-k8s:2.2.0`

gsanchezgavier:
Fixes #75


roobre (approved):
Looking awesome, thanks for fixing this!

PR #414, opened 2022-05-09
Release charts also from v2 branch

kang-makes:
This allows emptying `helm-charts` repository from legacy charts.

The helm chart for `v2` is waiting in the PR #415

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -4,7 +4,8 @@ on:
   pull_request:
   push:
     branches:
-      - "main"
+      - main
+      - v2
 
 jobs:
   e2eTests:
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e.yaml
```diff
@@ -4,7 +4,8 @@ on:
   pull_request:
   push:
     branches:
-      - "main"
+      - main
+      - v2
```
paologallinaharbur, line 8: Does it worth it to move the chart, update the bundle and then deprecate again the v2 chart?

Or it creates confusion to have it there?

kang-makes: There are links in sites out of our control (like forums) and people are trying to look to the values file of the v3 in the `helm-charts` repository.

This effort is to clean the `helm-charts` repository as much as we can while making the release process and the maintenance burden the least painful.

paologallinaharbur (approved):

PR #422, opened 2022-05-11
[chart] Passthrough environment variables to integrations

kang-makes:
With this change, users should be able to add their own custom environment variables and files so they can use them in their custom configurations.

### DIFF
--- a/charts/newrelic-infrastructure/templates/kubelet/_agent-config_helper.tpl
+++ b/charts/newrelic-infrastructure/templates/kubelet/_agent-config_helper.tpl
@@ -6,10 +6,6 @@ http_server_enabled: true
 http_server_port: 8003
 features:
   docker_enabled: false
-{{- with .Values.kubelet.passthroughEnvironment }}
-passthrough_environment:
-  {{- toYaml . | nindent 2 }}
-{{- end }}
 {{- if not ( include "newrelic.common.privileged" . ) }}
 is_secure_forward_only: true
 overide_host_root: ""  # Typo from here: https://github.com/newrelic/infrastructure-agent/blob/master/pkg/config/config.go#L267
@@ -29,7 +25,8 @@ enable_process_metrics: {{ .Values.enableProcessMetrics }}
 {{- $agentDefaults := fromYaml ( include "newrelic.common.agentConfig.defaults" . ) -}}
 {{- $kubelet := fromYaml ( include "nriKubernetes.kubelet.agentConfig.defaults" . ) -}}
 {{- $agentConfig := fromYaml ( include "newrelic.compatibility.agentConfig" . ) -}}
+{{- $kubeletAgentConfig := .Values.kubelet.agentConfig -}}
 {{- $customAttributes := dict "custom_attributes" (dict "clusterName" (include "newrelic.common.cluster" . )) -}}
 
-{{- mustMergeOverwrite $agentDefaults $kubelet $agentConfig $customAttributes | toYaml -}}
+{{- mustMergeOverwrite $agentDefaults $kubelet $agentConfig $kubeletAgentConfig $customAttributes | toYaml -}}
 {{- end -}}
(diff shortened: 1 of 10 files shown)


charts/newrelic-infrastructure/templates/kubelet/_agent-config_helper.tpl
```diff
@@ -6,6 +6,10 @@ http_server_enabled: true
 http_server_port: 8003
 features:
   docker_enabled: false
+{{- with .Values.kubelet.passthroughEnvironment }}
+passthrough_environment:
+  {{- toYaml . | nindent 2 }}
+{{- end }}
```
roobre, line 12: What is the downside of specifying `passthrough_environment` in `agentConfig` directly?

roobre: I get the side effect but, does it matter? Agents in CP/KSM scrapers run in forward mode and therefore cannot run integrations.

kang-makes: I did not think about agents running in `forward_only` mode.

Anyway, I have done it in a generic way and added documentation for that.

charts/newrelic-infrastructure/values.yaml
```diff
@@ -75,9 +75,23 @@ kubelet:
       effect: "NoExecute"
   nodeSelector: {}
   affinity: {}
+  # -- Config for the Infrastructure agent.
+  # Will be used by the forwarder sidecars and the agent running integrations.
```
roobre, line 79: This one will not be used by the agent running integrations, only by the control plane sidecar.

charts/newrelic-infrastructure/values.yaml
```diff
@@ -120,6 +134,10 @@ ksm:
               matchLabels:
                 app.kubernetes.io/name: kube-state-metrics
           weight: 100
+  # -- Config for the Infrastructure agent.
+  # Will be used by the forwarder sidecars and the agent running integrations.
+  # See: https://docs.newrelic.com/docs/infrastructure/install-infrastructure-agent/configuration/infrastructure-agent-configuration-settings/
+  agentConfig: {}
```
roobre, line 140: Ditto, only for the KSM sidecar

charts/newrelic-infrastructure/values.yaml
```diff
@@ -188,6 +207,10 @@ controlPlane:
   # -- Run Control Plane scraper with `hostNetwork`.
   # `hostNetwork` is required for most control plane configurations, as they only accept connections from localhost.
   hostNetwork: true
+  # -- Config for the Infrastructure agent.
+  # Will be used by the forwarder sidecars and the agent running integrations.
```
roobre, line 211: This one is correct but I think message should be more specific so users can tell the difference between this one and the `common` one.

charts/newrelic-infrastructure/values.yaml
kang-makes: Solved in [#34a64bf](https://github.com/newrelic/nri-kubernetes/pull/422/commits/34a64bfdba0341b3f159726705bcf0e54541c9f4)

charts/newrelic-infrastructure/values.yaml
kang-makes: Solved in [#34a64bf](https://github.com/newrelic/nri-kubernetes/pull/422/commits/34a64bfdba0341b3f159726705bcf0e54541c9f4)

charts/newrelic-infrastructure/values.yaml
kang-makes: I changed the wording in the comit [#34a64bf](https://github.com/newrelic/nri-kubernetes/pull/422/commits/34a64bfdba0341b3f159726705bcf0e54541c9f4)

LMKWYT

roobre (approved):

PR #423, opened 2022-05-12
Fix v2 pipeline

kang-makes:
The chart was released but the pipeline seems to ignore the workflows that are on the `main` branch.

This should add needed workflows to this branch so we can release charts from here.

I am also reverting the changes I made to be able to release charts that were merged into `main`. Take a look at #424.


.github/workflows/push_pr.yaml
```diff
@@ -2,7 +2,6 @@ name: static-analysis and test code
 on:
   push:
     branches:
-      - main
       - v2
   pull_request:
```
roobre, line 6: I think we might want to restrict this to PRs made to the `v2` branch as well.

roobre (approved):

PR #425, opened 2022-05-12
[chart] Add pixie self monitoring

kang-makes:
The functionality to self-monitor pixie was added while `v2` was in beta on the `v2` chart and we did not port that functionality to the `v3` chart.

This should add that lost feature again.


paologallinaharbur (approved):

kang-makes:
Adding `do-not-merge` label until @bpschmitt finds time to test this :)

bpschmitt (approved):
Tested this out and it worked as expected.  Looks good!

PR #427, opened 2022-05-13
[chart][v2] Upgrade image in v2 chart

kang-makes:
We released a new version of the v2 in the PR #426 but did not upgrade the chart.


kang-makes:
I tried to merge it to `main` instead of `v2` and the CLA hook went nuts.

PR #428, opened 2022-05-13
[chart][v2] Upgrade image in v2 chart

kang-makes:
We released a new version of the v2 in PR #426 but did not upgrade the chart.


roobre (approved):
I'd rather have the bump in sync with the appVersion bump, but it's just a nit.

PR #457, opened 2022-05-24
Create the config and namespace filterer

marcsanmi:
Closes https://github.com/newrelic/newrelic-coreint/issues/324.

Filtering namespaces basically works this way:
- Everything is scraped by default.
- We can include namespaces by specific labels using `namespaceSelector.matchLabels`.
- We can include or exclude namespaces using more complex expressions using `namespaceSelector.matchExpressions` (You can check as well https://pkg.go.dev/k8s.io/apimachinery/pkg/labels#Parse)

For more in depth details about the final solution, take a look at this internal [doc](https://newrelic.atlassian.net/wiki/spaces/INST/pages/2338062534/FY23Q1+-+K8s+-+Data+governance+-+Add+remove+components+to+be+instrumented#Solution).


internal/discovery/namespace_filter.go
```diff
@@ -0,0 +1,94 @@
...
+	if nf.c.NamespaceSelector == nil {
+		return true
+	}
+
+	// Scrape namespaces by honoring the matchLabels values.
+	if nf.c.NamespaceSelector.MatchLabels != nil {
+		namespaceList, err := nf.lister.List(labels.SelectorFromSet(nf.c.NamespaceSelector.MatchLabels))
+		if err != nil {
+			log.Errorf("listing namespaces with MatchLabels: %v", err)
+			return false
```
sigilioso, line 54: I wonder if the value returned when there is an error with the underlying namespace lister should be `false` (nothing is Allowed) or `true`(everything is allowed)

internal/discovery/namespace_filter.go
```diff
@@ -0,0 +1,94 @@
...
+	factory.WaitForCacheSync(stopCh)
+
+	return &NamespaceFilter{
+		c:      c,
+		lister: lister,
+	}, stopCh
+}
+
+// IsAllowed checks given any namespace, if it's allowed to be scraped by using the NamespaceLister
+func (nf *NamespaceFilter) IsAllowed(namespace string) bool {
```
sigilioso, line 43: As discussed, I wonder if it would be possible to not only used the cached list of namespaces but also the `IsAllowed` results for each namespace. However, it probably doesn't worth it to look into it further because it may be quite complex and just a premature optimization.

sigilioso (commented):
Great job so far! I left some comments, mostly nits.

internal/discovery/namespace_filter.go
```diff
@@ -0,0 +1,94 @@
...
+	IsAllowed(namespace string) bool
+}
+
+// NamespaceFilter is a struct holding pointers to the config and the namespace lister.
+type NamespaceFilter struct {
+	c      *config.Config
+	lister listersv1.NamespaceLister
+}
+
+// NewNamespaceFilter inits the namespace lister and returns a new NamespaceFilter.
```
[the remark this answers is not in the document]
marcsanmi: Done in 37ffb092acb2ec708e25b3116f06a70006e29ef0.

internal/discovery/namespace_filter_test.go
```diff
@@ -0,0 +1,200 @@
...
+		metav1.CreateOptions{},
+	)
+	require.NoError(t, err)
+
+	// Give some room to the informer to sync, and check that the new namespace is filtered properly.
+	if err := wait.PollImmediateUntilWithContext(ctx, 1*time.Second, func(context.Context) (bool, error) {
+		return nsFilter.IsAllowed(anotherNamespaceName), nil
+	}); err != nil {
+		t.Fatalf("Timed out waiting for the informer to sync: %v", err)
+	}
```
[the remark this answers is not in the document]
marcsanmi: Updated in 37ffb092acb2ec708e25b3116f06a70006e29ef0.

internal/discovery/namespace_filter.go
marcsanmi: Yes, the thing is that adding this cache maintained by us would add extra complexity and taking into account the amount of namespaces a big cluster can have (some hundreds?) the clients would barely notice the difference in performance; maybe we'd be over engineering a bit.

Anyway, a good point to bring up. Let's see if anyone has something to add about this.

marcsanmi: Added the cache layer here 3f558b7cb9c5a01a8b5fff641506b2131c8fab68

internal/discovery/namespace_filter.go
```diff
@@ -0,0 +1,126 @@
...
+	// By default, we scrape every namespace.
+	if nf.c.NamespaceSelector == nil {
+		return true
+	}
+
+	// Check if the namespace is already in the cache.
+	var allowed bool
+	if _, err := nf.storer.Get(namespace, &allowed); err == nil {
+		return allowed
+	}
```
sigilioso, line 60: Using the same `storer` to cache, a different approach which would be availing ourselves of the `NamespaceFilter` interface definition. I'm thinking of something like this:

```go
type CachedNamespaceFilter struct {
    nsFilter NamespaceFilter
    cache storer.Storer
}


func (nf *CachedNamespaceFilter) IsAvailable(namespace string) bool {
// checks cache and calls the underlying NamespaceFilter if the result is not found
}
```

It would be possible to keep tests as they where and test the whole cache thing separately.

WDYT?

internal/discovery/namespace_filter.go
```diff
@@ -0,0 +1,126 @@
...
+// NamespaceFilter is a struct holding pointers to the config and the namespace lister.
+type NamespaceFilter struct {
+	c      *config.Config
+	lister listersv1.NamespaceLister
+	stopCh chan<- struct{}
+	storer storer.Storer
+}
+
+// NewNamespaceFilter inits the namespace lister and returns a new NamespaceFilter.
+func NewNamespaceFilter(c *config.Config, client kubernetes.Interface, storer storer.Storer, options ...informers.SharedInformerOption) *NamespaceFilter {
```
sigilioso, line 31: Just to check if I'm understanding it correctly: in worst case scenario, something is changed and the integration won't notice until cache's ttl elapses, right?

internal/discovery/namespace_filter.go
```diff
@@ -0,0 +1,126 @@
...
+type NamespaceFilterer interface {
+	IsAllowed(namespace string) bool
+}
+
+// NamespaceFilter is a struct holding pointers to the config and the namespace lister.
+type NamespaceFilter struct {
+	c      *config.Config
+	lister listersv1.NamespaceLister
+	stopCh chan<- struct{}
+	storer storer.Storer
```
gsanchezgavier, line 27: not sure if this is the storer interface that this filter needs. It looks a bit odd that returns a value while this needs a bool i think

gsanchezgavier (commented):
looks really good, i lefts some questions and comments

internal/config/config.go
```diff
@@ -218,38 +224,83 @@ type MTLS struct {
...
+		case string:
+			str = v
+		case bool:
+			str = strconv.FormatBool(v)
+		case int:
+			str = strconv.FormatInt(int64(v), 10)
+		case float32, float64:
+			str = fmt.Sprintf("%f", v)
+		default:
+			log.Errorf("parsing expression value: %v, type %v", val, v)
```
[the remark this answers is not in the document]
marcsanmi: I didn't find a clean way to pass the logger into the config. Another approach I thought is that we shouldn't be logging from the config package and just remove the log 😅

internal/discovery/namespace_filter_test.go
```diff
@@ -0,0 +1,351 @@
...
+	t.Parallel()
+
+	type testData struct {
+		namespaceLabels   labels.Set
+		namespaceSelector config.NamespaceSelector
+		expected          bool
+	}
+
+	testCases := map[string]testData{
+		"namespace_allowed_by_default": {
```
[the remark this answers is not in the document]
marcsanmi: Added here ccdd7e20e710e890723333e9b9f3a8f5bf153429.

internal/discovery/namespace_filter.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+// NamespaceFilterer provides an interface to filter namespaces.
+type NamespaceFilterer interface {
+	IsAllowed(namespace string) bool
+}
+
+// NamespaceFilter is a struct holding pointers to the config and the namespace lister.
+type NamespaceFilter struct {
+	c      *config.Config
```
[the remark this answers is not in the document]
marcsanmi: Good point, updated here ccdd7e20e710e890723333e9b9f3a8f5bf153429.

internal/discovery/namespace_filter.go
```diff
@@ -0,0 +1,126 @@
+package discovery
+
+import (
+	"errors"
+
+	"github.com/newrelic/nri-kubernetes/v3/internal/config"
+	"github.com/newrelic/nri-kubernetes/v3/internal/storer"
+
+	log "github.com/sirupsen/logrus"
```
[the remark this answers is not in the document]
marcsanmi: Updated by passing the logger as a dependency. Updated here ccdd7e20e710e890723333e9b9f3a8f5bf153429.

internal/config/config.go
```diff
@@ -218,38 +224,83 @@ type MTLS struct {
...
+	// MatchLabels is a list of labels to filter namespaces with.
+	MatchLabels map[string]string `mapstructure:"matchLabels"`
+	// MatchExpressions is a list of namespaces selector requirements.
+	MatchExpressions []Expression `mapstructure:"matchExpressions"`
+}
+
+// Expression hold the values to generate an expression to filter namespaces by MatchExpressions.
+type Expression struct {
+	// Key it the key of the label.
+	Key string `mapstructure:"key" json:"key"`
```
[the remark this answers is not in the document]
marcsanmi: This shouldn't be here, good catch. It was me trying a different approach and I forgot to remove it. Done in ccdd7e20e710e890723333e9b9f3a8f5bf153429.

internal/discovery/namespace_filter.go
marcsanmi: Would you suggest to implement a new cache?

alvarocabanas: What you can do, to continue using the same storer is creating another interface:
```
 type NamespaceStorer interface {
	Set(key string, value bool) int64
	Get(key string, valuePtr *bool) (int64, error)
}
```
And then create a wrapper that basically does:
```
type NamespaceInMemoryStore struct {
	InMemoryStore
}

func (ns *NamespaceInMemroyStore) func Set(key string, value bool) int64 {
   ns.InMemoryStore.Set(key, value)
}
[...]
```
WDYT?

internal/config/config.go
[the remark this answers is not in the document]
alvarocabanas: What about returning string and error and evaluating it where you call it, and logging there?

internal/discovery/namespace_filter.go
sigilioso: > Yes, I thought in setting a properly TTL and interval to fit within each integration run. WDYT?

I think it is OK, I asked because we have two different caches: one cache is handled internally by the namespace informer, an another handled by the storer. I wonder if the storer TTL and the informer resync time should be related.

internal/discovery/namespace_filter.go
```diff
@@ -0,0 +1,128 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	listersv1 "k8s.io/client-go/listers/core/v1"
+)
+
+// NamespaceFilterer provides an interface to filter namespaces.
+type NamespaceFilterer interface {
+	IsAllowed(namespace string) bool
+}
```
roobre, line 20: I wonder if this interface is too specific. What would be the advantage of having a filter specific to namespaces, as opposed to a filter that takes in `fetch.RawMetrics` and returns a bool? `NamespaceFilter` could be a particular implementation of this, looking for `metrics["namespace"]`

internal/discovery/namespace_filter.go
```diff
@@ -0,0 +1,128 @@
...
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	listersv1 "k8s.io/client-go/listers/core/v1"
+)
+
+// NamespaceFilterer provides an interface to filter namespaces.
+type NamespaceFilterer interface {
+	IsAllowed(namespace string) bool
```
roobre, line 19: Minor naming thing, but I have a weakness for single-word methods :)

```suggestion
	Match(namespace string) bool
```

internal/discovery/namespace_filter.go
```diff
@@ -0,0 +1,128 @@
...
+		logger: logger,
+	}
+}
+
+// IsAllowed checks given any namespace, if it's allowed to be scraped by using the NamespaceLister
+func (nf *NamespaceFilter) IsAllowed(namespace string) bool {
+	// By default, we scrape every namespace.
+	if nf.c == nil {
+		return true
+	}
```
roobre, line 56: All these comments could be `log.Debugf` or `log.Tracef`
```suggestion
	if nf.c == nil {
		log.Tracef("Allowing %q as selector is nil", namespace)
		return true
	}
```

internal/discovery/namespace_filter.go
```diff
@@ -0,0 +1,128 @@
...
+			if err != nil {
+				nf.logger.Errorf("listing namespaces with MatchExpressions: %v", err)
+				return true
+			}
+
+			if !containsNamespace(namespace, namespaceList) {
+				return false
+			}
+		}
+	}
```
roobre, line 100: These two snippets seems quite independent, could we extract them to separate functions? This woudl make this method way easier to test.

internal/discovery/namespace_filter.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// Save the namespace in the cache.
+	_ = nf.storer.Set(namespace, true)
```
roobre, line 103: If we have a cache, I'd love to have a single method whose only responsibility is to get the cache, the real thing, and return the second if it is not found in the first.

```go
func Match(namespace string) bool {
	if match, found := cache.Match(namespace); found {
		return match
	}

	match := real.Match(namespace)
	cache.Put(namespace, match)

	return match
}

File: internal/storer/storer.go [hunk 18]
```diff
@@ -19,6 +19,11 @@ const (
 	DefaultInterval = 15 * time.Minute
 )
 
+type Storer interface {
+	Set(key string, value interface{}) int64
+	Get(key string, valuePtr interface{}) (int64, error)
+}
```
roobre, line 25: Storer was built to be used with SDK integrations. I'd rather not reuse it for other things as it is not super good. What about just a map instead?

roobre (changes requested):

internal/config/config.go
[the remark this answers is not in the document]
gsanchezgavier: agree with Alvaro, i would leave the the caller to action the error.

internal/discovery/namespace_filter.go
marcsanmi: Updated in 7158e361f86e9713a3d2497ab1ff194a9d33b044.

internal/discovery/namespace_filter.go
marcsanmi: Yes, updated here e57ba695a157903790cc9878ad0f6cbf032beb9f

internal/discovery/namespace_filter.go
marcsanmi: Done in e57ba695a157903790cc9878ad0f6cbf032beb9f.

internal/discovery/namespace_filter.go
marcsanmi: You're right, I updated the interface definition and implementation here e57ba695a157903790cc9878ad0f6cbf032beb9f.

internal/discovery/namespace_filter.go
marcsanmi: Done in e57ba695a157903790cc9878ad0f6cbf032beb9f.

internal/discovery/namespace_filter.go
marcsanmi: After talking with @alvarocabanas and @sigilioso, we decided to move back again to the initial definition. Implementing the solution suggested above would imply some refactor in order to avoid cyclic dependencies between definition and discovery packages and to make the code maintainable. 

Done in 95f8b9d068328ecb9f115091592e9d6ab91efe9a. 

We can talk about this when you come back 🙂

internal/discovery/namespace_filter.go
marcsanmi: Finally I implemented a new cache.

internal/config/config.go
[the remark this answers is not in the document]
marcsanmi: Done in c28d9bd4f0faa4815221090d323763d6eb91d3c8

internal/discovery/namespace_filter.go
marcsanmi: Done in 89b8e6004d8f746a9c2ce9cdaa28c96d7ca696e4.

internal/storer/storer.go
marcsanmi: Done in 89b8e6004d8f746a9c2ce9cdaa28c96d7ca696e4.

internal/discovery/namespace_filter.go
marcsanmi: I added the cache to clean itself on each interval. This interval is configurable and should be smaller than the integration interval. Done in 89b8e6004d8f746a9c2ce9cdaa28c96d7ca696e4

alvarocabanas (approved):
LGTM

internal/discovery/namespace_cache.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"sync"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	// DefaultTTL is default ttl of the cache entries.
+	DefaultTTL = 10 * time.Minute
```
sigilioso, line 12: This is not used, is it?

marcsanmi: Removed this as well as the struct field, they were from the old cache. Done in 41e75891a250c10f062d391f5218e08850cb941c.

gsanchezgavier (approved):
LGTM

charts/newrelic-infrastructure/values.yaml
```diff
@@ -49,6 +49,9 @@ common:
     # behave properly. Any non-nil value will override the `lowDataMode` default.
     # @default -- `15s` (See [Low data mode](README.md#low-data-mode))
     interval:
+    # -- Config for filtering ksm and kubelet metrics by namespace.
```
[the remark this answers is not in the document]
marcsanmi: I updated it here e2fe8a97006e96411e5a4c35d774fb788a004e19. When we have the docs published I'd be good to add a link pointing to the documentation for more details.

PR #460, opened 2022-05-25
adapt integration tests to support k8s v1.23

sigilioso:
This PR includes the integration tests updated (including the corresponding static test data for kubernetes 1.23) and the required changes because we where relying on an alpha and deprecated metric which has been replaced in k8s 1.23.

As static test data changes may be noisy, reviewing commits may be easier.


src/controlplane/controlplane_test.go
```diff
@@ -40,6 +40,17 @@ const (
...
+// apiServer123DeprecationExclusion defines an specific exclusion introduced to support k8s 1.23.
+func apiServer123DeprecationExclusion(version testutil.Version) exclude.Func {
+	return func(group string, spec *definition.Spec, ent *integration.Entity) bool {
+		// apiserverStorageObjects replaces etcObjectCounts in k8s versions above 1.23
+		if testutil.IsBelow(version, testutil.Testdata123) {
+			return exclude.Metrics("apiserverStorageObjects")(group, spec, ent)
+		}
+		return exclude.Metrics("etcdObjectCounts")(group, spec, ent)
+	}
+}
```
sigilioso, line 52: I don't like including this version-related check here, but I think this is something we need to test. I considered some alternatives:

- Defining the exclusions closer to version definition (something like `version.Exclusions()`): I discarded it because the exclusion only make sense for a particular control-plane tests.
- Excluding all optional metrics y the 'all control plane test' and create a new tests for this metric: I discarded it because it may sound simpler but it is more code and harder to read.

Please, let me know your thoughts.

roobre: I get the idea but not why we are checking the version here, rather than outside: https://github.com/newrelic/nri-kubernetes/blob/3cd1b6b2ef50951fd129f62300e2d00eea6d7ace/src/controlplane/controlplane_test.go#L116

We could check the version in the test loop, and add the exclusion function for that test only, rather than checking the version inside the FxcludeFunc. WDYT?

src/controlplane/controlplane_test.go
```diff
@@ -101,8 +112,10 @@ func Test_Scraper_Autodiscover_all_cp_components(t *testing.T) {
 				t.Fatalf("running scraper: %v", err)
 			}
 
+			// Include specific specific exclusions for this version.
+			versionAsserter := asserter.Excluding(apiServer123DeprecationExclusion(version))
```
roobre, line 116: Something like (string comparison abused for brevity):
```suggestion
			if version >= testutil.Testdata123 {
				asserter = asserter.Excluding(apiServer123DeprecationExclusions())
			}
```

src/controlplane/controlplane_test.go
sigilioso: Well, it would be a bit different because we need to include a exclusion for every version (one when "version < 1.23", another when "version >= 1.23". So, including the version check in the loop would look like this:

```go
versionAsserter := asserter
if testutil.IsBelow(version, testutil.Testdata123) {
    versionAsserter = asserter.Exclude("apiserverStorageObjects")
else {
    versionAsserter = asserter.Exclude("etcdObjectCounts")
}
versionAsserter.On(i.Entitites).Assert(t)
```

If we added more version checks, I would not like to see the loop growing... So I prefer checking the version there.

Maybe the function name is confusing, maybe something like `apiserverObjectsExclusion` would be easier to read (the version is a parameter and now I see that one of its values should not be part of the function name).

src/metric/definition.go
```diff
@@ -66,10 +66,22 @@ var APIServerSpecs = definition.SpecGroups{
...
+			},
+			// apiserver_storage_objects was introduced in k8s 1.21 and replaces etcd_object_counts in 1.23
+			{
+				Name: "apiserverStorageObjects",
+				ValueFunc: fetchIfMissing(
+					prometheus.FromValueWithOverriddenName("apiserver_storage_objects", "apiserverStorageObjects"),
+					prometheus.FromValueWithOverriddenName("etcd_object_counts", "etcdObjectCounts"),
+				),
+				Type:     sdkMetric.GAUGE,
+				Optional: true,
```
roobre, line 84: Should it be optional though? `etcdObjectCounts` wasn't, so shouldn't this metric be always present, due to the `fetchIfMissing` logic?

sigilioso: I think you are right... 🤔  I guess I kept it as optional because the first draft of `fetchIfMissing` returned an error instead of an `empty.FetchedValues{}` when the metric to be replaced was not missing. I'm double-checking if everything is taken into account and removing the optional flag. Thanks!

src/controlplane/controlplane_test.go
sigilioso: Updated here: 654bdea917bc89cd313853086eb9a08ac13ccbc2

src/metric/definition.go
sigilioso: That was it, thanks again! Updated here: dee1a3e3b9ea4285d8896fc79a49bbaf9ff6f975

src/controlplane/controlplane_test.go
```diff
@@ -101,8 +101,16 @@ func Test_Scraper_Autodiscover_all_cp_components(t *testing.T) {
...
 			}
 
+			// Include specific specific exclusions that depend on version.
+			versionAsserter := asserter
+			// apiserverStorageObjects replaces etcObjectCounts in k8s versions above 1.23
+			if testutil.IsBelow(version, testutil.Testdata123) {
+				versionAsserter = versionAsserter.Excluding(exclude.Metrics("apiserverStorageObjects"))
+			} else {
+				versionAsserter = versionAsserter.Excluding(exclude.Metrics("etcdObjectCounts"))
+			}
```
roobre, line 111: Minor stuff, but I wonder why you chose to use a different variable in this case?
```suggestion
			// apiserverStorageObjects replaces etcObjectCounts in k8s versions above 1.23
			if testutil.IsBelow(version, testutil.Testdata123) {
				asserter = versionAsserter.Excluding(exclude.Metrics("apiserverStorageObjects"))
			} else {
				asserter = versionAsserter.Excluding(exclude.Metrics("etcdObjectCounts"))
			}
```

roobre (approved):

src/controlplane/controlplane_test.go
sigilioso: I'm using a different variable because `asserter` is defined outside the loop. `t.Parallel` is being used so the behaviour if we keep the same variable becomes unpredictable, besides if subtests weren't running in parallel, the last iteration would include the exclusions from all the previous iterations. Maybe keeping the same name could be more idiomatic (`asserter := asserter`) but I think a different one is more readable.

roobre: Fair enough!

roobre (approved):

PR #461, opened 2022-05-26
health: add liveness server

nadiamoe:
Right now, it is not possible for Kubernetes to know the state of the integration. While the integration currently exits whenever an error occurs, before entering the main loop it will be still waiting for some initialization to complete.

This PR adds a health server that can be used for the application to expose whether it is healthy or not. Currently, the main command of the integration sets the server to Healthy after the first successful scrape.

### DIFF
--- a/charts/newrelic-infrastructure/Chart.yaml
+++ b/charts/newrelic-infrastructure/Chart.yaml
@@ -38,7 +38,3 @@ keywords:
   - infrastructure
   - newrelic
   - monitoring
-
-annotations:
-  artifacthub.io/changes: |
-    - Added liveness probe
--- a/charts/newrelic-infrastructure/Chart.yaml
+++ b/charts/newrelic-infrastructure/Chart.yaml
@@ -8,7 +8,7 @@ sources:
   - https://github.com/newrelic/nri-kubernetes/tree/master/charts/newrelic-infrastructure
   - https://github.com/newrelic/infrastructure-agent/
 
-version: 3.5.1
+version: 3.6.0
 appVersion: 3.2.0
 
 dependencies:
@@ -38,3 +38,7 @@ keywords:
   - infrastructure
   - newrelic
   - monitoring
+
+annotations:
+  artifacthub.io/changes: |
+    - Added liveness probe
--- a/charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
+++ b/charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
@@ -80,6 +80,12 @@ spec:
             - name: "NRI_KUBERNETES_VERBOSE"
               value: {{ include "newrelic.common.verboseLog.valueAsBoolean" . | quote }}
 
+            {{- if include "nriKubernetes.controlPlane.hostNetwork" . }}
+            {{/* If we are running in hostNetwork, do not set up the health probe */}}
+            - name: "NRI_KUBERNETES_HEALTH_ADDRESS"
+              value: ""
+            {{- end }}
+
             - name: "NRI_KUBERNETES_NODENAME"
               valueFrom:
                 fieldRef:
@@ -97,6 +103,18 @@ spec:
           {{- with .Values.controlPlane.extraEnvFrom }}
           envFrom: {{ toYaml . | nindent 12 }}
           {{- end }}
+
+          {{- if not (include "nriKubernetes.controlPlane.hostNetwork" .) }}
+          ports:
+            - containerPort: 8999
+          startupProbe:
+            periodSeconds: 30
+            failureThreshold: 3
+            httpGet:
+              path: /health
+              port: 8999
+          {{- end }}
+
           volumeMounts:
             - name: nri-kubernetes-config
               mountPath: /etc/newrelic-infra/nri-kubernetes.yml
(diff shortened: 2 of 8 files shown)


cmd/nri-kubernetes/main.go
```diff
@@ -74,6 +75,17 @@ func main() {
...
+	healthServer := health.New(health.WithLogger(logger))
+
+	if c.HealthAddress != "" {
+		go func(address string) {
+			err := healthServer.ListenAndServe(address)
+			if err != nil {
+				logger.Errorf("Error starting health server: %v", err)
+			}
+		}(c.HealthAddress)
+	}
```
sigilioso, line 87: Nit: maybe the 'health server start goroutine function' could be responsibility of the healthServer itself (just to avoid adding complexity to the main function). Maybe something like this:

```go
if c.HealthAddress != "" {
    healthServer := health.New(health.WithLogger(logger))
    go healtServer.Start(c.HealdAddress)
}
```
The healthServer already has the logger, so error the handling can be done there.

WDYT?

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
```diff
@@ -97,6 +103,18 @@ spec:
           {{- with .Values.controlPlane.extraEnvFrom }}
           envFrom: {{ toYaml . | nindent 12 }}
           {{- end }}
+
+          {{- if not (include "nriKubernetes.controlPlane.hostNetwork" .) }}
+          ports:
+            - containerPort: 8999
```
sigilioso, line 109: Would it worth it to make the health port configurable?

cmd/nri-kubernetes/main.go
roobre: Fair point! When I first wrote that the logger was not inside yet, but now it makes total sense.

roobre: Giving it a second thought... It is not the use case, but if the do it that way, the caller will never have a way to see if the `Server` has errored.
Again we are not checking it here, but from the API design perspective it does not feel 100% right removing that capability.

sigilioso: Fair enough! If making the main more complex was a problem we could always use a helper to start the healthServer. 👍

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
roobre: I've started implementing this, but I'm finding UX concerns that are pretty hard to solve. The main one is that, right now, it's easy to expose the `containerPort` and to disable everything if `hostNetwork` is on. If we expose the config, that logic gets more complicated, especially if we want to override this setting depending on `hostNetwork`.

charts/newrelic-infrastructure/Chart.yaml
```diff
@@ -38,3 +38,7 @@ keywords:
   - infrastructure
   - newrelic
   - monitoring
+
+annotations:
+  artifacthub.io/changes: |
+    - Added liveness probe
```
kang-makes, line 44: NIT: adding this now will make the future changes even more painful.

I would prefer to make the effort in automating the release process than add more complexity to the already complex process.

charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
roobre: I'm not a big fan of the hardcoded probe as it is now either... Any ideas, @kang-makes ?

kang-makes: It should listen to `localhost` and it is disabled when `hostNetwork` is enabled so I have no big issues with hardcoding it.

As it could be set in the configuration file we could test in the `.Values.common.config` has a `healthAddress` and set it.

sigilioso: Is there any other already configurable port which may collide if it is set to `8999`? If we finally keep it hardcoded, that is the only cause for concern I can think of...

kang-makes: There is no configurable port in our helm charts. We could configure LOTS of things so we timeboxed and we run out of time.

It is prepared for the port to be variable but right now it is hardcoded in helpers:
https://github.com/newrelic/nri-kubernetes/blob/main/charts/newrelic-infrastructure/templates/controlplane/_agent-config_helper.tpl#L7
https://github.com/newrelic/nri-kubernetes/blob/main/charts/newrelic-infrastructure/templates/ksm/_agent-config_helper.tpl#L7
https://github.com/newrelic/nri-kubernetes/blob/main/charts/newrelic-infrastructure/templates/kubelet/_agent-config_helper.tpl#L6

kang-makes: I wanted to have preview but Github seems to be drunk (again)

Ports 8001, 8002, and 8003 are hardcoded right now.

nadiamoe:
Closing as this has largely drifted out of date and I won't continue working on it. Maintainers are welcome to rebase this if they think it's useful.

mangulonr:
Thanks @nadiamoe for your contributions to this repo

cc @ramkrishankumarN @danielstokes @dbudziwojskiNR  @kondracek-nr

PR #464, opened 2022-05-27
chart: test controlPlane obeys global.hostNetwork

roobre:


roobre:
This should not happen: `global.hostNetwork` does not affect the control plane scraper by design.

kang-makes:
https://github.com/newrelic/nri-kubernetes/blob/main/charts/newrelic-infrastructure/README.md#hostnetwork-toggle

> In versions below v3, changing the privileged mode affected the hostNetwork. We changed this behavior and now you can set pods to use hostNetwork using the corresponding [flags from the common library](https://github.com/newrelic/helm-charts/blob/master/library/common-library/README.md) (.global.hostNetwork and .hostNetwork) but the component that scrapes data from the control plane has always set hostNetwork enabled by default (Look in the [values.yaml](https://github.com/newrelic/nri-kubernetes/blob/main/charts/newrelic-infrastructure/values.yaml) for controlPlane.hostNetwork: true)

This is the intended behavior. Closing.

PR #465, opened 2022-05-30
fix asserter corner case

sigilioso:


roobre (approved):
Awesome catch, thanks for fixing this!

PR #467, opened 2022-05-30
chart: set overide_host_root as an env var rather than a config entry

roobre:
This is needed as env vars take precedence over config values. We were setting this as a config value while unaware that the image we ship set these env vars in the Dockerfile.

### DIFF
--- a/charts/newrelic-infrastructure/tests/unprivileged_override_host_root_test.yaml
+++ b/charts/newrelic-infrastructure/tests/unprivileged_override_host_root_test.yaml
@@ -1,0 +1,33 @@
+suite: Test NRIA_OVERRIDE_HOST_ROOT
+templates:
+  - templates/kubelet/daemonset.yaml
+  - templates/kubelet/scraper-configmap.yaml
+  - templates/kubelet/agent-configmap.yaml
+  - templates/kubelet/integrations-configmap.yaml
+  - templates/agent-configmap.yaml
+  - templates/secret.yaml
+tests:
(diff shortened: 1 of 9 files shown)


charts/newrelic-infrastructure/tests/unprivileged_override_host_root_test.yaml
```diff
@@ -0,0 +1,33 @@
...
+      licenseKey: test
+      cluster: test
+      privileged: false
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[1].env
+          content:
+            name: "NRIA_OVERRIDE_HOST_ROOT"
+            value: ""
+        template: templates/kubelet/daemonset.yaml
```
kang-makes, line 33: Could you test `notContains` when privileged is set to false?

kang-makes (approved):
It was a good idea when I thought about it but I didn't think about the env priority. 

I left a NIT about testing but LGMT.

Thanks for taking care of this.

charts/newrelic-infrastructure/tests/unprivileged_override_host_root_test.yaml
roobre: ... That was what this test was supposed to test. Now I wonder how it's passing.

roobre: WDYM exactly? The env var _should_ be there (set to `""`) when `privileged: false`.

roobre:
Made a quick manual test and this indeed seems to work as expected, no extraneous errors in unprivileged mode.

PR #472, opened 2022-05-31
go: upgrade to 1.18

roobre:
Go version in `go.mod` was updated in https://github.com/newrelic/nri-kubernetes/pull/403, but build pipelines did not reflect this change before this PR.

For traceability these changes were made with
```
find .github -type f  | xargs sed -i 's/1\.17/1.18/'
```
And then manually editing `e2e.yaml` to undo the wrong replacement on k8s 1.17.x.


sigilioso (approved):
LGTM, thanks for taking care of it

PR #476, opened 2022-06-03
Namespace filterer logic

alvarocabanas:
- Add the logic to filter objects by namespace for both KSM and kubelet


internal/storer/storer.go
```diff
@@ -17,6 +17,8 @@ const (
 	DefaultTTL = 10 * time.Minute
 	// DefaultInterval is default interval to execute the "garbage collection" of the cache.
 	DefaultInterval = 15 * time.Minute
+	// NamespaceStoreExtraInterval is the extra time added to the integration interval to execute the "gc"
+	NamespaceStoreExtraInterval = 5 * time.Second
```
sigilioso, line 21: Maybe it would make sense to define this in `discovery` package (in the same place the NamespaceFilter is defined).  I think defining it here feels like coupling the storer with a particular instance of the storer, WDYT?

cmd/nri-kubernetes/main.go
```diff
@@ -120,9 +122,19 @@ func main() {
 		os.Exit(exitClients)
 	}
 
+	var nsFilter *discovery.NamespaceFilter
```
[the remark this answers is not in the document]
sigilioso: I think that no-op filter would make this PR and #457 independent. This could be merged even before, as long as, after both PRs are merged we create a follow-up PR just modifying the instance creation.

src/metric/definition.go
```diff
@@ -493,8 +493,9 @@ var EtcdQueries = []prometheus.Query{
 // KSMSpecs are the metric specifications we want to collect from KSM.
 var KSMSpecs = definition.SpecGroups{
 	"replicaset": {
-		IDGenerator:   prometheus.FromLabelValueEntityIDGenerator("kube_replicaset_created", "replicaset"),
-		TypeGenerator: prometheus.FromLabelValueEntityTypeGenerator("kube_replicaset_created"),
+		IDGenerator:     prometheus.FromLabelValueEntityIDGenerator("kube_replicaset_created", "replicaset"),
+		TypeGenerator:   prometheus.FromLabelValueEntityTypeGenerator("kube_replicaset_created"),
+		NamespaceGetter: prometheus.FromLabelGetNamespace(),
```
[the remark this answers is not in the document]
sigilioso: I think `kubeletSpecs` use the `NamespaceGetter` from `kubeletMetrics` instead of the `prometheus` one.

src/prometheus/definition.go
```diff
@@ -137,6 +137,18 @@ func FromLabelsValueEntityIDGeneratorForPendingPods() definition.EntityIDGenerat
...
+	return func(metrics definition.RawMetrics) string {
+		for _, metric := range metrics {
+			m, ok := metric.(Metric)
+			if ok {
+				return m.Labels["namespace"]
+			}
+		}
+		return ""
+	}
+}
```
sigilioso, line 150: As it does no receive any arguments, it could be defined like this:

```go
func FromLabelGetNamespace(metrics definition.RawMetrics) string {
// ...
}
```
and then, reference the static function `prometheus.FromLabelGetNamespace` (without `()`) in the spec file.

WDYT?

src/kubelet/metric/metric.go
```diff
@@ -330,6 +330,16 @@ func FromRawGroupsEntityTypeGenerator(groupLabel string, rawEntityID string, gro
...
 
+func FromLabelGetNamespace() definition.NamespaceGetterFunc {
+	return func(metrics definition.RawMetrics) string {
+		if ns, ok := metrics["namespace"].(string); ok {
+			return ns
+		}
+
+		return ""
+	}
+}
```
sigilioso, line 341: The same that the previous comment in `prometheus.definition.go` about the namespace getter function definition.

internal/storer/storer.go
alvarocabanas: Yes, this is temporary, I still need to see how does the other branch evolve, since maybe there is a custom interface for the ns storer.

src/metric/definition.go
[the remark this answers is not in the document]
alvarocabanas: Exactly, that is the thing, KSM and Kubelet have different getters

src/prometheus/definition.go
alvarocabanas: yes, fair enough.

cmd/nri-kubernetes/main.go
```diff
@@ -120,9 +122,19 @@ func main() {
...
+			clients.k8s,
+			storer.NewInMemoryStore(c.Interval, c.Interval+storer.NamespaceStoreExtraInterval, logger),
+			nil,
+		)
+	}
+
 	var kubeletScraper *kubelet.Scraper
 	if c.Kubelet.Enabled {
-		kubeletScraper, err = setupKubelet(c, clients)
+		kubeletScraper, err = setupKubelet(c, clients, nsFilter)
```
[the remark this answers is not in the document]
alvarocabanas: Yes, I can move it to the setup and I will create a main_test ensuring the setupKubelet constructs what I want, and the same with setupKSM

cmd/nri-kubernetes/main.go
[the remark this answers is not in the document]
alvarocabanas: The NOOPFilter was a great idea but doesn't make sense because the  IntegrationPopulator function in populate recieves a config but doesn't have constructor, so it can be nil, for example in the unit tests.

cmd/nri-kubernetes/main.go
```diff
@@ -201,13 +213,18 @@ func runScrapers(c *config.Config, ksmScraper *ksm.Scraper, kubeletScraper *kube
...
 	providers := ksm.Providers{
 		K8s: clients.k8s,
 		KSM: clients.ksm,
 	}
 
-	ksmScraper, err := ksm.NewScraper(c, providers, ksm.WithLogger(logger))
+	scraperOpts := []ksm.ScraperOpt{ksm.WithLogger(logger)}
+	if nsFilter != nil {
+		scraperOpts = append(scraperOpts, ksm.WithFilterer(nsFilter))
+	}
```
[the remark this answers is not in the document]
alvarocabanas: This has been changed

src/kubelet/metric/metric.go
alvarocabanas: Done in last commit

src/prometheus/definition.go
alvarocabanas: done

cmd/nri-kubernetes/main.go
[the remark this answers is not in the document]
alvarocabanas: done

cmd/nri-kubernetes/main.go
```diff
@@ -244,7 +258,20 @@ func setupKubelet(c *config.Config, clients *clusterClients) (*kubelet.Scraper,
...
+
+	if c.NamespaceSelector != nil {
+		nsFilter := discovery.NewNamespaceFilter(
+			c.NamespaceSelector,
+			clients.k8s,
+			storer.NewInMemoryStore(c.Interval, c.Interval+storer.NamespaceStoreExtraInterval, logger),
+			logger,
+		)
+		scraperOpts = append(scraperOpts, kubelet.WithFilterer(nsFilter))
+	}
```
sigilioso, line 272: would it make sense a helper function to avoid repeating the `nsFilter` creation?

```go
nsFilter := buildNamespaceFilter(c, clients.k8s, logger)
if nsFilter != nil {
    scraperOpts = append(scraperOpts, kubelet.WithFilterer(nsFilter))
}
```

cmd/nri-kubernetes/main.go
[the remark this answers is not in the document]
sigilioso: Then wouldn't it make sense to include a config builder (or constructor)? We could use it both in unit tests (the `testConfig` would be replaced there) and in its usage in `scrape_jobs`.

Another alternative would to not to include the filterer in config at all. This could be achieved with a second parameter for `IntegrationPopulator` or making it a struct (now it has more responsibilities than it previously had).

cmd/nri-kubernetes/main.go
alvarocabanas: sure.

cmd/nri-kubernetes/main.go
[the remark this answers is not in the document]
alvarocabanas: As we discussed in our call, we will leave it as it is to not make a bigger refactor (that can be done in the future in the Populator and other parts), right now the IntegrationPopulator get Namespace would be called if the filterer is not null and having the Noop doesn't add any value.

src/kubelet/kubelet_test.go
```diff
@@ -84,6 +91,53 @@ func TestScraper(t *testing.T) {
...
 
+func TestScraper_FilterNamespace(t *testing.T) {
+	// We test with a specific version to not modify number of entities
+	version := testutil.Version(testutil.Testdata122)
+
+	t.Run(fmt.Sprintf("for_version_%s", version), func(t *testing.T) {
+		testServer, err := version.Server()
+		if err != nil {
+			t.Fatalf("Cannot create fake kubelet server: %v", err)
+		}
```
sigilioso, line 102: Nit: as testify is already been used, I'd change the error-check ifs by `require.NoError`

sigilioso (approved):
👍 LGTM, I've just let a nit (sorry I didn't see it in the first review)

cmd/nri-kubernetes/main.go
```diff
@@ -244,14 +251,34 @@ func setupKubelet(c *config.Config, clients *clusterClients) (*kubelet.Scraper,
...
+func buildNamespaceFilter(c *config.Config, clients *clusterClients) *discovery.CachedNamespaceFilter {
+	nsFilter := discovery.NewNamespaceFilter(
+		c.NamespaceSelector,
+		clients.k8s,
+		logger,
+	)
+
+	return discovery.NewCachedNamespaceFilter(
+		nsFilter,
+		discovery.NewNamespaceInMemoryStore(c.Interval, logger),
```
marcsanmi, line 278: Probably I'd set this interval to be slightly smaller than the integration one, as I'd be good that the cache is cleaned before the new run.

src/kubelet/kubelet_test.go
alvarocabanas: I've done it with require.noError in this new tests but didn't modify it for the old integration tests since the errors are wrapped and I am not sure if adding this context is really meaningful

cmd/nri-kubernetes/main.go
marcsanmi: There's no TTL in the new cache, it just calls the vacuum on each interval.

alvarocabanas: I've modified it with the new call to Vacuum, you can see the change in the last 2 commits.

cmd/nri-kubernetes/main.go
```diff
@@ -202,7 +206,7 @@ func runScrapers(c *config.Config, ksmScraper *ksm.Scraper, kubeletScraper *kube
 	return nil
 }
 
-func setupKSM(c *config.Config, clients *clusterClients) (*ksm.Scraper, error) {
+func setupKSM(c *config.Config, clients *clusterClients, namespaceCache *discovery.NamespaceInMemoryStore) (*ksm.Scraper, error) {
```
sigilioso, line 209: Nit: now we need to add a new parameter to have a common cache, maybe it would make more sense to have your initial approach (directly adding `nsFilter` as an argument to `setupKSM` and `setupKubelet`)

marcsanmi (approved):
LGTM 👍

sigilioso (approved):
LGTM, I left a nit just in case you want to re-consider your previous approach.

cmd/nri-kubernetes/main.go
alvarocabanas: I prefer to do it like this, since the cache is always going to be vacuumed, even if there is no namespaceFilter, moving the logic outside the setupKSM and setupKubelet would make me create the check namespace != nil outside and an extra nsFilter != nil WDYT?

PR #487, opened 2022-06-13
Move filter creation to KSM & Kubelet methods

alvarocabanas:


marcsanmi (approved):
👍

PR #489, opened 2022-06-13
Update chart to the latest version of nri-kubernetes with namespace filtering

alvarocabanas:
- Update chart to the latest version of nri-kubernetes with namespace filtering

- Update clusterrole to allow ns watch and list

### DIFF
--- a/charts/newrelic-infrastructure/Chart.yaml
+++ b/charts/newrelic-infrastructure/Chart.yaml
@@ -8,8 +8,8 @@ sources:
   - https://github.com/newrelic/nri-kubernetes/tree/master/charts/newrelic-infrastructure
   - https://github.com/newrelic/infrastructure-agent/
 
-version: 3.5.5
-appVersion: 3.2.0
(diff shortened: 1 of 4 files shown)


charts/newrelic-infrastructure/Chart.yaml
```diff
@@ -8,8 +8,8 @@ sources:
   - https://github.com/newrelic/nri-kubernetes/tree/master/charts/newrelic-infrastructure
   - https://github.com/newrelic/infrastructure-agent/
 
-version: 3.5.5
```
paologallinaharbur, line 11: Can you generate as well new docs?

I usually do it running:
```
 docker run --rm --volume "$(pwd):/helm-docs" -u $(id -u) jnorwood/helm-docs:latest 
```

alvarocabanas: Sure!

paologallinaharbur (approved):

Issue #435, opened 2022-05-18
Nodes `fullHostname` appears as a service name if any service is bound to the host IP


roobre:
## Description

If in the cluster there is a service having the host's IP address as an endpoint (e.g. a pod with `hostNetwork: true`), CoreDNS will return the service name as the reverse for the Host's IP, rather than forwarding this upstream.

This causes the agent to report that service name as the `fullHostname` for the node.

## Expected Behavior

`fullHostname` for the nodes should be the FQDN of the node, not a service name.

roobre:
To be discussed with @rubenruizdegauna

jayntguru:
Just wanted to say that we're seeing this as of V3 and it's causing some issues.

paologallinaharbur:
This issue is fixed in the nri-kubernetes chart, but it will take some days to get propagated to nri-bundle

PR #492, opened 2022-06-14
fix(hostname): overriding fullhostname

paologallinaharbur:
Fix https://github.com/newrelic/nri-kubernetes/issues/435

After some discussion, to avoid the fullHostname to be changing depending on the service bounded to the same nodeIP, the only path forward is to overwrite that with the node name.

 - In cloud environments the GUID is computed thanks to ID coming from the different vendors (es: ARN). 
 - On premise an unique identifier is not available and this change increases the chances of collisions (but [different parameters](https://github.com/newrelic/infrastructure-agent/blob/c19c3bad1913255fa1438775eb10b77ecb5f09c8/pkg/helpers/fingerprint/fingerprint.go#L61) are taken into account ). However, without the change the GUID could change depending on the fullHostname and therefore depending on the services running in the cluster. 

Notice that we need to specify both short and full hostname. Due to two separate issues.

### DIFF
--- a/charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
+++ b/charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
@@ -123,17 +123,8 @@ spec:
                   name: {{ include "newrelic.common.license.secretName" . }}
                   key: {{ include "newrelic.common.license.secretKeyName" . }}
 
-            - name: "NRIA_OVERRIDE_HOSTNAME_SHORT"
-              valueFrom:
-                fieldRef:
-                  apiVersion: "v1"
-                  fieldPath: "spec.nodeName"
-
-            - name: "NRIA_OVERRIDE_HOSTNAME"
-              valueFrom:
-                fieldRef:
-                  apiVersion: "v1"
-                  fieldPath: "spec.nodeName"
+            - name: "NRIA_DNS_HOSTNAME_RESOLUTION"
+              value: "false"
 
             - name: "K8S_NODE_NAME"
               valueFrom:
--- a/charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
+++ b/charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
@@ -129,6 +129,12 @@ spec:
                   apiVersion: "v1"
                   fieldPath: "spec.nodeName"
 
+            - name: "NRIA_OVERRIDE_HOSTNAME"
+              valueFrom:
+                fieldRef:
+                  apiVersion: "v1"
+                  fieldPath: "spec.nodeName"
+
             - name: "K8S_NODE_NAME"
               valueFrom:
                 fieldRef:
--- a/charts/newrelic-infrastructure/templates/ksm/deployment.yaml
+++ b/charts/newrelic-infrastructure/templates/ksm/deployment.yaml
@@ -117,6 +117,12 @@ spec:
                   apiVersion: "v1"
                   fieldPath: "spec.nodeName"
 
+            - name: "NRIA_OVERRIDE_HOSTNAME"
+              valueFrom:
+                fieldRef:
+                  apiVersion: "v1"
+                  fieldPath: "spec.nodeName"
+
             - name: "K8S_NODE_NAME"
               valueFrom:
                 fieldRef:
--- a/charts/newrelic-infrastructure/templates/ksm/deployment.yaml
+++ b/charts/newrelic-infrastructure/templates/ksm/deployment.yaml
@@ -111,17 +111,8 @@ spec:
                   name: {{ include "newrelic.common.license.secretName" . }}
                   key: {{ include "newrelic.common.license.secretKeyName" . }}
 
-            - name: "NRIA_OVERRIDE_HOSTNAME_SHORT"
-              valueFrom:
-                fieldRef:
-                  apiVersion: "v1"
-                  fieldPath: "spec.nodeName"
-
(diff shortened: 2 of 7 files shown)


charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml
```diff
@@ -129,6 +129,12 @@ spec:
                   apiVersion: "v1"
                   fieldPath: "spec.nodeName"
 
+            - name: "NRIA_OVERRIDE_HOSTNAME"
+              valueFrom:
+                fieldRef:
+                  apiVersion: "v1"
+                  fieldPath: "spec.nodeName"
+
```
kang-makes, line 137: I would ask for review to @newrelic/caos here. Bur while we decide if we would have to ask them or not this change does not seem correct. because we still have `NRIA_OVERRIDE_HOSTNAME_SHORT`.

We should only need `NRIA_OVERRIDE_HOSTNAME` for this change. We may also want to disable (or do a test) [disabling DNS resolution](https://github.com/newrelic/infrastructure-agent/blob/6832e08d7bf4fdb4589138cf88ac0bd071f6f445/pkg/config/config.go#L195-L200): 
```yaml
// DnsHostnameResolution When true, the full hostname is resolved by performing a reverse lookup of the hosts
// address; otherwise, it will be retrieved with the hostname command on Linux, and from the TCP/IP parameters of
// the registry on Windows.
// Default: True
// Public: Yes
DnsHostnameResolution bool `yaml:"dns_hostname_resolution" envconfig:"dns_hostname_resolution"`
```

paologallinaharbur: - The `NRIA_OVERRIDE_HOSTNAME_SHORT` is still needed even if you specify `NRIA_OVERRIDE_HOSTNAME` otherwise the short hostname get resolved.
Screen Shot 2022-06-14 at 4 00 53 PM

 - However, it is possible that `DnsHostnameResolution=false` is enough. But as far as I understand it gets the result of `hostname` command for both full and short, that is the name of the pod. But I might have misunderstood you
Screen Shot 2022-06-14 at 4 00 03 PM

paologallinaharbur: Overriding both *and* disabling the DNS resolution could be useful to reduce some calls and save some cpu. But as far as I understood from the code if hostname are overwritten no resolution is performed. But @rubenruizdegauna can clarify if it is better to disable them anyway
Screen Shot 2022-06-14 at 4 05 38 PM

rubenruizdegauna: If both short and full are overridden, then no DNS resolution will be made, so I would stick to these.

charts/newrelic-infrastructure/templates/ksm/deployment.yaml
```diff
@@ -117,6 +117,12 @@ spec:
                   apiVersion: "v1"
                   fieldPath: "spec.nodeName"
 
+            - name: "NRIA_OVERRIDE_HOSTNAME"
+              valueFrom:
+                fieldRef:
+                  apiVersion: "v1"
+                  fieldPath: "spec.nodeName"
+
```
roobre, line 125: Do we need these on the non-kubelet pods? They are running with `IS_FORWARD_ONLY`

kang-makes: Yep. We had a GTS ticket about kube-events also hammering the DNS: https://github.com/newrelic/helm-charts/pull/670

So this change should also be done in all the other charts @paologallinaharbur

kang-makes: I am elaborating a bit further. 

For pods that are not sending the hostname because they are in forwarding mode, it may not make sense because it does not send information about the hostname although they are hammering the DNS server, and as [Ruben pointed out](https://github.com/newrelic/nri-kubernetes/pull/492#discussion_r896916188) "If both short and full are overridden, then no DNS resolution will be made, so I would stick to these."

So we are not solving only the issue of hammering DNS by putting a good "short hostname" but also by not making any DNS query.

paologallinaharbur: Yes it is **not** needed to avoid to send the fullHostname. 

However, I prefer to add it as kang-makes mentioned to:
 - avoid useless DNS queries
 - to be sure to have the exact same behaviour of hostnames across the three components.

roobre: If we want to avoid hammering the DNS, I'd rather pass `DnsHostnameResolution=false` than overriding the hostname.

My main concern is that the agent in these pods should not report any hostname altogether. If we set this to the same name with which other agents _will_ be reporting, we might hide bugs, or confuse ourselves later.

rubenruizdegauna: If the hostname is overridden (`override_hostname`) then the DNS will not be queried

paologallinaharbur: Then I change it to leverage `NRIA_DNS_HOSTNAME_RESOLUTION` for KSM and controlPlane

roobre (approved):
LGTM!
Should we add a dumb test for this, or isn't it necessary?

paologallinaharbur:
dumb tests added 😄 , I'll merge as soon as the pipeline succeeds

PR #493, opened 2022-06-15
Wrap noData error with filter specific message

alvarocabanas:

### DIFF
--- a/src/ksm/scraper.go
+++ b/src/ksm/scraper.go
@@ -1,6 +1,7 @@
 package ksm
 
 import (
+	"errors"
 	"fmt"
 	"net/url"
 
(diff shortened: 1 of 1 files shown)


src/ksm/scraper.go
```diff
@@ -154,7 +157,10 @@ func (s *Scraper) Run(i *integration.Integration) error {
 	}
 
 	if !populated {
-		return fmt.Errorf("KSM data was not populated after trying all endpoints")
+		if s.Filterer != nil {
+			return fmt.Errorf("%v, check if namespace filters are set properly", errNoData)
```
paologallinaharbur, line 161: %w?

alvarocabanas: Yes, thanks for spotting it!, it works but it's not correct.

alvarocabanas: Fixed in https://github.com/newrelic/nri-kubernetes/pull/493/commits/1caff07f4c9e8c1ea91ae567437481833b46c500

paologallinaharbur (approved):

PR #494, opened 2022-06-15
Parse matchLabels bool values as strings

marcsanmi:
Closes https://github.com/newrelic/newrelic-coreint/issues/342.

Viper even when unmarshalling to a `map[string]string` a bool values it magically parse it as `"1"`. This PR fixes this by parsing it as `"true"`.


alvarocabanas (approved):
LGTM

PR #495, opened 2022-06-15
Parse matchLabels bool values as strings

marcsanmi:
Closes https://github.com/newrelic/newrelic-coreint/issues/342.

Viper even when unmarshalling to a `map[string]string` a bool values it parses it as `"1"`. This PR fixes this by parsing it as `"true"`.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -226,7 +226,7 @@ type MTLS struct {
 // NamespaceSelector contains config options for filtering namespaces.
 type NamespaceSelector struct {
 	// MatchLabels is a list of labels to filter namespaces with.
-	MatchLabels map[string]string `mapstructure:"matchLabels"`
+	MatchLabels map[string]interface{} `mapstructure:"matchLabels"`
 	// MatchExpressions is a list of namespaces selector requirements.
 	MatchExpressions []Expression `mapstructure:"matchExpressions"`
 }
@@ -315,5 +315,36 @@ func LoadConfig(filePath string, fileName string) (*Config, error) {
 		return nil, err
 	}
 
+	if cfg.NamespaceSelector != nil && cfg.NamespaceSelector.MatchLabels != nil {
+		matchLabels := make(map[string]interface{})
+		for k, v := range cfg.NamespaceSelector.MatchLabels {
+			str, err := ParseValueToString(v)
+			if err != nil {
+				return &cfg, err
+			}
+			matchLabels[k] = str
+		}
+
+		cfg.NamespaceSelector.MatchLabels = matchLabels
+	}
+
 	return &cfg, nil
+}
+
+func ParseValueToString(val interface{}) (string, error) {
+	var str string
+	switch v := val.(type) {
+	case string:
+		str = v
+	case bool:
+		str = strconv.FormatBool(v)
+	case int:
+		str = strconv.FormatInt(int64(v), 10)
+	case float32, float64:
+		str = fmt.Sprintf("%f", v)
+	default:
+		return "", fmt.Errorf("parsing matchLabels, invalid value type %T for val: %v", v, val)
+	}
+
+	return str, nil
 }
--- a/internal/discovery/namespace_filter_test.go
+++ b/internal/discovery/namespace_filter_test.go
@@ -44,7 +44,7 @@ func TestNamespaceFilterer_IsAllowed(t *testing.T) {
 				"newrelic.com/scrape": "true",
 			},
 			namespaceSelector: config.NamespaceSelector{
-				MatchLabels: map[string]string{
+				MatchLabels: map[string]interface{}{
 					"newrelic.com/scrape": "true",
 				},
 			},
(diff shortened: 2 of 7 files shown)


internal/discovery/namespace_filter_test.go
```diff
@@ -44,7 +44,7 @@ func TestNamespaceFilterer_IsAllowed(t *testing.T) {
 				"newrelic.com/scrape": "true",
 			},
 			namespaceSelector: config.NamespaceSelector{
-				MatchLabels: map[string]string{
+				MatchLabels: map[string]interface{}{
```
kang-makes, line 47: Is it really needed to support this? Kubernetes does not:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: test
  labels:
    test: test
    test2: test2
    algo: true
    algo2: 1
data:
  key: value
```
Gives this error:
```
% kubectl apply -f cm.yaml
error: unable to decode "cm.yaml": json: cannot unmarshal bool into Go struct field ObjectMeta.metadata.labels of type string
error: unable to decode "cm.yaml": json: cannot unmarshal number into Go struct field ObjectMeta.metadata.labels of type string
```

marcsanmi: Yes, that's true... we could directly fail. We agreed with @alvarocabanas to support it but maybe it doesn't make sense at all.

alvarocabanas (approved):
LGTM

internal/discovery/namespace_filter_test.go
alvarocabanas: For me as a user I would like to have it as we did, allowing me to bot put bool or string but if it's a kubernetes standard...

internal/discovery/namespace_filter_test.go
```diff
@@ -53,7 +53,7 @@ func TestNamespaceFilterer_IsAllowed(t *testing.T) {
 		"match_labels_excluded_namespaces_not_allowed": {
 			namespaceLabels: labels.Set{"newrelic.com/scrape": "false"},
 			namespaceSelector: config.NamespaceSelector{
-				MatchLabels: map[string]string{
+				MatchLabels: map[string]interface{}{
```
paologallinaharbur, line 56: could we add a new test that would not pass with the old implementation? to show which use-case we are trying to cover

internal/discovery/namespace_filter_test.go
paologallinaharbur: (I hope I have enough context) 
Basically we are deciding to be more flexible allowing them to set:
```
    namespaceSelector: {}
    # If you want to include only namespaces with a given label you could do so by adding:
    # matchLabels:
    #   newrelic.com/scrape: true
```
That now it would be interpreted as `newrelic.com/scrape: "1"`

I get the point of @kang-makes and I agree with it. The only issue is that before it is very error prone. From my point of view could decide either to 
 - do what is implemented here knowing that we are being less strict than k8s it could be an issue at some point
 - fail with a nice message if we detect a bool/int
 - use the default parser adding an example with `"true"` and a comment explaining that `"` are needed 
 

 Maybe the last one is my favourite one, but I am fine with all of them

internal/config/config.go
```diff
@@ -226,7 +226,7 @@ type MTLS struct {
 // NamespaceSelector contains config options for filtering namespaces.
 type NamespaceSelector struct {
 	// MatchLabels is a list of labels to filter namespaces with.
-	MatchLabels map[string]string `mapstructure:"matchLabels"`
+	MatchLabels map[string]interface{} `mapstructure:"matchLabels"`
```
paologallinaharbur, line 229: Could we leverage from metav1:
```
// A label selector is a label query over a set of resources. The result of matchLabels and
// matchExpressions are ANDed. An empty label selector matches all objects. A null
// label selector matches no objects.
// +structType=atomic
type LabelSelector struct {
	// matchLabels is a map of {key,value} pairs. A single {key,value} in the matchLabels
	// map is equivalent to an element of matchExpressions, whose key field is "key", the
	// operator is "In", and the values array contains only "value". The requirements are ANDed.
	// +optional
	MatchLabels map[string]string `json:"matchLabels,omitempty" protobuf:"bytes,1,rep,name=matchLabels"`
	// matchExpressions is a list of label selector requirements. The requirements are ANDed.
	// +optional
	MatchExpressions []LabelSelectorRequirement `json:"matchExpressions,omitempty" protobuf:"bytes,2,rep,name=matchExpressions"`
}
```

also in the package labels there is something similar

paologallinaharbur: we use it for example [here](https://github.com/newrelic/newrelic-infra-operator/blob/a945b9a1b6d9f0f0f7065c97d5978dc0c4041acf/internal/mutator/pod/agent/injector.go#L86)

(I know that this PR as a different scope, I just noticed it now)

internal/discovery/namespace_filter_test.go
marcsanmi: Do you mean `true` as string? I added the tests in the config tests here https://github.com/newrelic/nri-kubernetes/pull/495/files#diff-78b78dc33eda1ed8d509bfa5a00f66709af57bd30d971cbaec454571a91c3e62R47

paologallinaharbur (approved):
🎉

kang-makes (approved):
Great work :)

PR #496, opened 2022-06-15
Add nrFiltered attribute to namespaceSample when filtering by namespace

alvarocabanas:

### DIFF
--- a/src/definition/populate.go
+++ b/src/definition/populate.go
@@ -9,6 +9,8 @@ import (
 	"github.com/newrelic/nri-kubernetes/v3/internal/discovery"
 )
 
+const NamespaceGroup = "namespace"
+
 // GuessFunc guesses from data.
 type GuessFunc func(clusterName, groupLabel, entityID string, groups RawGroups) (string, error)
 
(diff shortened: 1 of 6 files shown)


src/definition/populate.go
```diff
@@ -65,7 +67,10 @@ func IntegrationPopulator(config *IntegrationPopulateConfig) (bool, []error) {
 				if nsGetter := config.Specs[groupLabel].NamespaceGetter; nsGetter != nil {
 					ns := nsGetter(metrics)
 					if !config.Filterer.IsAllowed(ns) {
-						continue
+						if groupLabel != NamespaceGroup {
```
alvarocabanas, line 70: Here we are breaking the Genericness of the function, but it was the best way we found to add this label

paologallinaharbur (approved):

src/metric/definition.go
```diff
@@ -727,9 +727,8 @@ var KSMSpecs = definition.SpecGroups{
 		},
 	},
 	"hpa": {
-		IDGenerator:     prometheus.FromLabelValueEntityIDGenerator("kube_hpa_labels", "hpa"),
-		TypeGenerator:   prometheus.FromLabelValueEntityTypeGenerator("kube_hpa_labels"),
-		NamespaceGetter: prometheus.FromLabelGetNamespace,
+		IDGenerator:   prometheus.FromLabelValueEntityIDGenerator("kube_hpa_labels", "hpa"),
+		TypeGenerator: prometheus.FromLabelValueEntityTypeGenerator("kube_hpa_labels"),
```
paologallinaharbur, line 731: horizontalpodautoscalers          hpa          autoscaling/v1                         true         HorizontalPodAutoscaler

alvarocabanas: I've added the filtering because HPA are indeed namespaced as Paolo pointed out from the kubectl api-resources call.

marcsanmi (approved):
LGTM 👍

Issue #500, opened 2022-06-24
Passthrough environment are difficult to be configured


paologallinaharbur:
### What broke
[Context]( https://docs.newrelic.com/docs/infrastructure/install-infrastructure-agent/configuration/configure-infrastructure-agent/#passthrough) of the broken feature:

Starting with infrastructure agent 1.14.0, environment variables can be defined in the integrations config file with the {{}} notation. For example:
```
integration_name: com.newrelic.redis
instances:
  - name: redis-metrics
    command: metrics
    arguments:
      hostname: localhost
      port: 6379
      password: {{MY_REDIS_PASSWORD}}
```

Having in the yaml `"{{ MY_REDIS_PASSWORD }}"` or `{{ MY_REDIS_PASSWORD }}` or
```
      password: |-
        {{MY_REDIS_PASSWORD}}
```

results in error at templating time.
Having in the yaml `\{\{ MY_REDIS_PASSWORD \}\}` is rendered but then the agent does not perform the substitution

In V2 `password: "{{MY_REDIS_PASSWORD}}"` used to work

### Workaround 
 The passthrough env variables uses unfortunately the very same chars `{{}}` than Helm and therefore the template cannot be rendered correctly.

I tested locally a workaround that seems to be working:
Having in the values:

```
[...]
      integrations:
        - name: nri-rabbitmq
          env:
            HOSTNAME: '{{ printf "{{ NR_RMQ_HOSTNAME }}" }}'
            USERNAME: '{{ printf "{{ NR_RMQ_USERNAME }}" }}'
            PASSWORD: '{{ printf "{{ NR_RMQ_PASSWORD }}" }}' 
[...]
```

and just for testing setting:
```
[...]
  kubelet:
    extraEnv:
      - name: NR_RMQ_HOSTNAME
        value: secret-hostname
      - name: NR_RMQ_USERNAME
        value: secret-username
      - name: NR_RMQ_PASSWORD
        value: secret-password 
[...]
```

The section in the integration config is rendered as:

```
[...]
    - env:                
        HOSTNAME: '{{ NR_RMQ_HOSTNAME }}'
        PASSWORD: '{{ NR_RMQ_PASSWORD }}'
        USERNAME: '{{ NR_RMQ_USERNAME }}'
[...]
```

In the agent logs I can see that the env vars have been correctly populated

```
[...]

PR #501, opened 2022-06-24
fix(chart): removing doublerender to support old passthorugh notation

paologallinaharbur:
I checked and that tpl was there since the very first implementation of v3, not really sure why we needed double templating (in that case we need to find a different solution). 

I am not aware of an use case based on that.

Fix https://github.com/newrelic/nri-kubernetes/issues/500


sigilioso (approved):
👍 thanks for taking care of this
 
As you exposed in #500 , here we are fixing an unintentional breaking change (this was working on v2). 

Maybe rendering was not a bad idea because then _helm variables_ can also be used in `integrations` values. However, that makes variables passthrough really hard (consumers need the printf workaround or something similar). So I'd rather prefer the approach you have implemented here.

Issue #397, opened 2022-04-14
Control plane "metric attributes exceeds 240 limit, some might be lost"


paologallinaharbur:
The agent limits the number of attributes to 240 for several reasons.

Our integration was not enforcing this limit, but most of the samples have a fixed number of attributes unless the users attaches hundreds of labels or custom attributes.

Unfortunately some samples gather metrics from prometheus  automatically transforming them into attributes. If cardinality of the dimensions is too big we would overcome the 240 limit and the sample is dropped entirely.

```
apiserverRequestDelta_verb_VERB_code_CODE | Difference of the number of apiserver requests, broken out for each verb and HTTP response code
-- | --
apiserverRequestRate_verb_VERB_code_CODE | Rate of apiserver requests, broken out for each verb and HTTP response code
restClientRequestsDelta_code_CODE_method_METHOD | Difference of the number of HTTP requests, partitioned by method and code
restClientRequestsRate_code_CODE_method_METHOD | Rate of the number of HTTP requests, partitioned by method and code
etcdObjectCounts_resource_RESOURCE-KIND
```

---

In particular a user sent us a payload affected by this issue and the "exploding" dimensions are:
 - etcdObjectCounts in particular with many CRD
 - apiserverRequestsDelta
 - apiserverRequestsRate

---

In big cluster we could do some of the following:
 - split samples if the #attributes>240
 - give the possibility to user to filter any of the metrics (this could helm in different scenarios)
 - avoid reporting metrics for any code+method have a 0 rate -> risk to confuse missing data with 0
 
 
 [blocker for apiServer monitoring in large cluster ]

daboom:
Hey guys, are there any news on this or any tests I can do on the cluster we're trying to get metrics from?

daboom:
> Hey guys, are there any news on this or any tests I can do on the cluster we're trying to get metrics from?
any news?

PR #507, opened 2022-06-29
Filter k8s metrics by label values

marcsanmi:
This PR adds the capability to filter k8s metrics by label values.

This will avoid errors like this one  https://github.com/newrelic/nri-kubernetes/issues/397 we faced in the past.

### DIFF
--- a/src/controlplane/controlplane_test.go
+++ b/src/controlplane/controlplane_test.go
@@ -109,7 +109,13 @@ func Test_Scraper_Autodiscover_all_cp_components(t *testing.T) {
 
 			// apiserverStorageObjects replaces etcObjectCounts in k8s versions above 1.23
 			if testutil.IsBelow(version, testutil.Testdata123) {
-				versionAsserter = versionAsserter.Excluding(exclude.Metrics("apiserverStorageObjects"))
+				versionAsserter = versionAsserter.Excluding(
+					exclude.Metrics(
+						"apiserverStorageObjects",
+						"apiserverCurrentInflightRequestsMutating",
+						"apiserverCurrentInflightRequestsReadOnly",
+					),
+				)
 			} else {
 				versionAsserter = versionAsserter.Excluding(exclude.Metrics("etcdObjectCounts"))
 			}
--- a/src/metric/definition.go
+++ b/src/metric/definition.go
@@ -50,19 +50,23 @@ var APIServerSpecs = definition.SpecGroups{
 			},
 			{
 				Name: "apiserverCurrentInflightRequestsMutating",
-				ValueFunc: prometheus.FromValueWithOverriddenName(
+				ValueFunc: prometheus.FromValueWithLabelsFilter(
 					"apiserver_current_inflight_requests",
 					"apiserverCurrentInflightRequestsMutating",
-					prometheus.IncludeOnlyWhenLabelMatchFilter(map[string]string{"request_kind": "mutating"}),
+					prometheus.IncludeOnlyWhenLabelMatchFilter(map[string]string{
+						"request_kind": "mutating",
+					}),
 				),
 				Type: sdkMetric.GAUGE,
 			},
 			{
 				Name: "apiserverCurrentInflightRequestsReadOnly",
-				ValueFunc: prometheus.FromValueWithOverriddenName(
+				ValueFunc: prometheus.FromValueWithLabelsFilter(
 					"apiserver_current_inflight_requests",
 					"apiserverCurrentInflightRequestsReadOnly",
-					prometheus.IncludeOnlyWhenLabelMatchFilter(map[string]string{"request_kind": "readOnly"}),
+					prometheus.IncludeOnlyWhenLabelMatchFilter(map[string]string{
+						"request_kind": "readOnly",
+					}),
 				),
 				Type: sdkMetric.GAUGE,
 			},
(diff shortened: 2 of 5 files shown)


src/metric/definition.go
```diff
@@ -206,6 +215,7 @@ var ControllerManagerSpecs = definition.SpecGroups{
 				ValueFunc: prometheus.FromValueWithOverriddenName(
 					"node_collector_evictions_total",
 					"nodeCollectorEvictionsDelta",
+					prometheus.IgnoreLabelsFilter("zone"),
```
gsanchezgavier, line 218: why this?

src/metric/definition.go
```diff
@@ -49,11 +49,20 @@ var APIServerSpecs = definition.SpecGroups{
...
 			},
 			{
-				Name: "apiserverCurrentInflightRequests",
+				Name: "apiserverCurrentInflightRequestsMutating",
 				ValueFunc: prometheus.FromValueWithOverriddenName(
 					"apiserver_current_inflight_requests",
-					"apiserverCurrentInflightRequests",
-					prometheus.IncludeOnlyLabelsFilter("request_kind"),
+					"apiserverCurrentInflightRequestsMutating",
+					prometheus.IncludeOnlyWhenLabelMatchFilter(map[string]string{"request_kind": "mutating"}),
```
gsanchezgavier, line 56: what would happen if another label is present?

marcsanmi: I added a test case [here](https://github.com/newrelic/nri-kubernetes/pull/507/files#diff-feb770271b232b58c3cdf2f08592a3df14a22d3d2d9bd94f86255699a8071e86R878). I'll be more exhaustive and a a specific case with multiple labels.

src/prometheus/definition.go
```diff
@@ -314,6 +314,28 @@ func IncludeOnlyLabelsFilter(labelsToInclude ...string) func(Labels) Labels {
...
+			include := false
+			for labelToInclude, valueToInclude := range labelsToInclude {
+				if label == labelToInclude && value == valueToInclude {
+					include = true
+					break
+				}
+			}
+
+			if include {
+				filteredLabels[label] = value
```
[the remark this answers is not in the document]
marcsanmi: Yes, I completely copy & paste the logic of `IncludeOnlyLabelsFilter`, thanks for the suggestion 👍

src/metric/definition.go
marcsanmi: I wanted to simply skip this label as it can have high cardinality...

src/metric/definition.go
marcsanmi: Added more tests here ffc5a85dd29850f2d6f3aa1b1c75cf39fbb2993c.

src/prometheus/definition.go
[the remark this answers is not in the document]
marcsanmi: Updated ffc5a85dd29850f2d6f3aa1b1c75cf39fbb2993c.

src/controlplane/controlplane_test.go
```diff
@@ -109,7 +109,13 @@ func Test_Scraper_Autodiscover_all_cp_components(t *testing.T) {
 
 			// apiserverStorageObjects replaces etcObjectCounts in k8s versions above 1.23
 			if testutil.IsBelow(version, testutil.Testdata123) {
-				versionAsserter = versionAsserter.Excluding(exclude.Metrics("apiserverStorageObjects"))
```
gsanchezgavier, line 112: i'm wondering why this test didn't catch the metric name "issue" we found with the e2e for some metrics like the `apiserver_current_inflight_requests`

src/metric/definition.go
gsanchezgavier: is this a new metric? if it is , it might have sense to ignore all the labels so the metric doesn't get renamed under any condition.

src/prometheus/definition.go
```diff
@@ -540,6 +558,82 @@ func FromSummary(key string) definition.FetchFunc {
 	}
 }
 
+// FromValueWithLabelsFilter creates a FetchFunc that fetches values from prometheus metrics values given specific
+// labels filter.
+func FromValueWithLabelsFilter(metricName string, nameOverride string, labelsFilter ...LabelsFilter) definition.FetchFunc {
```
gsanchezgavier, line 563: is `nameOverride` needed here? It might be but since we are already defining the metric name in [Name](https://github.com/newrelic/nri-kubernetes/pull/507/files#diff-a0716046905e3ee93ccaf0bc65ae874ed8d9edda79c4d8a3bc394d57bad9bbe6R63) and also we are not planning to add any suffix it confuses me.

src/prometheus/definition.go
```diff
@@ -540,6 +558,82 @@ func FromSummary(key string) definition.FetchFunc {
...
+		}
+		return nil, fmt.Errorf(
+			"incompatible metric type for %s. Expected: Metric or []Metric. Got: %T",
+			metricName,
+			value,
+		)
+	}
+}
+
+func fetchedValuesFromRawMetricsWithoutSuffix(
```
gsanchezgavier, line 584: perhaps we could call this `fetchedValuesFromRawMetrics` and change the other something like `fetchedValuesFromRawMetricsAddingSuffix` to make things more clear WDYT

gsanchezgavier: also perhaps we could extract some of the shared logic into functions that both could share

src/prometheus/definition.go
marcsanmi: AFAIK, it seems necessary. `metricName` in a form like `apiserver_current_inflight_requests` is used to fetch the metric from raw metrics but then we need to override metric name as we expect to have it in a form like `apiserverCurrentInflightRequestsMutating`. Does it make sense to you?

src/prometheus/definition.go
marcsanmi: I agree, updated it here 97e8b42149397faa07c96010501130a1bf0c1906.

src/metric/definition.go
marcsanmi: For now I just leave it explicitly excluding the single label it was.. It's not clear (I'd say not possible) how to exclude all the labels for a metric.

gsanchezgavier: no sorry, 🤦 this was what we had before , jes.... this methods . Perhaps a comment there for the future us would be good in this case

src/prometheus/definition.go
gsanchezgavier: could you add an example on the description, I think that will help us to understand what the function does.

gsanchezgavier (approved):
LGTM

PR #517, opened 2022-07-18
fix(trivy): reduce trivy noise

paologallinaharbur:
Following the same approach of https://github.com/newrelic/infrastructure-bundle/pull/207

For example now we are pointing to the latest alpine version available, but there is still a dependency marked as HIGH vulnerability.

We could fix that manually, but the approach is not scalable for all vulnerabilities marked as HIGH.

Regarding the go-restful vulnerability it is a false positive in 2.16.0, trivy  is no noticing that the backport fixed it   -> https://github.com/emicklei/go-restful/pull/503


sigilioso (approved):
👍 Thanks for taking care of this

paologallinaharbur:
Thank to @sigilioso for pointing out the existence of trivyignore

PR #522, opened 2022-07-25
[helm chart] Patch job should also honor the global service account flag

### LINKED ISSUE #514: controlplane serviceAccount not-matching name when custom is used
[NOTE]: # ( ^^ Provide a general summary of the issue in the title above. ^^ )

## Description
[NOTE]: # ( Describe the problem you're encountering. )
[TIP]:  # ( Do NOT share sensitive information, whether personal, proprietary, or otherwise! )
If custom Service Account is used like this:
```yaml
serviceAccount:
  annotations: {}
  create: false
  name: "my-service-account"
```
The controlplane [Daemonset](https://github.com/newrelic/nri-kubernetes/blob/v3.4.0/charts/newrelic-infrastructure/templates/controlplane/daemonset.yaml#L62) will try to look for `my-service-account-controlplane`, while the rest will work[ pick the correct one](https://github.com/newrelic/nri-kubernetes/blob/v3.4.0/charts/newrelic-infrastructure/templates/kubelet/daemonset.yaml#L52).

## Expected Behavior
[NOTE]: # ( Tell us what you expected to happen. )
Use the SA name configured.
## [Troubleshooting](https://discuss.newrelic.com/t/troubleshooting-frameworks/108787) or [NR Diag](https://docs.newrelic.com/docs/using-new-relic/cross-product-functions/troubleshooting/new-relic-diagnostics) results
[NOTE]: # ( Provide any other relevant log data. )
[TIP]:  # ( Scrub logs and diagnostic information for sensitive information )

## Steps to Reproduce
[NOTE]: # ( Please be as specific as possible. )
[TIP]:  # ( Link a sample application that demonstrates the issue. )

## Your Environment
[TIP]:  # ( Include as many relevant details about your environment as possible including the running version of New Relic software and any relevant configurations. )

## Additional context
[TIP]:  # ( Add any other context about the problem here. For example, relevant community posts or support tickets. )

kang-makes:
There is an issue open by GTS saying that not all the charts honor the flag `.global.serviceAccount.create` and `.global.serviceAccount.name`: newrelic/helm-charts#875

There is not also the issue that they are not honored by the deployments but also by the jobs that create and patch webhook certificates.

Fixes #514

### DIFF
--- a/charts/newrelic-infrastructure/Chart.yaml
+++ b/charts/newrelic-infrastructure/Chart.yaml
@@ -8,7 +8,7 @@ sources:
   - https://github.com/newrelic/nri-kubernetes/tree/master/charts/newrelic-infrastructure
   - https://github.com/newrelic/infrastructure-agent/
 
-version: 3.7.2
+version: 3.7.3
 appVersion: 3.4.0
 
 dependencies:
--- a/charts/newrelic-infrastructure/Chart.yaml
+++ b/charts/newrelic-infrastructure/Chart.yaml
@@ -8,7 +8,7 @@ sources:
   - https://github.com/newrelic/nri-kubernetes/tree/master/charts/newrelic-infrastructure
   - https://github.com/newrelic/infrastructure-agent/
 
-version: 3.7.1
+version: 3.7.2
 appVersion: 3.4.0
 
 dependencies:
--- a/charts/newrelic-infrastructure/templates/_helpers_compatibility.tpl
+++ b/charts/newrelic-infrastructure/templates/_helpers_compatibility.tpl
@@ -64,8 +64,8 @@ Returns legacy annotations if available
 Returns agent configmap merged with legacy config and legacy eventQueueDepth config
 */}}
 {{- define "newrelic.compatibility.agentConfig" -}}
-{{- $oldConfig := .Values.config | default dict -}}
-{{- $newConfig := .Values.common.agentConfig  -}}
+{{- $oldConfig := deepCopy (.Values.config | default dict) -}}
+{{- $newConfig := deepCopy .Values.common.agentConfig -}}
 {{- $eventQueueDepth := dict -}}
 
 {{- if .Values.eventQueueDepth -}}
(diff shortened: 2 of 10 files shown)


paologallinaharbur:
This is closing https://github.com/newrelic/nri-kubernetes/pull/513 as well, right?

charts/newrelic-infrastructure/Chart.yaml
```diff
@@ -8,7 +8,7 @@ sources:
   - https://github.com/newrelic/nri-kubernetes/tree/master/charts/newrelic-infrastructure
   - https://github.com/newrelic/infrastructure-agent/
 
-version: 3.7.1
+version: 3.7.3
```
paologallinaharbur, line 11: this could be 3.7.2

paologallinaharbur (approved):
Thanks for taking care of this! 🚀

paologallinaharbur:
Just a doubt, in the old PR we were changing also the default:
```
Having the create: true on the default values of the chart was preventing the global values to overwrite this config
```
would make sense to include the test and the change here as well?

kang-makes:
> would make sense to include the test and the change here as well?

It is already tested in the common library: https://github.com/newrelic/helm-charts/blob/a0b1859/library/CHART-TEMPLATE/tests/cl-serviceaccount_create_test.yaml#L77-L89

In fact, this is intended behavior. It was us that did not change this in the values when we ported this chart to use the `common-library`.

charts/newrelic-infrastructure/Chart.yaml
kang-makes: I overlooked this comment and I merged it with 3.7.3, sorry.

paologallinaharbur:
I am not following, how the user can use the global if in [the local values](https://github.com/newrelic/nri-kubernetes/blob/ee737ac0e1b047d575244c48f7d071b24982428c/charts/newrelic-infrastructure/values.yaml#L452) there is already true? 

```
      global:
        serviceAccount:
          create: false
```

is ignored with a default value in serviceAccount.create, that is what @gsanchezgavier was doing in https://github.com/newrelic/nri-kubernetes/pull/513

kang-makes:
Yeah, you are right. I misread you and did not cover the whole issue.

I am owning the PR from Guillermo right now. Sorry.

PR #523, opened 2022-07-25
[helm chart] `deepCopy` values when we are going to do a `mergeOverwrite`

kang-makes:
This is the only chart remaining to fix the issue newrelic/helm-charts#858

Go to the issues linked for having all the context.

TL;DR: `mergeOverwrite` does not only return the resulting dict but also modifies the first parameter so we have to `deepCopy`  it to not change values for the rest of the helpers/templates.

Fixes newrelic/helm-charts#858

### DIFF
--- a/charts/newrelic-infrastructure/Chart.yaml
+++ b/charts/newrelic-infrastructure/Chart.yaml
@@ -8,7 +8,7 @@ sources:
   - https://github.com/newrelic/nri-kubernetes/tree/master/charts/newrelic-infrastructure
   - https://github.com/newrelic/infrastructure-agent/
 
-version: 3.7.2
+version: 3.7.3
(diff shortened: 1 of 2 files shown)


paologallinaharbur (approved):

paologallinaharbur:
something broke the e2e, but it should be not related 🤔

paologallinaharbur:
would make sense to unify the two PRs to avoid two releases and to add two changelogs?

kang-makes:
> would make sense to unify the two PRs to avoid two releases and to add two changelogs?

We could. I wanted to have the PRs separated as they are two different issues to make it easy to review. We can merge it to the first PR whenever you want.

paologallinaharbur (approved):
let's merge it then 😄

PR #532, opened 2022-08-18
Allow kind to be set for kubelet

matharoo:
In our use case we are trying to setup external RDS monitoring using this infrastructure agent. We do not want the kubelet running as dameonset because it sets up pods on all nodes. To save resources on our cluster we want to be able to deploy it as deployment so that we can control the replicas. 

Currently only controlplane allows this, we would like the kubelet to be able to that as well.

### DIFF
--- a/charts/newrelic-infrastructure/templates/NOTES.txt
+++ b/charts/newrelic-infrastructure/templates/NOTES.txt
@@ -1,3 +1,4 @@
+{{- if .Values.kubelet.enabled }}
 {{- if not .Values.forceUnsupportedInterval }}
 {{- $max := 40 }}
 {{- $min := 10 }}
@@ -9,6 +10,7 @@
 {{- end }}
 {{- if lt ( .Values.common.config.interval | trimSuffix "s" | int64 ) $min }}
 {{ fail (printf "Intervals smaller than %ds are not supported" $min) }}
+{{- end }}
 {{- end }}
 {{- end }}
 
--- a/charts/newrelic-infrastructure/templates/deployment/_affinity_helper.tpl
+++ b/charts/newrelic-infrastructure/templates/deployment/_affinity_helper.tpl
@@ -1,0 +1,33 @@
+{{- /*
+Patch to add affinity in case we are running in fargate mode
+*/ -}}
+{{- define "nriKubernetes.deployment.affinity.fargateDefaults" -}}
+nodeAffinity:
+  requiredDuringSchedulingIgnoredDuringExecution:
+    nodeSelectorTerms:
+      - matchExpressions:
+          - key: eks.amazonaws.com/compute-type
+            operator: NotIn
+            values:
+              - fargate
+{{- end -}}
+
+
+
+{{- /*
+As this chart deploys what it should be three charts to maintain the transition to v3 as smooth as possible.
+This means that this chart has 3 affinity so a helper should be done per scraper.
+*/ -}}
+{{- define "nriKubernetes.deployment.affinity" -}}
+
+{{- if or .Values.deployment.affinity .Values.nodeAffinity  -}}
+    {{- $legacyNodeAffinity := fromYaml ( include "newrelic.compatibility.nodeAffinity" . ) | default dict -}}
+    {{- $valuesAffinity := .Values.deployment.affinity | default dict -}}
+    {{- $affinity := mustMergeOverwrite $legacyNodeAffinity $valuesAffinity -}}
+    {{- toYaml $affinity -}}
+{{- else if include "newrelic.common.affinity" . -}}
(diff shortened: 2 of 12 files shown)


roobre:
Hi @matharoo,

Unfortunately simply switching this to a `Deployment` is not something that can be trivially done, as it might have unforeseen consequences. Monitoring external workloads is not really considered within the use case of this chart, which has baked-in defaults assuming that:
1. The node where this pod is running can and should be monitored
2. All integrations (including databases) monitored by a pod are running in the same node where said pod is running

These two assumptions are rooted strongly in the architecture of the integration, and are used when generating unique IDs, computing which metadata should be mirrored from the host to the database, etc. Those are the reasons because of which the Kind of the kubelet scraper is not configurable: Doing so would conflict with these assumptions and is definitely outside of our test radar.

I would suggest a custom deployment of the https://github.com/newrelic/infrastructure-bundle image for this, setting the env var `NRIA_FORWARD_ONLY` to `1` to instruct the infrastructure agent to only run integrations and not monitor the host.

matharoo:
> Hi @matharoo,
> 
> Unfortunately simply switching this to a `Deployment` is not something that can be trivially done, as it might have unforeseen consequences. Monitoring external workloads is not really considered within the use case of this chart, which has baked-in defaults assuming that:
> 
> 1. The node where this pod is running can and should be monitored
> 2. All integrations (including databases) monitored by a pod are running in the same node where said pod is running
> 
> These two assumptions are rooted strongly in the architecture of the integration, and are used when generating unique IDs, computing which metadata should be mirrored from the host to the database, etc. Those are the reasons because of which the Kind of the kubelet scraper is not configurable: Doing so would conflict with these assumptions and is definitely outside of our test radar.
> 
> I would suggest a custom deployment of the https://github.com/newrelic/infrastructure-bundle image for this, setting the env var `NRIA_FORWARD_ONLY` to `1` to instruct the infrastructure agent to only run integrations and not monitor the host.

thanks for the information, i understand that setting it to deployment may have unforeseen consequences. I have already disabled the cluster and other metrics by setting `is_forward_only: true`.

However with this PR change we can give option to the user to be able to set the `kind` to `deployment`.  In the use-case of external rds integration only one pod is needed which is possible with `Deployment`. We do not need the pod to run on every node as done by `Daemonset`. I have tested the change to deploy it as a Deployment and it works normally and we are able to send the metrics via integration external RDS. 

This change gives the user more flexibility while retaining the original functionality and should be safe to merge.

roobre:
Hi @matharoo,

I'm happy to hear that this is working well for your use case. However, our concern is so that might not be the case for other users. For example, you are using a static (external) target for your RDS database. If a customer were to point the agent to an in-cluster target, as our documentation suggests and covers, they won't be seeing the expected results. This is because, as I mentioned, discovery works talking to the local node/kubelet only.

It would be possible to add this option together with a large warning next to it, saying something similar to "please do not change this unless you are only using this specific feature and your don't really care about these other two". In my experience, any program option belongs to a continuum that ranges between innocuous, dangerous, and a footgun. It is hard to objectively place options in this continuum, but my opinion on this particular one is that switching the Kubelet agent to Deployment is definitely between dangerous and footgun.

For this reason our team is not inclined to expose this as an option to the general public, as we believe the corner case it covers it is not enough to justify the confusion and support burden it could bring. I understand this is a subjective decision and I'm sorry for leaving your use case out, specially considering that Helm makes very hard to modify things if the upstream (us) doesn't specifically add a toggle for it.

I still believe that the best course forward for your use case is to create a custom deployment with the `infrastructure-bundle` image, as you are toggling off most of the features of `nri-bundle` anyway.

matharoo:
> 

Hey @roobre , I completely understand concerns of your team for our specific use case and support burden. I have reverted my kubelet changes for kind. However as you suggested I have added a separate `Deployment` to handle our external RDS monitor, i have only kept agent container in it required for fetching and sending the metrics over to NR, and deployment creates its own separate configuration so nothing impacts the existing functionality. The changes seem to work fine for our use case as we are able to have multiple external RDS monitoring agents running on one cluster, without taking up too much resources as it was happening with Daemonsets.

roobre:
Hey, @matharoo

Happy to see this is working for you. Unfortunately I still think the fundamental issue is the same: Exposing this feature would carry a support and maintenance burden that right now we do not think it is worth the trouble. One more time I recommend that you set up a separate (meaning, out of this chart) deployment for this task.

Again, I'm sorry this does not fit our vision for this chart. I'll be closing this now.

matharoo:
> 

ok yeah no problem, we will do a custom chart. thanks for looking into it though. much appreciated. 
😞

roobre:
@matharoo You are welcome, please let me know if you struggle creating the custom deployment. I'll be happy to help :)

PR #536, opened 2022-08-29
chore(v2): updated dependencies and infrastructure-bundle

paologallinaharbur:
Updated dependencies and infrastructure bundle

Then we will update the chart in V2 Branch and the operator default config


gsanchezgavier (approved):
thanks

gsanchezgavier (approved):
thanks

PR #552, opened 2022-09-13
update pixie check default

bpschmitt:
The flex-based Pixie health check config is generating errors and needs to be debugged.  Until I have time to investigate, I think this should be disabled by default so the `pixieHealthCheck` events aren't spammed with useless data.


bpschmitt:
Closing and will redo.

PR #553, opened 2022-09-13
set pixie check to false

bpschmitt:
Disables the flex-based `pixieHealthCheck` events by default.  There is an error in the flex script syntax that needs to be debugged and it's currently creating noise.  Will re-enable once I get time to fix it.


kang-makes:
We do not know why tests are failing.

We think that it was somehow related to the incidents that GitHub has been having these previous days and somehow it has a corrupted version of the e2e action. It does not make sense the errors you are getting because that action has been 8 months without changes.

I reopened the PR, new branch, new commit, same change... And now tests are passing: https://github.com/newrelic/nri-kubernetes/actions/runs/3091166383/jobs/5000904048

I am closing this PR in favor of #556

PR #556, opened 2022-09-20
set pixie check to false

kang-makes:
@bpschmitt originally posted in PR #553 this:

> Disables the flex-based `pixieHealthCheck` events by default.  There is an error in the flex script syntax that needs to be debugged and it's currently creating noise.  Will re-enable once I get time to fix it.

He opened this PR while GitHub was having an incident and seems to have cached a glitched version of the code/action because it is failing with an action that has has no changes since 8 months ago.

We are recreating manually the PR to test it again.


carlossscastro (approved):
👍

PR #565, opened 2022-10-14
bump deps v2

gsanchezgavier:
- Upgrade infrastructure-bundle to 2.8.32
    - See https://github.com/newrelic/infrastructure-bundle/releases/tag/2.8.32 for more details about the upgraded integrations in this release of the infrastructure-bundle
- Updated dependencies


Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG MODE=normal
-ARG BASE_IMAGE=newrelic/infrastructure-bundle:2.8.28
+ARG BASE_IMAGE=newrelic/infrastructure-bundle:2.8.32
```
kang-makes, line 2: This adds to this integration the issue of `debug` being moved to `trace`.

Nothing that we can fix but a thing to have in mind with other open issues about the `v2` version...

charts/newrelic-infrastructure/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 description: "DEPRECATED: (In maintenance mode) Helm chart for deploying New Relic Infrastructure Agent v2"
 name: newrelic-infrastructure
-version: 2.10.5
-appVersion: 2.13.2
+version: 2.10.6
+appVersion: 2.13.3
```
kang-makes, line 5: If we are bumping the chart also, change the `verboseLog` helper to set `trace` to the agent.

paologallinaharbur, line 5: we are not bundling the last one, is it intended?

paologallinaharbur (approved):

charts/newrelic-infrastructure/Chart.yaml
gsanchezgavier: no this is an error i will correct , i just bumped one patch assuming it was correct

kang-makes (approved):
The fix for the chart is a NIT as the chart is in maintenance mode.

Thanks for taking care!

Let's wait for the tests to finish before merging tho :)

PR #570, opened 2022-11-02
chore: update testing certificates

sigilioso:
Certificates to test the control-plane authenticator are updated, those certificates are only used for testing purposes.


CHANGELOG.md
```diff
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## 3.5.1
```
paologallinaharbur, line 8: do we need a new release?

sigilioso: Actually we don't, it can perfectly wait for the next one. I included it because I was releasing before. Thanks!

sigilioso: Reverted here: [4f330d0](https://github.com/newrelic/nri-kubernetes/pull/570/commits/4f330d0714f2de0937805afb3127099559bc2a3e)

src/controlplane/client/authenticator/README.md
```diff
@@ -0,0 +1,104 @@
+# Certificates generation for authenticator tests
+
+Testing certificates where generated using [cfssl](https://github.com/cloudflare/cfssl) tool.
+
+## Steps to generate new certificates
+
+Install `cfssl` and generate configuration files a follows:
```
marcsanmi, line 7:
```suggestion
Install `cfssl` and generate configuration files as follows:
```

marcsanmi (approved):
LGTM 👍

kang-makes (approved):
I would love to make the certificates directly with OpenSSL instead of relying on more tooling. But as certs are going to last for 100 years luckily we will not need to rotate them ;)

Thanks for taking care. LGTM.

PR #571, opened 2022-11-02
chore: bump chart version and dependencies

sigilioso:
We still need to update `newrelic/infrastructure-bundle` when [this PR](https://github.com/newrelic/infrastructure-bundle/pull/241) is merged.


kang-makes (approved):
I would prefer to renovate to do it because automation is more reliable but I think that you want to make things faster... So be it.

LGTM.

sigilioso:
> I would prefer to renovate to do it because automation is more reliable but I think that you want to make things faster... So be it.
> 
> LGTM.

Thank you!

PR #577, opened 2022-11-17
NR-43831: ksm v2 support

sigilioso:
## kube-state-metrics v2 support

Specific changes:

* Update metric definitions to fech `horizontalpodautoescaler` metrics, as [those where renamed in v2](https://github.com/kubernetes/kube-state-metrics/blob/master/CHANGELOG.md?plain=1#L200).
* Update metric definitions to fetch `kube_daemonset_status_updated_number_scheduled` which [was also renamed in v2](https://github.com/kubernetes/kube-state-metrics/blob/master/CHANGELOG.md?plain=1#L170).
* Extend definition capabilities to keep the current `hpa` event names even if we needed to rename the label group from `hpa` to `horizontalpodautoscaler`.
    - Extend spec definition and populator to allow a custom metric-set-type guessing function (so it is possible to define a new one to use a custom group instead of the label group to compose the event name.
    - Implement a type generator where the group can be customized.
    - Extend the asserter to set up label group aliases.
 * Include additional customizable values in `datagen.sh`.
 * Bump kube-state-metric dependency in `e2e` chart to the latest version.
 * Update the static-data used in integration tests so it is generated with ksm v2.
 * Remove static data for k8s versions older than 1.19.

There are changes in a lot of files, but most of them are part of the static-data updates.


charts/internal/e2e-resources/Chart.yaml
```diff
@@ -1,12 +1,12 @@
...
+version: 1.3.0-devel
 description: This chart creates e2e resources for nri-kubernetes.
 name: e2e-resources
 
 dependencies:
   - name: kube-state-metrics
-    version: 2.13.2
-    repository: https://kubernetes.github.io/kube-state-metrics
+    version: 4.23.0
+    repository: https://prometheus-community.github.io/helm-charts
```
marcsanmi, line 9: Since we are making all these changes in this PR, what about updating the Helm [setup](https://github.com/newrelic/nri-kubernetes/blob/d442f74853b9f23512e7b0625dfea51955393c5e/.github/workflows/e2e.yaml#L65) repo for the e2e here? 🤔

sigilioso: Updated here: [cac66c0](https://github.com/newrelic/nri-kubernetes/pull/577/commits/cac66c05853994b8e50ac0ce04dc3d1ebe757119)

src/definition/populate.go
```diff
@@ -116,7 +116,12 @@ func IntegrationPopulator(config *IntegrationPopulateConfig) (bool, []error) {
 				extraAttributes...,
 			)
 
-			msType, err := config.MsTypeGuesser(config.ClusterName, groupLabel, entityID, config.Groups)
+			msTypeGuesser := config.MsTypeGuesser
+			if customGuesser := config.Specs[groupLabel].MsTypeGuesser; customGuesser != nil {
+				msTypeGuesser = customGuesser
```
marcsanmi, line 121: Are we testing this overriding behaviour? Not separately but integrated.

marcsanmi (commented):
Good job! I left some comments.

internal/testutil/datagen/README.md
```diff
@@ -31,7 +31,8 @@ After this, it will deploy KSM and programmatically hit all the endpoints requir
   - etcd `/metrics`
   - scheduler `/metrics`
 
-It will do this by spawning a privileged, `hostNetwork` `alpine:latest` pod in the cluster and running itself from inside.
+It will do this by spawning a privileged, `hostNetwork` `alpine:latest` pod in the cluster and running itself from inside,
```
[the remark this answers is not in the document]
sigilioso: Good catch, fixed here: [af38ba7](https://github.com/newrelic/nri-kubernetes/pull/577/commits/af38ba77d94fb06e6747ccff83aa5807b634bf96)

internal/testutil/asserter/assert.go
```diff
@@ -140,14 +154,25 @@ func (a *Asserter) shouldExcludeGroup(group string) bool {
...
+		if specGroupNameMatch(e, pseudotype) {
 			appropriateEntities = append(appropriateEntities, e)
 		}
 	}
 
 	return appropriateEntities
 }
 
+// pseudotypeMatch returns true if the specGroupName match with the provided entity.
+func specGroupNameMatch(entity *integration.Entity, specGroupName string) bool {
```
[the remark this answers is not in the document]
sigilioso: I wanted to move the "heuristic" logic to that function, so `entitiesFor` just keeps the entities when `specGroupNameMatch`. So, If we changed the criteria again we would just need to change this new function. That is the reason why it receives the whole entity instead of directly the namespace.

WDYT?

src/definition/populate.go
sigilioso: That is a good point 👍 
We are definitely testing it with the integration tests: if we didn't use it current _horizontalpodautoscaler_ wouldn't be found in `ksm_test`. However, I think it should be possible to unit-test this new behavior by extending the populator tests, so I am giving it a try. Thanks!

marcsanmi: I used the word `integrated` unfortunately here. I didn't mean in the integration tests but in the unit 😅

internal/testutil/asserter/assert.go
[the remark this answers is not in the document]
marcsanmi: I see you point, seems good to me 👍

src/definition/populate.go
sigilioso: I added a new populator test here to check the custom msTypeGuesser is actually used. [d50f349](https://github.com/newrelic/nri-kubernetes/pull/577/commits/d50f349b403eaa4af1b25178978d150004e13e72)

paologallinaharbur:
Next time I would try to split similar PRs in two to make reviewing easier :)

charts/internal/e2e-resources/Chart.yaml
```diff
@@ -1,22 +1,26 @@
...
-version: 1.2.0-devel
+version: 1.3.0-devel
 description: This chart creates e2e resources for nri-kubernetes.
 name: e2e-resources
 
 dependencies:
   - name: kube-state-metrics
-    version: 2.13.2
-    repository: https://kubernetes.github.io/kube-state-metrics
+    version: 4.23.0
```
paologallinaharbur, line 8: would make sense to point to latest instead of a specific version?

paologallinaharbur: We could pin either the minor or the major 🤔

internal/testutil/testutil.go
```diff
@@ -29,9 +29,6 @@ func (v Version) K8s() (K8s, error) {
 // - REMEMBER TO ADD IT TO AllVersions() BELOW.
 
 const (
-	Testdata116 = "1_16"
-	Testdata117 = "1_17"
```
paologallinaharbur, line 33: since we are still supporting those versions why do we need to remove the test data?

src/definition/guessfunc.go
```diff
@@ -0,0 +1,15 @@
+package definition
+
+import (
+	"fmt"
+	"strings"
+)
+
+// K8sMetricSetTypeGuesser is the metric set type guesser for k8s integrations.
```
paologallinaharbur, line 8: since the behavior is not straightforward can we add a few more lines?

src/definition/guessfunc.go
```diff
@@ -0,0 +1,15 @@
...
+func K8sMetricSetTypeGuesser(_, groupLabel, _ string, _ RawGroups) (string, error) {
```
paologallinaharbur, line 9: by the way as far as I understand the weird signature is due to:
```
type GuessFunc func(clusterName, groupLabel, entityID string, groups RawGroups) (string, error)
```

Could we change that to have just the parameter used?

src/definition/populate.go
```diff
@@ -116,7 +116,12 @@ func IntegrationPopulator(config *IntegrationPopulateConfig) (bool, []error) {
 				extraAttributes...,
 			)
 
-			msType, err := config.MsTypeGuesser(config.ClusterName, groupLabel, entityID, config.Groups)
+			msTypeGuesser := config.MsTypeGuesser
+			if customGuesser := config.Specs[groupLabel].MsTypeGuesser; customGuesser != nil {
```
paologallinaharbur, line 120: I still do not understand why instead of having `config.Specs[groupLabel].MsTypeGuesser` we do not have `config.Specs[groupLabel].MsType` directly 😅 don't you have the feeling that there is too much code to decide a sample name?

src/prometheus/definition.go
```diff
@@ -35,40 +40,54 @@ var FromRawEntityIDGenerator = func(_, rawEntityID string, _ definition.RawGroup
...
+	return func(groupLabel string, rawEntityID string, g definition.RawGroups, clusterName string) (string, error) {
+		return defaultEntityTypeFromLabelValue(key, group, groupLabel, rawEntityID, g, clusterName)
+	}
+}
 
-			if namespace == "" {
-				return "", fmt.Errorf("empty namespace for generated entity type for %q", groupLabel)
-			}
-			return fmt.Sprintf("k8s:%s:%s:%s", clusterName, namespace, groupLabel), nil
+func defaultEntityTypeFromLabelValue( //nolint: cyclop
```
paologallinaharbur, line 55: What I find a bit confusing is that we have group and groupLabel. In one function those are the same since group and groupLabel are the same, in the second one they have different values.

Could we leverage an override concept?

paologallinaharbur (approved):
The issues are linked to the way the code used to be. I would try to refactor a little more, maybe hardcoding same values and removing unnecessary abstraction

paologallinaharbur:
I opened a small PR https://github.com/newrelic/nri-kubernetes/pull/581 to show what I mean

src/definition/populate_test.go
```diff
@@ -734,6 +734,38 @@ func TestIntegrationPopulator_PrometheusFormatFilterNamespace(t *testing.T) {
...
+
+	populated, errs := definition.IntegrationPopulator(config)
+	assert.True(t, populated)
+	assert.Empty(t, errs)
+	for _, e := range intgr.Entities {
+		if e.Metadata.Namespace != "k8s:cluster" { // the custom MsTypeGuesser does not apply to the cluster entity.
+			for _, ms := range e.Metrics {
+				assert.Equal(t, "Custom", ms.Metrics["event_type"])
+			}
+		}
```
marcsanmi, line 765: NIT: I like avoiding nested code in cases like this as IMHO it makes it easier to follow the code.
```suggestion
		if e.Metadata.Namespace == "k8s:cluster" { // the custom MsTypeGuesser does not apply to the cluster entity.
		    continue
		}
		
		for _, ms := range e.Metrics {
		    assert.Equal(t, "Custom", ms.Metrics["event_type"])
		}
```

charts/internal/e2e-resources/Chart.yaml
sigilioso: Good point! As it is a testing chart I think we can pin the major.

internal/testutil/testutil.go
sigilioso: I had issues to launch local kubernetes clusters for those versions using a modern minikube in arm. We could dig into it or we could keep the data we had and add exceptions for the _horizontalpodautoscaler_ metrics, but I am not sure if maintaining the static data for those versions (which are [not supported by kubernetes](https://kubernetes.io/releases/patch-releases/#non-active-branch-history)) is useful.

src/definition/guessfunc.go
sigilioso: I simply moved this function (it already existed) to avoid a circular dependency, but you are totally right, it is quite confusing and I will take the chance to extend the comment.

src/definition/guessfunc.go
```diff
@@ -0,0 +1,15 @@
...
+	var sampleName string
+	for _, s := range strings.Split(groupLabel, "-") {
+		sampleName += strings.Title(s) //nolint: staticcheck // TODO: use golang.org/x/text/cases assuring no breaking changes.
```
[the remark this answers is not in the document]
sigilioso: I don't think so. As I see it, it would require a specific task. It shouldn't be really complex (if both integration and e2e tests pass we should be fine) but I didn't want to do it as part of this PR to avoid making its scope even wider.

src/definition/guessfunc.go
sigilioso: You actually addressed it in #581, thanks for simplifying 🚀

src/definition/populate_test.go
sigilioso: Thanks! Done: [c423d36](https://github.com/newrelic/nri-kubernetes/pull/577/commits/c423d360b45742a1d20edf100aab0043ca2b87d3)

charts/internal/e2e-resources/Chart.yaml
sigilioso: Changed here: [bb86b3f](https://github.com/newrelic/nri-kubernetes/pull/577/commits/bb86b3fa72ac3af74f33c17a4d5afed497e11ccb)

src/definition/guessfunc.go
sigilioso: Updated the comment here: [65c1156](https://github.com/newrelic/nri-kubernetes/pull/577/commits/65c11564ad5aae1b114fdac74f44ba23e493ca88)

src/definition/populate.go
paologallinaharbur: offline but online 😄

src/prometheus/definition.go
sigilioso: I wanted to define a really specific function for horizontalpodautoscaller entity type where it is called, but I couldn't do it because we would have circular dependencies.
On the other hand, some kind of override or having the caller passing both arguments would require changing all the current usages and again, I would like to avoid that kind of refactor in this PR so its scope doesn't get even wider.

marcsanmi (approved):
Thanks for addressing it!

PR #592, opened 2022-12-15
Update hostnetwork configuration for ksm deployment

dpacheconr:
KSM does not required hostnetwork by default, apart from some corner use cases.
Currently if we disable hostnetwork it will also disable to kubelet as hostnetwork is set by global/common variable
This PR makes it configurable, if hostnetwork is set in values it will be used otherwise it will still use global/common variable 
There are various ways to tackle this and you may choose different one, but please consider making ksm hostnetwork independently configurable

---
Context added by @roobre: This would allow to deploy the `kubelet` scraper with `hostNetwork` enabled while keeping it off for the rest. The main motivation for this is to be able to access [AWS Metadata](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html) when it is [limited to 1 hop only](https://docs.aws.amazon.com/config/latest/developerguide/ec2-token-hop-limit-check.html), which is the recommended value from AWS.

### DIFF
--- a/charts/newrelic-infrastructure/Chart.yaml
+++ b/charts/newrelic-infrastructure/Chart.yaml
@@ -8,7 +8,7 @@ sources:
   - https://github.com/newrelic/nri-kubernetes/tree/main/charts/newrelic-infrastructure
   - https://github.com/newrelic/infrastructure-agent/
 
-version: 3.10.1
+version: 3.11.0
 appVersion: 3.6.0
 
 dependencies:
--- a/charts/newrelic-infrastructure/Chart.yaml
+++ b/charts/newrelic-infrastructure/Chart.yaml
@@ -8,7 +8,7 @@ sources:
   - https://github.com/newrelic/nri-kubernetes/tree/main/charts/newrelic-infrastructure
   - https://github.com/newrelic/infrastructure-agent/
 
-version: 3.10.0
+version: 3.10.1
 appVersion: 3.6.0
 
 dependencies:
--- a/charts/newrelic-infrastructure/templates/ksm/_host_network.tpl
+++ b/charts/newrelic-infrastructure/templates/ksm/_host_network.tpl
@@ -1,0 +1,22 @@
+{{/* Returns whether the ksm scraper should run with hostNetwork: true based on the user configuration. */}}
+{{- define "nriKubernetes.ksm.hostNetwork" -}}
+{{- /* `get` will return "" (empty string) if value is not found, and the value otherwise, so we can type-assert with kindIs */ -}}
+{{- if get .Values.ksm "hostNetwork" | kindIs "bool" -}}
+    {{- if .Values.ksm.hostNetwork -}}
+        {{- .Values.ksm.hostNetwork -}}
+    {{- end -}}
+{{- else if include "newrelic.common.hostNetwork" . -}}
+    {{- include "newrelic.common.hostNetwork" . -}}
+{{- end -}}
+{{- end -}}
+
+
+
+{{/* Abstraction of "nriKubernetes.ksm.hostNetwork" that returns true of false directly */}}
+{{- define "nriKubernetes.ksm.hostNetwork.value" -}}
+{{- if include "nriKubernetes.ksm.hostNetwork" . -}}
+    true
+{{- else -}}
+    false
+{{- end -}}
+{{- end -}}
--- a/charts/newrelic-infrastructure/templates/ksm/deployment.yaml
+++ b/charts/newrelic-infrastructure/templates/ksm/deployment.yaml
@@ -50,7 +50,7 @@ spec:
       {{- end }}
       serviceAccountName: {{ include "newrelic.common.serviceAccount.name" . }}
       hostNetwork: {{ include "nriKubernetes.ksm.hostNetwork.value" . }}
-      {{- if include "nriKubernetes.controlPlane.hostNetwork" . }}
+      {{- if include "nriKubernetes.ksm.hostNetwork" . }}
       dnsPolicy: ClusterFirstWithHostNet
       {{- end }}
       
--- a/charts/newrelic-infrastructure/templates/ksm/deployment.yaml
+++ b/charts/newrelic-infrastructure/templates/ksm/deployment.yaml
@@ -49,11 +49,11 @@ spec:
         {{- . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "newrelic.common.serviceAccount.name" . }}
-      hostNetwork: {{ include "newrelic.common.hostNetwork.value" . }}
+      hostNetwork: {{ include "nriKubernetes.ksm.hostNetwork.value" . }}
       {{- if include "nriKubernetes.controlPlane.hostNetwork" . }}
       dnsPolicy: ClusterFirstWithHostNet
       {{- end }}
-
+      
       {{- if .Values.ksm.initContainers }}
       initContainers: {{- tpl (.Values.ksm.initContainers | toYaml) . | nindent 8 }}
       {{- end }}
--- a/charts/newrelic-infrastructure/tests/hostNetwork_test.yaml
+++ b/charts/newrelic-infrastructure/tests/hostNetwork_test.yaml
@@ -157,27 +157,6 @@ tests:
           value: true
         template: templates/kubelet/daemonset.yaml
 
-  - it: hostNetwork is set by the common library
-    set:
-      licenseKey: test
-      cluster: test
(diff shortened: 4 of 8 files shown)


roobre:
Hey @dpacheconr, thanks a lot for your PR!

I'd like to know a bit more about the use case you are facing. You say that:

> Currently if we disable hostnetwork it will also disable to kubelet as hostnetwork is set by global/common variable

While this is correct, I think it is worth noting that `hostNetwork` is `false` by default. This means that, most likely, you do not need to "disable hostnetwork". Is there a use case where you need to turn that on, and where you need to do so only for the kubelet scraper?

dpacheconr:
In this use case, we need kubelet to run as host network and ksm without host network.

If using global or common variables it’s not possible I believe


Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

charts/newrelic-infrastructure/templates/ksm/_host_network.tpl
```diff
@@ -0,0 +1,22 @@
...
+
+
+{{/* Abstraction of "nriKubernetes.ksm.hostNetwork" that returns true of false directly */}}
+{{- define "nriKubernetes.ksm.hostNetwork.value" -}}
+{{- if include "nriKubernetes.ksm.hostNetwork" . -}}
+    true
+{{- else -}}
+    false
+{{- end -}}
+{{- end -}}
```
roobre, line 22: I think this approach is the way to go, implementing the same mechanism as we do for the control plane scraper.

However now it feels like the kubelet scraper is being given some kind of special treatment, since the other two (ksm and cp) has an individual toggle while the kubelet hasn't.
Could you add the equivalent of these to the kubelet as well?

charts/newrelic-infrastructure/templates/ksm/deployment.yaml
```diff
@@ -49,11 +49,11 @@ spec:
         {{- . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "newrelic.common.serviceAccount.name" . }}
-      hostNetwork: {{ include "newrelic.common.hostNetwork.value" . }}
+      hostNetwork: {{ include "nriKubernetes.ksm.hostNetwork.value" . }}
       {{- if include "nriKubernetes.controlPlane.hostNetwork" . }}
       dnsPolicy: ClusterFirstWithHostNet
       {{- end }}
```
roobre, line 55: This is a bit out of the scope of this PR, but a known bug is that `dnsPolicy` is set depending on `nriKubernetes.controlPlane.hostNetwork`, not `nriKubernetes.ksm.hostNetwork`. Do you think you could change that as well?

charts/newrelic-infrastructure/values.yaml
```diff
@@ -583,4 +586,4 @@ fedramp:
 
 # -- (bool) Sets the debug logs to this integration or all integrations if it is set globally. Can be configured also with `global.verboseLog`
 # @default -- `false`
-verboseLog:
+verboseLog:
```
roobre, line 589: Seems like the POSIX-compliant newline got removed somehow. Could we restore it?

roobre (changes requested):
Thanks again for the PR! I think it is looking great. In addition to the inline comments, I think it would be great to:
- Bump the minor of the chart version in `Chart.yaml`
- Add a test for these new cases. This should be really easy as we already have tests for the hostNetwork toggle here: https://github.com/newrelic/nri-kubernetes/blob/main/charts/newrelic-infrastructure/tests/hostNetwork_test.yaml. I think it would be great to cover that individual toggles change individual component's hostNetwork (and not others'), and that they can override the common hostNetwork settings. This may seem dumb but it is a good safeguard against copypaste artifacts.

charts/newrelic-infrastructure/values.yaml
dpacheconr: Done

charts/newrelic-infrastructure/templates/ksm/deployment.yaml
dpacheconr: Done

charts/newrelic-infrastructure/Chart.yaml
```diff
@@ -8,7 +8,7 @@ sources:
   - https://github.com/newrelic/nri-kubernetes/tree/main/charts/newrelic-infrastructure
   - https://github.com/newrelic/infrastructure-agent/
 
-version: 3.10.0
+version: 3.10.1
```
roobre, line 11:
```suggestion
version: 3.11.0
```

charts/newrelic-infrastructure/tests/hostNetwork_test.yaml
```diff
@@ -114,3 +114,130 @@ tests:
...
+          value: true
+        template: templates/ksm/deployment.yaml
+      - equal:
+          path: spec.template.spec.hostNetwork
+          value: true
+        template: templates/controlplane/daemonset.yaml
+      - equal:
+          path: spec.template.spec.hostNetwork
+          value: true
+        template: templates/kubelet/daemonset.yaml
```
roobre, line 179: Seems like this test case is the same as the one that already exists

charts/newrelic-infrastructure/tests/hostNetwork_test.yaml
```diff
@@ -114,3 +114,130 @@ tests:
...
```
roobre, line 242: Also duplicated

charts/newrelic-infrastructure/values.yaml
```diff
@@ -84,6 +84,9 @@ kubelet:
     - operator: "Exists"
       effect: "NoExecute"
   nodeSelector: {}
+  # -- (bool) Sets pod's hostNetwork. Uncomment to configure hostnetwork to ksm deployment only, bypasses global/common variable 
+  # @default ==  `global.hostNetwork`
+  #hostNetwork: false
```
roobre, line 89: I think having the value commented out will cause [helm-docs](https://github.com/norwoodj/helm-docs) to ignore it. Can you verify this?
```suggestion
  # -- (bool) Sets pod's hostNetwork. Uncomment to configure hostnetwork to ksm deployment only, bypasses global/common variable 
  # @default -- Same as `global.hostNetwork`
  hostNetwork:
```

charts/newrelic-infrastructure/values.yaml
```diff
@@ -133,6 +136,9 @@ ksm:
...
```
roobre, line 141: Same as above, I think helm-docs does not understand this syntax.

roobre (commented):
Few more comments, regarding duplicated tests and helm docs comments.

After taking a look at those, could you regenerate the `README.md` file with [helm-docs](https://github.com/norwoodj/helm-docs) and commit it as well? Sorry for not mentioning this earlier!

charts/newrelic-infrastructure/tests/hostNetwork_test.yaml
dpacheconr: well, I did it so we know ksm.hostNetwork: null, but that is the same as more generic one
will deleted

charts/newrelic-infrastructure/tests/hostNetwork_test.yaml
dpacheconr: well, I did it so we know kubelet.hostNetwork: null, but that is the same as more generic one
will deleted

charts/newrelic-infrastructure/values.yaml
dpacheconr: done

charts/newrelic-infrastructure/values.yaml
dpacheconr: done

charts/newrelic-infrastructure/values.yaml
```diff
@@ -84,6 +84,9 @@ kubelet:
     - operator: "Exists"
       effect: "NoExecute"
   nodeSelector: {}
+  # -- (bool) Sets pod's hostNetwork. When set bypasses global/common variable 
+  # @default ==  `false`
+  hostNetwork: false
```
roobre, line 89:
```suggestion
  # @default -- Not set
  hostNetwork:
```

charts/newrelic-infrastructure/values.yaml
```diff
@@ -133,6 +136,9 @@ ksm:
...
```
roobre, line 141:
```suggestion
  # @default -- Not set
  hostNetwork:
```

PR #594, opened 2022-12-27
Add k8s v1.26.0 integration test static data

vuqtran88:
This PR is still WIP. In fact, some existing e2e tests break because of these changes. In this PR, I generate and add a new integration test static data targets a k8s v1.26.0 cluster to the integration test data set. More specifically, the following changes are made:

1. Conditionally check if `autoscaling/v2beta2` or `autoscaling/v2` is available to use in `charts/internal/e2e-resources/templates/hpa.yaml`.
2. `Since autoscaling/v2beta2` is no longer available to use in k8s v1.26.0, `autoscaling/v2` is used instead. This is accomplished by step 1 above. However,  `autoscaling/v2` is only supported by KSM v2.7.0; therefore, it is necessary to  bump the  kube-state-metrics chart version in `charts/internal/e2e-resources/Chart.yaml` to v4.25.0. This change will install KSM v2.7.0 onto the test k8s v1.26.0 cluster to capture HPA related metrics. However, the change in 2. breaks the existing e2e tests against older versions of k8s such as v1.22 or lower.


vuqtran88:
I actually found one of the solutions for the 2nd bullet point in this PR's description. My solution is removing the kube-state-metrics dependency in the `charts/internal/e2e-resources/Chart.yaml` and let the CI manually installs the correct version of the kube-state-metrics  corresponding to the k8s version installed on the test cluster. I have made the change as well as updated the README.md. All the tests passed so far.

roobre (commented):
Looking great so far, thanks for the good work @vuqtran88 !

.github/workflows/e2e.yaml
```diff
@@ -68,6 +68,25 @@ jobs:
...
+        # - The kube-state-metrics chart v4.23.0 directly maps to kube-state-metrics application v2.6.0.
+        run: |
+          MINOR=$(echo "${{ matrix.k8sVersion }}"|sed -e 's_v\([0-9]\+\)\.\([0-9]\+\)\.\([0-9]\+\)_\2_')
+
+          if [ "$MINOR" -le 25 ]; then
+            KSM_CHART_VERSION="--version v4.23.0"
+          fi
+
+          helm install prometheus-community/kube-state-metrics $KSM_CHART_VERSION --generate-name
+
```
kang-makes, line 89: This snippet of code is not being tested in this PR because this PR does not add Kubernetes 1.26 cluster testing.

So tests pass but are not testing the new code (in the e2e).

This is the reason I am reviewing directly PR #595 directly as it does the integration tests and e2e.

kang-makes (commented):
So far so good, but this PR by itself only tests the integration and not the e2e and includes changes for the e2e in it so I am reviewing PR #595 which includes both at the same time.

vuqtran88:
Closes this PR as #595 has already included these changes.

PR #595, opened 2022-12-29
Add K8s v1.26 e2e and integration tests

vuqtran88:
In this PR, I add new integration tests and e2e tests targets a k8s v1.26.0 cluster. More specifically, the following changes are made:

1. Conditionally check if `autoscaling/v2beta2` or `autoscaling/v2` is available to use in `charts/internal/e2e-resources/templates/hpa.yaml`.

2. `Since autoscaling/v2beta2` is no longer available to use in k8s v1.26.0, `autoscaling/v2` is used instead. This is accomplished by step 1 above. However,  `autoscaling/v2` is only supported by KSM v2.7.0; therefore, it is necessary to  bump the  kube-state-metrics chart version in `charts/internal/e2e-resources/Chart.yaml` to v4.25.0. This change will install KSM v2.7.0 onto the test k8s v1.26.0 cluster to capture HPA related metrics. However, this change alone breaks the existing e2e tests against older versions of k8s such as v1.22 or lower. My solution is removing the kube-state-metrics dependency in the charts/internal/e2e-resources/Chart.yaml and let the CI manually installs the correct version of the kube-state-metrics corresponding to the k8s version installed on the test cluster. I have made the change as well as updated the README.md.

3.  Add e2e tests target v1.26


kang-makes (commented):
The PR looks pretty good but I am leaving it without approving it as I am raising an issue that affects a product decision.

kang-makes (approved):

Issue #598, opened 2023-01-04
[nri-bundle] global.nodeSelector has error


Hokwang:
Hi,

**Bug description**
I use nri-bundle chart with below custom values.yaml
```
global:
  nodeSelector:
    node: infra
```

But this affected to newrelic-infrastructure.kubelet
kubelet is daemonset, so it does not use nodeSelector from global. 

**Version of Helm and Kubernetes**
all version

**Which chart?**
nri-bundle chart version 5.0.2

kang-makes:
Hi @Hokwang,

We are still triaging it but in case this issue exists it is in this repository as the chart affected is the infrastructure one.

I'll reach you again after triaging it :)

kang-makes:
Hi @Hokwang,

I am sorry but we are not able to reproduce it.

```bash
helm template test prod-helm-charts/nri-bundle \
  --set global.cluster=test \
  --set global.licenseKey=test \
  --version 5.0.2 \
  -s charts/newrelic-infrastructure/templates/kubelet/daemonset.yaml | \
yq .spec.template.spec.nodeSelector
```
> `null`

I create a test values file.
![Uploading Screenshot 2023-01-09 at 10.59.10.png…]()

```bash
helm template test prod-helm-charts/nri-bundle \
  --set global.cluster=test \
  --set global.licenseKey=test \
  --version 5.0.2 \
  -s charts/newrelic-infrastructure/templates/kubelet/daemonset.yaml \
  -f values-test.yaml | \
yq .spec.template.spec.nodeSelector
```
> `node: selector`

On the other hand, we were not testing this use case in our pipeline so I created PR #599 to add these test and it seems that tests are passing correctly so there is no further tests we can do.

The next best step is to visit [New Relic Support](https://support.newrelic.com/), where you can engage the New Relic Support Community or open a support ticket depending on your support level. The support team is best positioned to assist with your specific needs.

Please provide a link to this GitHub issue when submitting your community post or support ticket.

Thanks!

PR #599, opened 2023-01-09
Add more tests to the chart

kang-makes:
Issue #598 states that the global `nodeSelector` is not templating correctly.

I cannot reproduce it, but I am adding that test to prevent a regression.


paologallinaharbur (approved):
Thanks for adding more tests 😄

kang-makes:
Rebasing before merging...

PR #602, opened 2023-01-19
Bump chart version and dependencies

marcsanmi:
Ping new version of infra bundle that solves some vulns; you can follow the conversation [here](https://newrelic.slack.com/archives/CP2EWSX7C/p1673970141838969). Leveraged and updated some other dependencies.


sigilioso (approved):
LGTM, thanks for taking care of it 👍

PR #603, opened 2023-01-19
matchExpressions values need to be string now

pschrimpf:
Just ran into copying and pasting this example and received `invalid matchExpressions value: false, type: bool`

It has been changed with the PR https://github.com/newrelic/nri-kubernetes/pull/495,


marcsanmi:
Hi @pschrimpf, thanks for creating the PR.

However, the pipeline will fail because of your permissions while running the e2e as it comes from a forked repo. I'll keep this open just to remember us to address this in any other PR.

htroisi:
Thanks for the contribution @pschrimpf! I pulled this change into #675.

PR #607, opened 2023-01-24
Update readme to include starting tilt step

svetlanabrennan:
The local dev steps need to include a step to start tilt.


svetlanabrennan:
Disregard PR. `make local-env-start` starts up tilt.

PR #609, opened 2023-01-26
Add CronJob KSM metrics.

htroisi:
### Summary

The Kubernetes team is working on adding new metrics for all Kubernetes workloads, including `CronJob`. This PR:
- Adds the stable KSM CronJob metrics to the metric definition file.
- Updates the e2e integration k8s spec.

### Related PRs
- Update testing resources to include a `CronJob` workload: #608 #622
- Update static test data to include KSM `CronJob` metrics: #623
- Update `infra-integration-specs` [PR](https://source.datanerd.us/infrastructure/infra-integration-specs/pull/302) 
- Update `infra-metrics-utils` [PR](https://source.datanerd.us/infrastructure/infra-metrics-utils/pull/79) 
- Update `dirac` [PR](https://source.datanerd.us/dirac/dirac/pull/7844)

### Description of `CronJob` kube-state-metrics

The [KSM CronJob metric docs](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/cronjob-metrics.md) do not list descriptions for each metric, but these can be pulled from the KSM `/metrics` endpoint.

Descriptions of newly added metrics:
- `kube_cronjob_labels` Kubernetes labels converted to Prometheus labels.
- `kube_cronjob_info` Info about cronjob.
- `kube_cronjob_created` Unix creation timestamp
- `kube_cronjob_status_active` Active holds pointers to currently running jobs.
- `kube_cronjob_status_last_schedule_time` LastScheduleTime keeps information of when was the last time the job was successfully scheduled.
- `kube_cronjob_status_last_successful_time` LastSuccessfulTime keeps information of when was the last time the job was completed successfully.
- `kube_cronjob_spec_suspend` Suspend flag tells the controller to suspend subsequent executions.
- `kube_cronjob_spec_starting_deadline_seconds` Deadline in seconds for starting the job if it misses scheduled time for any reason.
- `kube_cronjob_next_schedule_time` Next time the cronjob should be scheduled. The time after lastScheduleTime, or after the cron job's creation time if it's never been scheduled. Use this to determine if the job is delayed.
- `kube_cronjob_metadata_resource_version` Resource version representing a specific version of the cronjob.
- `kube_cronjob_spec_successful_job_history_limit` Successful job history limit tells the controller how many completed jobs should be preserved.
- `kube_cronjob_spec_failed_job_history_limit` Failed job history limit tells the controller how many failed jobs should be preserved.

### Testing Plan

[...]

PR #610, opened 2023-01-26
Add pv and pvc metric definitions

svetlanabrennan:
Not ready for review/merge yet

### DIFF
--- a/src/metric/definition.go
+++ b/src/metric/definition.go
@@ -565,6 +565,47 @@ var EtcdQueries = []prometheus.Query{
 
 // KSMSpecs are the metric specifications we want to collect from KSM.
 var KSMSpecs = definition.SpecGroups{
+	"persistentvolume": {
+		IDGenerator: prometheus.FromLabelValueEntityIDGenerator("kube_persistentvolume_created", "persistentvolume"),
+		// TypeGenerator:   prometheus.FromLabelValueEntityTypeGenerator("kube_persistentvolume_created"),
+		NamespaceGetter: prometheus.FromLabelGetNamespace,
+		Specs: []definition.Spec{
+			{Name: "createdAt", ValueFunc: prometheus.FromValue("kube_persistentvolume_created"), Type: sdkMetric.GAUGE},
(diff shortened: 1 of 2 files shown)


htroisi:
The static tests are currently failing for this PR. Before we regenerate the static test data, we should merge #609 #626 and  #629 first, since those PRs both:
- update static test data
- modify the `e2e-resources` which changes the output of the static test data generation

charts/internal/e2e-resources/README.md
```diff
@@ -35,8 +35,8 @@ This chart creates e2e resources for nri-kubernetes.
 | loadBalancerService.enabled | bool | `true` |  |
 | loadBalancerService.fakeIP | string | `""` |  |
 | pending.enabled | bool | `true` |  |
-| persistentVolume.enabled | bool | `false` |  |
```
htroisi, line 38: Could you rebase this branch? I believe the `e2e-rsources` changes are included in main now.

svetlanabrennan: Done

src/metric/definition.go
```diff
@@ -565,6 +565,42 @@ var EtcdQueries = []prometheus.Query{
 
 // KSMSpecs are the metric specifications we want to collect from KSM.
 var KSMSpecs = definition.SpecGroups{
+	"persistentvolume": {
+		// kube_persistentvolume_created is marked as an experimental metric, so we instead use the namespace and name
```
juanjjaramillo, line 569: Why do we scrape an experimental metric that we do not use? This metric is not reported in Kubernetes v1.19 and v1.20. The metric was only added on May 23 2022.

Consider removing the variable since static integration tests are failing for this metric

htroisi: We do use it for the `createdAt` metric. It is experimental, but seems very likely to change. To fix the static tests, let's make the `createdAt` metric below optional.

juanjjaramillo: Ok

nserrino (approved):

htroisi (approved):

PR #608, opened 2023-01-24
Add cronjob to testing resources.

htroisi:
The Kubernetes team is working on adding new metrics for all Kubernetes workloads, including `cronjob`. As a first step in that process, this PR adds a `cronjob` workload that will be used in the local dev environment. 

### Testing Plan

1. Run local dev environment: `make local-env-start`
2. Observe `cronjob` running every 1 minute

PR #618, opened 2023-01-31
Fix local dev environment.

htroisi:
This PR fixes the local dev environment (`make local-env-start`) , which was broken in #608. In that PR, I added the `apiVersion` templating to the new `CronJob` resource in order to fix an E2E build error, but I forgot to go back and test the local dev environment. 

**Current broken behavior**

- When running the local dev environment, the [e2e-resource workloads](https://github.com/newrelic/nri-kubernetes/tree/main/charts/internal/e2e-resources/templates) are not deployed.  
- The local dev environment uses Tilt which uses `helm template` to deploy the e2e-resource workloads: https://github.com/newrelic/nri-kubernetes/blob/main/Tiltfile#L54
- Helm template doesn't have access to `.Capabilities.APIVersions` (helm/helm#3377) and I didn't have an else statement to provide a default `apiVersion` for the new `CronJob` workload
- The missing `CronJob` `apiVersion` causes a yaml validation error which prevents the deployment of the e2e-resources. 

**Limitations with the current solution**

- ⚠️ When running the local dev environment, the Tilt `helm template` command will always evaluate to `apiVersion: batch/v1` for the `CronJob` resource. This API has been available since Kubernetes `v1.21`

- ⚠️ To deploy the `CronJob` workload in the local dev environment with a K8s cluster version <  1.20, manually delete `.Capabilities.APIVersions` templating for the `CronJob` resource and use `apiVersion: batch/v1beta1`

**Testing plan**
- Run local dev environment (for K8s v1.25.3) and check that e2e-resources including `CronJob` deploy
- Run local dev environment (for K8s v1.20.0) and check that all e2e-resources except `CronJob` deploy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ make local-env-start
 
 Notice that local images are build and pushed to docker running inside the minikube cluster since we are running `eval $(minikube docker-env)` before launching Tilt.
 
-Note: when running the local-dev-environment with a Kubernetes cluster `< v1.21`, you will need to remove the `apiVersion` templating for the [CronJob resource](https://github.com/newrelic/nri-kubernetes/blob/main/charts/internal/e2e-resources/templates/cronjob.yml#L2) and manually set `apiVersion: batch/v1beta1`.
+Note: when running the local-dev-environment with a Kubernetes cluster `< v1.21`, you will need to remove the `apiVersion` templating for the [CronJob resource](https://github.com/newrelic/nri-kubernetes/blob/main/charts/internal/e2e-resources/templates/cronjob.yml#L2) and manually set `apiVersion: batch/v1beta1`. This is because Tilt uses `helm template` and helm template doesn't render capabilities: https://github.com/helm/helm/issues/3377.
 
 ## Running OpenShift locally using CodeReady Containers
 
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ make local-env-start
 
 Notice that local images are build and pushed to docker running inside the minikube cluster since we are running `eval $(minikube docker-env)` before launching Tilt.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -84,22 +86,26 @@ make test
...
 
 Start the local environment:
+
 ```shell
 make local-env-start
 ```
 
 Notice that local images are build and pushed to docker running inside the minikube cluster since we are running `eval $(minikube docker-env)` before launching Tilt.
 
+Note: for clusters with Kubernetes versions < `1.21`, manually set `apiVersion: batch/v1beta1` for the [e2e-resources/templates/cronjob.yml](https://github.com/newrelic/nri-kubernetes/blob/main/charts/internal/e2e-resources/templates/cronjob.yml).
```
htroisi, line 104: This is the only real change in this file. All of the other modifications were automated markdown linting.

charts/internal/e2e-resources/templates/cronjob.yml
```diff
@@ -1,8 +1,8 @@
 {{- if .Values.cronjob.enabled }}
-{{- if .Capabilities.APIVersions.Has "batch/v1/CronJob" }}
-apiVersion: batch/v1
-{{- else if .Capabilities.APIVersions.Has "batch/v1beta1/CronJob" }}
+{{- if .Capabilities.APIVersions.Has "batch/v1beta1/CronJob" }}
 apiVersion: batch/v1beta1
+{{- else }}
+apiVersion: batch/v1
```
htroisi, line 5: `batch/v1` available since K8s v1.21.

README.md
philkuz: Why do you need this Note if you have a template with an if statement?

philkuz: this Note should be very specific to local-dev-environment and mention this open issue isthe reason it happens. something like tilt uses `helm template` and `helm template` doesn't render capabilities 
https://github.com/helm/helm/issues/3377

htroisi: The `.Capabilities.APIVersions.Has` if statement is used by E2E testing.

philkuz (approved):

PR #620, opened 2023-02-04
Fix datagen.sh

vuqtran88:
In order to fix KSM and k8s versions incompatibility in e2e tests, the change in this [PR](https://github.com/newrelic/nri-kubernetes/pull/595) disabled KSM auto installation. As a result, CI scripts that ran e2e tests had to detect which k8s version  running and manually install the correct KSM version. Even though this pattern made all e2e tests to pass, but it affect the `datagen.sh` script to set up correct environment. This PR is to fix this issue.

This PR put back KSM auto installation feature, but still allow correct KSM version installation by adding back the KSM dependency in the charts/internal/e2e-resources/Chart.yaml file and making use of `--set kube-state-metrics.image.tag` in helm command.

### DIFF
--- a/charts/internal/e2e-resources/Chart.yaml
+++ b/charts/internal/e2e-resources/Chart.yaml
@@ -3,6 +3,11 @@ version: 1.3.0-devel
 description: This chart creates e2e resources for nri-kubernetes.
 name: e2e-resources
 
+dependencies:
+  - name: kube-state-metrics
+    version: 4.23.0
+    repository: https://prometheus-community.github.io/helm-charts
+
 maintainers:
   - name: alvarocabanas
     url: https://github.com/alvarocabanas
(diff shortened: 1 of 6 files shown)


charts/internal/e2e-resources/Chart.yaml
```diff
@@ -3,6 +3,11 @@ version: 1.3.0-devel
 description: This chart creates e2e resources for nri-kubernetes.
 name: e2e-resources
 
+dependencies:
+  - name: kube-state-metrics
+    version: 4.23.0
```
juanjjaramillo, line 8: So when installing the chart we are using the `--set kube-state-metrics.image.tag` flag to choose the correct dependency version for `kube-state-metrics`

Is that flag enough to override this dependency declaration? (I'm new to Helm, so feel free to ignore comment if it does not help)

juanjjaramillo (approved):

charts/internal/e2e-resources/Chart.yaml
vuqtran88: Hi Juan, good question. Just want to clarify, we are still using the same KSM Helm chart version (“4.23”), but we set to use different KSM image versions (aka application version). To my knowledge, the KSM image is the actually KSM application whereas KSM helm chart contains dependency info so that the KSM image can work without any problems.

PR #625, opened 2023-02-09
chore(deps): update dependencies

htroisi:
Dependabot PRs don't work with E2E testing. Recreating #606 so that we can run E2E tests.


alvarocabanas (approved):
LGTM

PR #633, opened 2023-02-17
Rerun Kubelet Scraper When Data is Populated

xqi-nr:
This PR addresses the issue when Kubelet becomes temporarily unavailable, no Kubelet metrics are populated, and the error "kubelet data was not populated after trying all endpoints" from Kubelet scraper is returned to main.  
The current behavior of K8s agent main is that if any (ksm, kubelet, controlplane) scrapers returns any error, the K8s agent will Immediately fail and exit with code 5.

Instead of exiting immediately when kubelet is not reachable, we can set a ScraperMaxReruns (e.g., 4 times) at the scraper level. This will improve customer experience when kubelet is temporarily not reachable.

### DIFF
--- a/charts/newrelic-infrastructure/Chart.yaml
+++ b/charts/newrelic-infrastructure/Chart.yaml
@@ -8,7 +8,7 @@ sources:
   - https://github.com/newrelic/nri-kubernetes/tree/main/charts/newrelic-infrastructure
   - https://github.com/newrelic/infrastructure-agent/
 
-version: 3.12.0
+version: 3.12.1
 appVersion: 3.6.0
 
 dependencies:
--- a/charts/newrelic-infrastructure/Chart.yaml
+++ b/charts/newrelic-infrastructure/Chart.yaml
@@ -8,7 +8,7 @@ sources:
   - https://github.com/newrelic/nri-kubernetes/tree/main/charts/newrelic-infrastructure
   - https://github.com/newrelic/infrastructure-agent/
 
-version: 3.12.1
+version: 3.12.0
 appVersion: 3.6.0
 
 dependencies:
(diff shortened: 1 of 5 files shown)


cmd/nri-kubernetes/main.go
```diff
@@ -192,7 +193,12 @@ func runScrapers(c *config.Config, ksmScraper *ksm.Scraper, kubeletScraper *kube
 	if c.Kubelet.Enabled {
 		err := kubeletScraper.Run(i)
 		if err != nil {
-			return fmt.Errorf("retrieving kubelet data: %w", err)
+			if kubeletScraper.IsMaxRerunReached(kubeletReruns) {
+				return fmt.Errorf("retrieving kubelet data: %w", err)
+			} else {
```
nserrino, line 198: looks like we can omit the else statement since the if above has a return? can help readability a bit. overall looks good!

xqi-nr: thanks, remove the else

nserrino (approved):

charts/newrelic-infrastructure/Chart.yaml
```diff
@@ -8,7 +8,7 @@ sources:
   - https://github.com/newrelic/nri-kubernetes/tree/main/charts/newrelic-infrastructure
   - https://github.com/newrelic/infrastructure-agent/
 
-version: 3.12.0
+version: 3.12.1
 appVersion: 3.6.0
```
htroisi, line 12: If we make changes to the agent, I believe we want to bump both the `appVersion` and the chart `version`. 

That said, I am new to helm charts and the `nri-kubernetes` release process, so I'm less clear on how this works in practice. 😅  Do we want to make these chart changes in this PR or wait until we do a release next Wednesday?

xqi-nr: @nserrino I made several updates to address the lints, including the version change, after you approval. :) So pls review the PR again. I will revert the version change if it is not proper.

xqi-nr: I reverted the chart change before the PR merge

PR #634, opened 2023-02-17
Add failed job to testing resources.

htroisi:
The Kubernetes team is working on adding new metrics for all Kubernetes workloads, including `Job`. This PR adds a failing `Job` workload that will be used in the local dev environment.

### Testing Plan
- Run local dev environment: make local-env-start
- Observe failed jobs

### DIFF
--- a/charts/internal/e2e-resources/templates/job-fails.yaml
+++ b/charts/internal/e2e-resources/templates/job-fails.yaml
@@ -1,0 +1,21 @@
+{{- if .Values.failingJob.enabled }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ .Release.Name }}-failjob
+spec:
+  activeDeadlineSeconds: 200
+  completions: 4
+  parallelism: 2
+  template:
+    spec:
+      restartPolicy: Never
+      containers:
+      - name: main
+        image: docker.io/library/bash:5
+        command: ["bash"]
+        args:
+        - -c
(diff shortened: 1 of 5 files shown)


charts/internal/e2e-resources/templates/job-fails.yaml
```diff
@@ -0,0 +1,21 @@
...
+    spec:
+      restartPolicy: Never
+      containers:
+      - name: main
+        image: docker.io/library/bash:5
+        command: ["bash"]
+        args:
+        - -c
+        - echo "Hello world! I'm going to exit with 42 to simulate a software bug." && sleep 5 && exit 42
+  backoffLimit: 6
```
juanjjaramillo, line 20: `backoffLimit` is capping retries to [6 minutes](https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-backoff-failure-policy), but `activeDeadlineSeconds` [takes precedence](activeDeadlineSeconds) so job will fail after 200 seconds with `reason: DeadlineExceeded`

Consider removing the unused `backoffLimit`

juanjjaramillo (approved):

charts/internal/e2e-resources/templates/job-fails.yaml
htroisi: I think `backoffLimit` might be # of tries, not # of minutes. So I think we're good.

PR #675, opened 2023-03-16
Fix matchExpressions value example

htroisi:
This PR recreates #603 since E2E testing doesn't work from forks.


vihangm (approved):

vihangm:
Let's update docs separately.

Issue #85, opened 2021-04-19
[Fargate] Monitor control plane in EKS/Fargate 


paologallinaharbur:
Investigate what can be scraped regarding the control plane clarifying if the current implementation of nri-kubernetes is capable to do so.
 - [x] In case we cannot scrape it modify nri-kubernetes to do so. Keep in mind we want to avoid scraping duplicate data

Then once possible:
 - [x] Modify manifest/chart configuration to help retrieving such data in Fargate
 - [x] Check if is there any overlap in the hybrid solution.

paologallinaharbur:
I created a "POC", it is indeed possible to scrape the control plane.
![Screen Shot 2021-04-19 at 6 01 11 PM](https://user-images.githubusercontent.com/43335750/115267230-3b001100-a139-11eb-8902-047dec338006.png)

The issue is that the code is not ready to perform the scraping from a slave node.
Moreover we currently do not have any leader election in place, therefore the only way to do it now would be to ask the user to add a flag to a SINGLE instance forcing `nri-kubernetes` to ignore the master label and scrape the control plane

In EKS there is currently no solution that do not involves weird configuration/labels.

paologallinaharbur:
In an hybrid scenario currently it would be not an issue since in EKS the control plane metrics are not scraped

paologallinaharbur:
Closing for now since we do not have a way to scrape such data and it is not requested for feature parity with EKS

PR #725, opened 2023-04-21
Adding new attribute `ReplicaFailure` to deployment workload

juanjjaramillo:
# Description
The Kubernetes team is updating metrics for all supported workloads and services. Specifically, [kube-state-metrics](https://github.com/kubernetes/kube-state-metrics) has upgraded several metrics and services from `experimental` to `stable`.

We noted in the Kubernetes repository that there are [only three valid conditions of a deployment](https://github.com/kubernetes/kubernetes/blob/3615291/pkg/apis/apps/types.go#L460-L473):
- `Available`
- `Progressing`
- `ReplicaFailure`

While they are [briefly mentioned in the documentation](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/), there is no statement indicating they are the only valid conditions.

# PR Changes
This PR makes changes to the agent to start scraping the attribute `ReplicaFailure` from [KSM's metric `kube_deployment_status_condition`](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/deployment-metrics.md#deployment-metrics) to make sure we scrape all valid conditions for Deployment workloads.

# Related PRs
[nri-kubernetes #725](https://github.com/newrelic/nri-kubernetes/pull/725)
[infra-integration-specs #313](https://source.datanerd.us/infrastructure/infra-integration-specs/pull/313)
[infra-metrics-utils #85](https://source.datanerd.us/infrastructure/infra-metrics-utils/pull/85)
[dirac #8087](https://source.datanerd.us/dirac/dirac/pull/8087)
[dirac_async_query_service #3347](https://source.datanerd.us/events-pipeline/dirac_async_query_service/pull/3347)
[window-aggregation-merger #811](https://source.datanerd.us/events-pipeline/window-aggregation-merger/pull/811)
[partial-aggregation-windower #529](https://source.datanerd.us/events-pipeline/partial-aggregation-windower/pull/529)

# Testing Plan
- [x] Run local testing environment and check metrics are ingested to NR1 in the respective `K8s*Sample` tables

### DIFF
--- a/e2e/k8s.yml
+++ b/e2e/k8s.yml
@@ -4288,18 +4288,25 @@ entities:
             - namespace
           legacyEventTypes:
             - K8sDeploymentSample
-      - name: k8s.deployment.isAvailable
+      - name: k8s.deployment.conditionAvailable
         type: string
         migrationInformation:
           legacyNames:
-            - isAvailable
+            - conditionAvailable
           legacyEventTypes:
             - K8sDeploymentSample
-      - name: k8s.deployment.isProgressing
+      - name: k8s.deployment.conditionProgressing
         type: string
         migrationInformation:
           legacyNames:
-            - isProgressing
+            - conditionProgressing
+          legacyEventTypes:
+            - K8sDeploymentSample
+      - name: k8s.deployment.conditionReplicaFailure
+        type: string
+        migrationInformation:
+          legacyNames:
+            - conditionReplicaFailure
           legacyEventTypes:
             - K8sDeploymentSample
       - name: entity.guid
--- a/src/ksm/ksm_test.go
+++ b/src/ksm/ksm_test.go
@@ -49,6 +49,11 @@ func TestScraper(t *testing.T) {
 				exclude.Groups("job_name"),
 				exclude.Optional(),
 			),
+			// Kubernetes deployment's `condition` attribute operate in a true-or-NULL basis, so it won't be present if false
+			exclude.Exclude(
+				exclude.Groups("deployment"),
+				exclude.Optional(),
+			),
 		).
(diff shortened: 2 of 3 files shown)


philkuz (approved):

e2e/k8s.yml
```diff
@@ -4288,18 +4288,25 @@ entities:
             - namespace
           legacyEventTypes:
             - K8sDeploymentSample
-      - name: k8s.deployment.isAvailable
+      - name: k8s.deployment.conditionAvailable
         type: string
         migrationInformation:
           legacyNames:
-            - isAvailable
+            - conditionAvailable
```
nserrino, line 4295: I might be misunderstanding what legacyNames is, but would we want to keep the old attribute name under there for migration purposes?

Also, will this change break user dashboards that are dependent on the old names isAvailable and isProgressing ?

juanjjaramillo: Good questions:

> I might be misunderstanding what legacyNames is, but would we want to keep the old attribute name under there for migration purposes?
This file is used to convert, in infra backend, Samples to Dimensional Metrics (DMs) at ingest time; so usually we want to keep the old attribute name, but as explained below, this is not the case for this PR.

> Also, will this change break user dashboards that are dependent on the old names isAvailable and isProgressing ?
No, these attributes are part of the same [NRIT ticket](https://issues.newrelic.com/browse/NR-84698) that covers this PR, they have not been released so far. I changed the name to better reflect the fact that all three of them are related.

htroisi: @juanjjaramillo just to confirm what you are saying, `k8s.deployment.conditionAvailable` and `k8s.deployment.isProgressing` are metrics you had previously added, but not released? So renaming them isn't an issue?

PR #799, opened 2023-06-15
chore(deps): Directly Use Prometheus Parser - remove prom2json dep 

isaacadeleke-nr:
Using prom2Json results in the implicit loss of all metric families when any error is encountered while reading the prometheus exposition format. 

This change is a sort of POC to show that by directly using the parser provided by the prometheus library, we the callers can choose what to do with some of the seen metrics before an error was encountered while parsing. 

None of the functionality of the scrapers were changed and they remain semantically the same. 

A test was added to show how `parseResponse` preserves metric families that are seen even when parsing errors are encountered. The behaviour is to only keep the metrics seen up until the time of failure

I have successfully run the e2e tests locally and through the github actions here. 

## Testing 
Added new unit/feature tests and locally, successfully ran e2e tests

### DIFF
--- a/src/prometheus/query.go
+++ b/src/prometheus/query.go
@@ -139,7 +139,6 @@ func parseResponse(resp *http.Response, ch chan<- *model.MetricFamily) error {
 
 	var parser expfmt.TextParser
 	metricFamilies, err := parser.TextToMetricFamilies(resp.Body)
-
 	if err != nil {
 		err = fmt.Errorf("reading text format failed: %w", err)
 	}
--- a/src/prometheus/query.go
+++ b/src/prometheus/query.go
@@ -132,7 +132,7 @@ func valueFromPrometheus(metricType model.MetricType, metric *model.Metric) Valu
  * Try our best to parse a response. Even if an error is encountered
  * midway through parsing we will put into the receiving channel any
  * metric families found along the way. We also return any error that
- * we did come along. Fail-fast, best attempt behaviour.
+ * we did come along. Fail-fast, best attempt behavior.
  */
 func parseResponse(resp *http.Response, ch chan<- *model.MetricFamily) error {
 	defer close(ch)
@@ -141,7 +141,7 @@ func parseResponse(resp *http.Response, ch chan<- *model.MetricFamily) error {
 	metricFamilies, err := parser.TextToMetricFamilies(resp.Body)
 
 	if err != nil {
-		err = fmt.Errorf("reading text format failed: %v", err)
+		err = fmt.Errorf("reading text format failed: %w", err)
 	}
 
 	for _, mf := range metricFamilies {
--- a/src/prometheus/query.go
+++ b/src/prometheus/query.go
@@ -5,7 +5,7 @@ import (
 	"net/http"
 
 	model "github.com/prometheus/client_model/go"
-	"github.com/prometheus/prom2json"
+	"github.com/prometheus/common/expfmt"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


src/prometheus/query.go
```diff
@@ -128,6 +128,28 @@ func valueFromPrometheus(metricType model.MetricType, metric *model.Metric) Valu
...
+ * Try our best to parse a response. Even if an error is encountered
+ * midway through parsing we will put into the receiving channel any
+ * metric families found along the way. We also return any error that
+ * we did come along. Fail-fast, best attempt behavior.
+ */
+func parseResponse(resp *http.Response, ch chan<- *model.MetricFamily) error {
+	defer close(ch)
+
+	var parser expfmt.TextParser
+	metricFamilies, err := parser.TextToMetricFamilies(resp.Body)
```
xqi-nr, line 141: Is there a way to let all valid metric families passed?

isaacadeleke-nr: I would have to explore the workings of the parser more, but I think it would require more thought and manipulation on our part. The parser as is, expects a http response that is a byte slice. The passed error tells us what line the error occurred in (for the response body). On failure, we would then have to figure out the line of error, skip past it and then start parsing again.

I suppose we could scan through the byte slice and find the `error line number` occurrence of the the fault and start the parsing again. My issue with this is that it would have to be performed every single time that the endpoint is sampled and seems a bit "hacky". It might be cleaner if we parse ourselves, but I don't think we should be doing that. 

I lean towards not being responsible for dancing around unsupported formats, but it would be a better experience for end users if all valid metric families passed. What are your thoughts?

xqi-nr: I wonder whether we can define better queries at line 153 to filter out the custom metrics or not.

isaacadeleke-nr: I'm not sure what `values.yml` is here. Would you mind explaining. 

The queries at line 153 are applied after the parsing is done. See lines 172 to 179. So I don't think those queries will help us in this case.

xqi-nr: is it possible to filter out the return metrics based on type through `/metrics` endpoints https://github.com/newrelic/nri-kubernetes/blob/19674a74fd3fa5df4b7489758af4f51c116acc26/src/ksm/scraper.go#L206

isaacadeleke-nr: I think it is possible to filter out return metrics based on type, but I don't think that will help us in this situation. 

Our problem is as follows: 

- The parser we use fails because of an unrecognized `TYPE` for prometheus 
- The parser will not continue parsing the contents of the `/metrics` endpoint once some failure occurs
- Filtering as currently happens occurs after parsing. To filter out certain metric types we must parse them to know what they are 

So fundamentally this feels like a parsing, where a more robust solution might be to try and get our desired change in at the parsing level (which we don't own).

isaacadeleke-nr: I spoke with @xqi-nr offline and I misunderstood his question. He was asking if it was possible to filter at the api/endpoint level. This seems to be an [open issue](https://github.com/kubernetes/kube-state-metrics/issues/1690) that is not yet supported.

xqi-nr (approved):

src/prometheus/query.go
csongnr: Man, it would be nice to be able to filter metrics from the KSM endpoint directly, but in the meantime, i think this is an acceptable alternative to make our parsing less brittle.

csongnr (approved):

PR #802, opened 2023-06-20
fix(parsing): Log an error instead of fully failing on partial parsing failure

isaacadeleke-nr:


csongnr:
I noticed you said you ran e2e tests on this, but this is a unit test. Did you run unit tests for this? i.e. run `make tests` from the root directory?

src/ksm/client/client_test.go
```diff
@@ -60,3 +84,27 @@ func testHTTPServer(t *testing.T, requestsReceived *int, timeout time.Duration)
...
+		func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusOK)
+			_, err := io.WriteString(w,
+				`# HELP kube_pod_status_phase The pods current phase. 
+				 # TYPE kube_pod_status_phase gauge
+				 kube_pod_status_phase{namespace="default",pod="123456789"} 1
+				 # HELP kube_custom_elasticsearch_health_status Elasticsearch CRD health status
+				 # TYPE kube_custom_elasticsearch_health_status stateset
+				 kube_custom_elasticsearch_health_status {customresource_group="elasticsearch.k8s.elastic.co"} 1
+				`)
```
csongnr, line 101: can you add a comment above this function to clarify what the expected outcome here is?
and comment the logged error in the PR description to clarify that the parsing passed for gauge and failed for stateset?

isaacadeleke-nr:
> I noticed you said you ran e2e tests on this, but this is a unit test. Did you run unit tests for this? i.e. run `make tests` from the root directory?

I did run unit tests in the root directory as well. I meant that both unit tests and e2e tests were validated to still be working (along with a new unit test)

src/ksm/client/client_test.go
isaacadeleke-nr: Will do

csongnr (approved):
lgtm! thanks for the comments

csongnr:
we can just rerun e2e tests later when code freeze is over, should be fixed by then :)

PR #817, opened 2023-07-25
NR-139168: Fix k8s.container.cpuCoresUtilization metric calculation

sachin-shankar:
### Background:

The 'cpuCoresUtilization' metric is a computed where the `value = cpuUsedCores / cpuLimitCores`. Further the 'cpuUsedCores' is another computed where `metric = 'usageNanoCores' / 10e9`

The 'usageNanoCores' is arbitrarily observed to hold a super high value driving up the 'cpuCoresUtilization' percentage value. Issue seems to be reproducible a few times per day on various EKS clusters.

### Analysis: 

The 'usageNanoCores' metric is scraped by the nri-agent's kubelet's pod both at the pod level and container level from the kubelet's `/metrics/summary` endpoint. I reviewed the `kubelet` open-source repo for more details and found this.

The kubelet currently gathers container metrics from either of the 2 sources - cAdvisor or Container-Runtime Interface (CRI). The CRI based metrics have to be enabled explicitly via the `PodAndContainerStatsFromCRI` kubernetes feature-gate. It is unlikely that the customer has enabled this.

The CPU UsageNanoCores metric is a computed metric and is cumulative. For cAdvisor please refer to the following code references to see how it is calculated.

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/stats/helper.go#L52

https://github.com/kubernetes/kubernetes/blob/master/vendor/github.com/google/cadvisor/info/v2/conversion.go#L276

https://github.com/kubernetes/kubernetes/blob/master/vendor/github.com/google/cadvisor/info/v2/conversion.go#L252

To summarize, the cAdvisor first checks if the currentValue is > lastValue and if it is true, it divides the difference with the time delta between current timestamp and last timestamp. Its hard to say precisely what could have caused this high value from the above calculation.

I tried searching for issues reported on the 'kubelet' and `cAdvisor` repos and found the following issues - https://github.com/kubernetes/kubernetes/issues/114057. 
https://github.com/google/cadvisor/issues/3199

Unfortunately the kubelet issue is closed as there were no owners assigned. I see a comment towards the end reporting that the issue was observed on v1.25 as well, so I don't suppose the cluster version upgrade will fix the above issue.
The cAdvisor issue is still open at the time of this writing.

### Suggested fix:

I believe as a fix on nri-kubernetes integration, we may treat the scraped 'UsageNanoCores' metric value as invalid if the value is arbitrarily large. This fix would also serve as an additional check to ensure the value we ingest into NR is within the reasonable bounds.

I ran the below NRQL queries to observe the correlation between `cpuUsedCores (aka usageNanoCores)`, `cpuLimitCores` and `cpuCoresUtilization` where:
`cpuCoresUtilization = (cpuUsedCores/cpuLimitCores) * 100`

256613879-d8cf9a89-6cab-4f5f-9ccc-b4e711e13f47
256613884-91d6dc3d-b332-4c0c-b257-83140e372b77
256613887-e40c964a-35bf-4aa4-af3d-88f0e6f1a7ae
256613888-e8a433f5-0469-4db3-8cd8-68cb7485f731

From the results, the observation is that the absurd cpuUsedCores numbers begin when :
`cpuCoresUtilization = (cpuUsedCores/cpuLimitCores) * 100`

So a reasonable assumption could be something like this:
`cpuUsedCores >  cpuLimitCores * 100`


svetlanabrennan (approved):
LGTM. Just need to update the screenshots and remove slack link.

And add this link to the issue with cadvisor since it'll useful to have connected to this pr: https://github.com/google/cadvisor/issues/3199

sachin-shankar:
> LGTM. Just need to update the screenshots and remove slack link.
> 
> And add this link to the issue with cadvisor since it'll useful to have connected to this pr: [google/cadvisor#3199](https://github.com/google/cadvisor/issues/3199)

src/metric/definition.go
```diff
@@ -1595,3 +1597,67 @@ func metricSetTypeGuesserWithCustomGroup(group string) definition.GuessFunc {
...
+		return nil, errEntityCheck
+	}
+
+	value, ok := entity["cpuLimitCores"]
+	if !ok {
+		// there is likely no CPU limit set for the container which means we have to assume a reasonable value
+		// since there is no way to know the max cpu cores for the current node, use default max of 96 cores supported by most cloud providers
+		// a higher value wouldn't hurt our calculation as the cpuUsedCores value will be a super high number
+		log.StandardLogger().Warnf("cpuLimitCores metric not available. using default max 96 cores")
+		value = 96000 // 9600m k8s cpu unit
```
juanjjaramillo, line 1641: Can you fix the typo in the comment? It is confusing to see both a constant of 96000 and the comment having a different value

juanjjaramillo (changes requested):

juanjjaramillo (approved):

PR #836, opened 2023-08-07
chore: Update CHANGELOG format

svetlanabrennan:
In order to use the release toolkit, we need to follow a certain changelog format: https://github.com/newrelic/release-toolkit/blob/main/generate-yaml/README.md

This PR updates the changelog format.

PR #837, opened 2023-08-07
chore: Add CHANGELOG github workflow

svetlanabrennan:
Adding a changelog action that valids if a PR updates a changelog by using the `newrelic/release-toolkit/validate-markdown@v1`.

Relates to this PR: https://github.com/newrelic/nri-kubernetes/pull/836

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,6 +1,5 @@
 # This action requires that any PR targeting the main branch should touch at
-# least one CHANGELOG file. If a CHANGELOG entry is not required, add the "Skip
-# Changelog" label to disable this action.
+# least one CHANGELOG file.
 
 name: changelog
 
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -18,3 +18,4 @@ jobs:
 
       - name: Check if CHANGELOG is valid
         uses: newrelic/release-toolkit/validate-markdown@v1
+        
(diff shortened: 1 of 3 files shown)


.github/workflows/changelog.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  lint-changelog:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Check if CHANGELOG is valid
+        uses: newrelic/release-toolkit/validate-markdown@v1
+
```
juanjjaramillo, line 20: Have you tested this works as intended? Reading the [documentation](https://newrelic.atlassian.net/wiki/spaces/TB/blog/2023/04/28/2856552590/Release+Automation+using+Release-Toolkit+and+Renovate#PR-validation) it looks like it only checks what has been introduced in the Unreleased section follows the right pattern. What if it is empty?

juanjjaramillo (approved):

.github/workflows/changelog.yml
svetlanabrennan: I've tested this on one of my test personal repos. If the Unreleased section is empty then that's fine - it'll pass. It's just the L3 headers can't be empty.

juanjjaramillo: Ok, so it is still honor system

Issue #499, opened 2022-06-22
[chart] resources are applied to both containers


paologallinaharbur:
Currently resources are applied to both containers and therefore the consumption could be x2 then expected.

This is not a bug since it was decided to leave it like this to make easier the configuration, but it should be better documented

cdobbyn:
Resources should be configurable here, but shouldn't be set by default.

dbudziwojskiNR:
Completed as part of [v4.0.0](https://github.com/newrelic/nri-kubernetes/releases/tag/v4.0.0).

PR #852, opened 2023-08-16
feat: update resource assignment

cdobbyn:
## Which problem is this PR solving?
There are two problems this solves:
1. There shouldn't be default resource assignments, the helm standard is to not assign but to recommend defaults as comments in the values file. The way it's configured today makes it so that by default this component uses WAY too many resources by default.
2. Individual containers should have their resources configured separately. Again the way this works today means this component uses too much resources.

Fixes: https://github.com/newrelic/nri-kubernetes/issues/853, https://github.com/newrelic/nri-kubernetes/issues/499

## Short description of the changes
- default resource assignment to not have resource assignment
- make it so individual containers have separate resource assignment

## Type of change

## New Tests?
n/a

## Checklist:

- [ ] Tests have been added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ### enhancement
-- Add changelog workflow by @svetlanabrennan in [#837](https://github.com/newrelic/nri-kubernetes/pull/837)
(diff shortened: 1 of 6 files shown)


juanjjaramillo:
Closing this PR after having an internal discussion on the issues that can be generated if we do not set resource limits to our integration

cdobbyn:
@juanjjaramillo https://github.com/newrelic/nri-kubernetes/pull/918

PR #876, opened 2023-09-12
Updated KSM unit tests

svetlanabrennan:
## Which problem is this PR solving?

Updated KSM tests after updating KSM dependency

## Short description of the changes
- Removed `Optional: true` from metric definition file so we can just use `exclude.Metrics(<metricName>)` in the test files instead. 
- Replaced `namespace` and `deploymentName` attribute on KSM "deployment" to be retrieved from `kube_deployment_created` instead of `kube_deployment_labels` since `kube_deployment_labels` is not enabled by default. 
- Removed some unnecessarily excluded metrics from unit tests

## Type of change

Please delete options that are not relevant.

## New Tests?
Please describe the new tests that were added (if applicable).

- [x] This change requires changes in testing:
  - [x] unit tests
  - [ ] e2e tests

## Checklist:

- [x] Tests have been added


src/ksm/ksm_test.go
```diff
@@ -42,34 +42,35 @@ func TestScraper(t *testing.T) {
 			// false.
 			exclude.Exclude(
 				exclude.Groups("horizontalpodautoscaler"),
-				exclude.Metrics("isActive", "isAble", "isLimited"),
+				exclude.Metrics("isLimited"),
 			),
 			// Kubernetes jobs either succeed or fail (but not both). Thus, the KSM metrics related to success (isComplete, completedAt)
 			// and failure (failed, failedPods, failedPodsReason) are marked as optional in src/metric/definition.go
```
juanjjaramillo, line 48: Can you change this comment from:
```
 are marked as optional in src/metric/definition.go
```
to:
```
are excluded.
```

juanjjaramillo (approved):

src/ksm/ksm_test.go
svetlanabrennan: Done

PR #916, opened 2023-10-23
feature: Adding support for GKE Autopilot

jiggyjigsj:
## Description
This pull request enhances the New Relic Helm chart by introducing a feature flag for Google Kubernetes Engine Autopilot, making it more compatible and functional in Autopilot environments. The key changes in this PR include:

## Type of change
* ControlPlane: Disabled all resource generation for control plane.
* Host Network: Disabled hostNetwork for all charts
* Kublet Configmap: Added defauly schema and port config as part of a single flag. 

- [ ] Security fix

## Checklist:

  - [ ] unit tests
  - [ ] E2E tests
  

Note: This is my first chart change, so I'm eager to get this pull request out and see if we can get a build created. Your feedback and guidance are greatly appreciated.


jiggyjigsj:
The new common-library helper is located on: https://github.com/newrelic/helm-charts/pull/1178

jiggyjigsj:
I've been attempting to write unit test, but it doesn't like `nil` values.  Keep running into, i can use a bit of help to fix that.  

```
executing "newrelic.common.gkeAutopilot" at <.Values.global.gkeAutopilot>: nil pointer evaluating interface {}.gkeAutopilot
```

charts/newrelic-infrastructure/templates/controlplane/_enabled_helper.tpl
```diff
@@ -0,0 +1,62 @@
+{{/*
```
paologallinaharbur, line 1: I am not sure if the best approach is to make the chart gkeAutopilot aware. 
HAve you considered managing the changes by config/docs?

Example. In the guided install/docs asking if we are deploying for GKE autopitol and in that case add:
 - controlPlane.enabled=false

paologallinaharbur: Is it intended to be a new hidden parameter? There is no explanation of its behavior in the values.yaml or docs

charts/newrelic-infrastructure/templates/controlplane/_enabled_helper.tpl
```diff
@@ -0,0 +1,62 @@
...
+Create ControlPlane Agent ConfigMap
+*/}}
+{{- define "nriKubernetes.controlPlane.agentConfigMap" -}}
```
paologallinaharbur, line 4: (Assuming that the answer to the first conversation is "yes")

I believe that this can be simplified into having just one helper function. Es:
`isControlPlaneNeeded = {{- if and (.Values.controlPlane.enabled) (not (include "newrelic.common.gkeAutopilot" .)) (not (include "newrelic.fargate" .)) -}}`

And then add a second if in the cases that are needed (RBACs/Service account)
```
if isControlPlaneNeeded{
   if {rbac.create} 
}
```

charts/newrelic-infrastructure/templates/controlplane/_enabled_helper.tpl
paologallinaharbur: An other possibility is to add a values.yaml in the values and provide just a safe check:

Example 
```
if autopilot==true && ( priviledge==true || controlPlane.enable == true){
fail " with autopilot you should disable blablabla"
}
```

charts/newrelic-infrastructure/templates/kubelet/_host_network.tpl
```diff
@@ -1,7 +1,9 @@
 {{/* Returns whether the kubelet scraper should run with hostNetwork: true based on the user configuration. */}}
 {{- define "nriKubernetes.kubelet.hostNetwork" -}}
 {{- /* `get` will return "" (empty string) if value is not found, and the value otherwise, so we can type-assert with kindIs */ -}}
-{{- if get .Values.kubelet "hostNetwork" | kindIs "bool" -}}
+{{- if (include "newrelic.common.gkeAutopilot" .) -}}
```
paologallinaharbur, line 4: (Assuming this is the [approach we want to follow](https://github.com/newrelic/nri-kubernetes/pull/916/files#r1382950552))

I think that privilege pods are not allowed with Autopilot, therefore  you should also take into account that

charts/newrelic-infrastructure/templates/controlplane/_enabled_helper.tpl
jiggyjigsj: @paologallinaharbur I was planning for a streamlined configuration to simplify New Relic integration for Autopilot (and other providers). However, it's not overly complex to configure individual settings either. Our plan is to automate manifest generation across all our clusters, so the method of configuration isn't a major concern for us. I'm open to whether we document the configuration in values or introduce a switch akin to gkeAutopilot.

Here is the library for newrelic.common.gkeAutopilot is configured in another [PR](https://github.com/newrelic/helm-charts/pull/1178/files#diff-f9d62a22d26fd59016e07cda0a450faa953e9574a5c867d77cb7fb0c8305a152). I can add some documentation to reference this if I get a green light on using the library.  

I introduced the [_enabled_helper.tpl](https://github.com/newrelic/nri-kubernetes/pull/916/files/f967a2789e9de19bf9e0620e3a4b6a734298f594#diff-06ccb5096b2deda174168410b2138c3225bbea1c7432195092e7f1e8dc6111a1) because it was generating a controlplane manifest even when the deployment was disabled, again this is not really an issue, but figured it could be properly disabled.

Let me know if you'd like to meet and discuss a path forward.  I am hoping this changes make it easier for consumer to deploy New Telic to Autopilot.

charts/newrelic-infrastructure/templates/controlplane/_enabled_helper.tpl
jiggyjigsj: That makes more sense.  I can update the helper if we land on this approach.

charts/newrelic-infrastructure/templates/kubelet/_host_network.tpl
jiggyjigsj: Correct, privilege pods are not allowed.  If we move away from this approach, we would just need to document override privilege config for any deployments/daemonset that are enabled.

PR #1066, opened 2024-06-12
fix: pin mountPropagation of host fs in privileged mode (NR-274598)

gsanchezgavier:
## Description

Infrastructure Agent captures Storage Samples of underlying nodes based on the mounted file system. Whenever a new pvc is mounted on the host, this mount needs to be propagated (`rslave`)to the mounted filesystem in the agent container. 

By default k8s uses `none` as mount propagation, and according to [k8s docs](https://kubernetes.io/docs/concepts/storage/volumes/#mount-propagation) depends on the CRI which propagation is used `rprivate` or in some cases like docker `rslave`. 

This PR configures the host file system volume to be mounted with `rslave` independently of the CRI. This fixes some miss behaving storage metrics samples where the device disk space of a newly mounted device has the value of the primary host storage ( and the issue was fixed upon Pod recreation)

## Type of change

- [ ] Security fix

## Checklist:

- [ ] Add changelog entry following the [contributing guide](../CONTRIBUTING.md#pull-requests)
  - [ ] unit tests
  - [ ] E2E tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### bugfix
 - Fix expired certificated @dbudziwojskiNR [#1064](https://github.com/newrelic/nri-kubernetes/pull/1064)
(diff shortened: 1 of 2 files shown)


juanjjaramillo (approved):

juanjjaramillo:
PR has been tested in #1068
All tests pass

PR #1175, opened 2025-03-06
Add Github Action to build infra-agent & nri-kubernetes Windows 2019 & 2022 images

kondracek-nr:
## Type of change

- [ ] Security fix

## Checklist:

- [ ] Add changelog entry following the [contributing guide](../CONTRIBUTING.md#pull-requests)
  - [ ] unit tests
  - [ ] E2E tests


.github/workflows/windows-infra-agent.yml
```diff
@@ -0,0 +1,105 @@
+name: Windows build
+
+on:
+  push:
+    branches:
+      - windowsImage*
```
dbudziwojskiNR, line 6: Why are we only building on these branches vs all branches? Also, why are we building the infra agent? Is the on host team not providing a windows image?

charts/internal/e2e-resources/templates/windows-deployment.yaml
```diff
@@ -10,37 +10,37 @@ spec:
...
+            - powershell
+            - " -Command"
+            - |
+              while ($true) {
+                Write-Host "Random Text: $(Get-Random)";
+                Start-Sleep -Seconds 3600;
+              }
 {{- end }}
 ---
 {{- if .Values.windows.is2022 }}
```
dbudziwojskiNR, line 46: Could we have both deployments exist instead of more config values and just switch to using this selector: [node.kubernetes.io/windows-build](https://kubernetes.io/docs/concepts/windows/user-guide/#handling-multiple-windows-versions-in-the-same-cluster)

.github/workflows/windows-infra-agent.yml
kondracek-nr: No, they are not. The on-host team will need more runway to do the work to build the image themselves via infrastructure-bundle so we're building it ourselves for public preview.

We aren't building on all branches yet because we need an agreement on where the images go (right now they're going to nri-kubernetes with tag management by me). Will need to sync with Mikel & on-host about that.

charts/internal/e2e-resources/templates/windows-deployment.yaml
kondracek-nr: We can make this change but for this PR the change to this e2e-resources doc was just fixing some yaml spacing.

dbudziwojskiNR: Yeah, sorry - I just saw the if statement and was like... hmmmm we can probably do better :) Could you create a follow up for that?

.github/workflows/windows-infra-agent.yml
dbudziwojskiNR: Gotcha. Yeah, it just seemed odd that we were building another team's product. As long as there's agreement for public preview and a transition plan for GA (we really shouldn't own this in the future) I'm good.

dbudziwojskiNR (changes requested):

dbudziwojskiNR (approved):

PR #1185, opened 2025-03-24
Update e2e-resources and e2e for local Windows testing

TmNguyen12:
## Description
This PR adds several resources to the e2e-resources chart that deploy onto Windows-based nodes (Windows Server 2019 & Windows Server 2022) when enabled.

Resource types include
- deployment / replicaset
- statefulset
- cronjob
- daemonset
- failing jobs

It also includes e2e tests meant to run _locally_, separately from the standard e2e tests. Our current automated e2e tests run using Minikube, which does not support Windows nodes, so we follow the basic local test script pointing at a different exceptions file & specs file for Windows. Automating this will be a little involved, and will come later.

Note that several tests present for Linux are not present for Windows due to known limitations in available metrics (eg most `netTx`/`netRx`/`netErrors`), tests that are not relevant for Windows (anything to do with `cpuCfs`), & some that I haven't looked into yet (PVC, container pending). This should cover the bulk of the tests, though.

Known issues:
- Until the PR updating the nri-kubernetes helm chart is merged, the main integration must be installed separately from the [windowsUpdateHelmChart](https://github.com/newrelic/nri-kubernetes/tree/windowsUpdateHelmChart) branch.
- I don't like the way I've laid out the 2019 vs 2022 tests, but it should do for now.
- There are only exceptions files for 1.29 & 1.32, as we do not currently have canaries for other k8s versions
- Resource requests/limits are fairly high for the new Windows resources. Powershell won't run on the resource requests/limits we had in place for linux-based resources (too low), but we could probably tune these requests down further.

## Type of change

- [ ] Security fix

## Checklist:

  - [ ] unit tests
  - [x] E2E tests


charts/internal/e2e-resources/README.md
```diff
@@ -32,11 +33,15 @@ This chart creates e2e resources for nri-kubernetes.
...
 | loadBalancerService.enabled | bool | `true` | Deploy a loadBalancer service |
 | loadBalancerService.fakeIP | string | `""` | If set, will deploy service with a loadBalancerIP set to this value |
+| openShift.enabled | bool | `false` |  |
 | pending.enabled | bool | `true` | Enable crashing and pending pods |
 | persistentVolume.enabled | bool | `true` | Create PVs |
+| persistentVolume.multiNode | bool | `true` | Changes PV type to run on multi-node clusters (e.g. GKE, OpenShift on GCP) |
 | persistentVolumeClaim.enabled | bool | `true` | Create PVCs |
 | scraper.enabled | bool | `false` | Deploy the scraper pod |
 | statefulSet.enabled | bool | `true` | Deploy a dummy statefulSet |
+| windows.is2019 | bool | `false` | Deploy resources on Windows Server 2019 nodes |
```
dbudziwojskiNR, line 43: Sorry to keep nit picking on this, but can we not use tolerations here? Really trying to avoid baking in new config options that we'll end up needing to support in the future. If the plan is to drop theses keys in a follow up PR because we're currently testing Windows, I'm okay with this. Just checking.

dbudziwojskiNR (changes requested):

charts/internal/e2e-resources/README.md
kondracek-nr: i'd like to keep the option in e2e-resources because we might not always want to deploy resources for windows. is there an expectation that we need to support e2e-resources for customers?

dbudziwojskiNR: I get that, but wouldn't the windows resources not deploy anyways due to tolerations?

TmNguyen12: because they are deployments they deploy and are stuck in pending because they cant' find a node. 

if they were daemonsets they would not deploy

kondracek-nr: only if the cluster the user is deploying e2e-resources to doesn't have windows nodes! our canaries, for example, have a windows node by default, so it's possible someone's just trying to test something real quick in linux nodes. i'd rather have more levers to pull to target testing, like we do with whether or not we want deployments/statefulsets/daemonsets in general.

dbudziwojskiNR (dismissed):

dbudziwojskiNR (approved):

PR #1335, opened 2025-11-12
Improve network metric gathering

kondracek-nr:
## Description
For OKE (and EKS, and OpenShift), the `/stats/summary` endpoint does not return top-level network metrics such as `rxBytes` and `txBytes`, with the result that `k8s.netRxBytesPerSecond` & related metrics are missing. This has resulted in customer complaints (for EKS - OKE support is not finalized for K8s integration). 

/stats/summary example for OKE node. note no name, and no rxBytes etc in top level.
```
"network": {
   "time": "2025-12-04T00:09:17Z",
   "name": "", 
   "interfaces": [
    {
     "name": "ens3",
     "rxBytes": 6987536335,
     "rxErrors": 0,
     "txBytes": 1075747019,
     "txErrors": 0
    },
    {
     "name": "ens5",
     "rxBytes": 1275926546,
     "rxErrors": 0,
     "txBytes": 3842418196,
     "txErrors": 0
    },
    {
     "name": "ocid264c098fd6",
     "rxBytes": 108606201,
     "rxErrors": 0,
     "txBytes": 97286036,
     "txErrors": 0
    },
```

/stats/summary example for AKS node, which has full top level network stats
```
   "network": {
    "time": "2025-12-04T00:10:54Z",
    "name": "eth0",
    "rxBytes": 788,
    "rxErrors": 0,
    "txBytes": 488,
    "txErrors": 0,
    "interfaces": [
     {
      "name": "eth0",
      "rxBytes": 788,
      "rxErrors": 0,
      "txBytes": 488,
      "txErrors": 0
     }
    ]
   },
```

This PR does the following:

- mounts `/host/proc` to the kubelet container in privileged mode
- changes the default route file path that is used when top-level network metrics are missing (`/proc/net` is a symlink to `self/net`, so even when `/host/proc` is mounted we can't get the default interface for the host from the file. using `/host/proc/1/net/route` allows us to use host PID 1 - generally init - which is not a symlink)
- introduces a heuristic that helps the integration identify the default network interface when top-level network metrics are not available AND it is not able to access `/host/proc`, either due to lack of permissions or running in `unprivileged` mode. when the heuristic is invoked, the integration uses known physical interface patterns to filter out unlikely candidates (CNI, veth) & identify the likely default interface. note that for pods, we skip the check for `/host/proc/1/net/route` since that's pointing to the host's network routing table rather than the pod. we could use the pod's `/host/proc/net/route` file, but here we're just falling back to the heuristic.
- adds a cache for network interfaces so that when we do have to determine the default interface (via proc file or heuristic fallback), we reduce potential performance issues. Cache is vacuumed every 10 iterations since interface names are pretty stable. 

Unrelated -
- removes the `log.StandardLogger().Debug("cpuLimitCores metric not available. using default max 96 cores")` log, which only logs out while checking for "impossibly high cpuUsedCoresVal" calculation, wasn't very helpful in the first place, confusing to customers at worst, & annoying to me personally.

Basic workflow:
- Check /stats/summary return for top level network metrics. If present, stop here.
- For nodes: check /host/proc/1/net/route for default network interface. If present, stop here
- Use heuristic to identify most likely default interface in child netw

Even with the cache, there is potential for performance issues in particular scenarios (mainly with EKS and OKE). We may want to introduce some or all of the following:
- a flag to disable more nuanced network metrics gathering
- allow customers to configure how often the cache vacuums/clears (or whether it clears)
- prepopulate the cache on startup

Image demonstrating a switch between the base/main nri-kubernetes image and the network metrics improvement image:
image

---
### Performance Analysis between and OKE cluster and an AKS cluster
For your edification...

OKE does not provide top-level network metrics for nodes and for hostNetwork pods. AKS provides top-level network metrics for all of these. This is a comparison of the base ("Base") deployment of nri-k8s and the improved network ("Network") metrics version of nri-k8s in an OKE cluster and an AKS cluster. Logs were taken for each pod over a 20 minute period of time. Analysis performed by Claude.

## Overview

This analysis compares the performance impact of the new network metrics implementation (with caching and heuristics) against the baseline version across AKS and OKE Kubernetes clusters.

**Overall Performance Impact: Negligible (-0.02%)**

Key findings:
- **Performance**: -0.04ms average delta (essentially zero impact)
- **AKS**: Cache not used (kubelet provides direct metrics - optimal path)
- **OKE**: Cache actively used (~1,200+ hits per run - efficient fallback)
- **Architecture**: Code correctly optimizes for different kubelet implementations
- **Stability**: No new errors

The implementation demonstrates zero overhead when metrics are available directly, efficient caching when fallback is needed.

### Overall Performance Impact

**Across all tests (352 base samples, 355 network samples):**
- Base mean: 173.98ms
- Network mean: 173.94ms
- **Delta: -0.04ms (-0.02%)**
- **Conclusion: Performance impact is negligible (within 5%)**

#### AKS Cluster

**Privileged Mode (90 base, 99 network samples):**
- Base: 223.40ms mean, 210ms median
- Network: 216.12ms mean, 212ms median
- **Delta: -7.28ms (-3.26%)** ✓ Slight improvement

**Unprivileged Mode (95 base, 85 network samples):**
- Base: 191.44ms mean, 195ms median
- Network: 200.59ms mean, 198ms median
- **Delta: +9.15ms (+4.78%)** ⚠️ Slight regression

#### OKE Cluster

**Privileged Mode (82 base, 87 network samples):**
- Base: 134.17ms mean, 133.5ms median
- Network: 136.49ms mean, 138ms median
- **Delta: +2.32ms (+1.73%)** ≈ Negligible

**Unprivileged Mode (85 base, 84 network samples):**
- Base: 140.55ms mean, 138ms median
- Network: 136.05ms mean, 134ms median
- **Delta: -4.51ms (-3.21%)** ✓ Slight improvement

## Type of change

- [ ] Security fix

## Checklist:

  - [x] unit tests
  - [ ] E2E tests

### DIFF
--- a/src/kubelet/metric/network.go
+++ b/src/kubelet/metric/network.go
@@ -36,90 +36,128 @@ var (
 // 1. Try top-level metric (e.g., n.Network.RxBytes).
 // 2. Try default interface from routing table (only for node entities, from /host/proc/1/net/route).
 // 3. Use heuristic to select primary interface (lowest-numbered physical interface).
-//
-//nolint:gocyclo,cyclop,gocognit
 func FromRawWithFallbackToDefaultInterface(metricKey string, cache *InterfaceCache) definition.FetchFunc {
 	return func(groupLabel, entityID string, groups definition.RawGroups) (definition.FetchedValue, error) {
-		g, ok := groups[groupLabel]
-		if !ok {
-			return nil, ErrGroupNotFound
-		}
-
-		e, ok := g[entityID]
-		if !ok {
-			return nil, ErrEntityNotFound
+		e, err := getEntityMetrics(groupLabel, entityID, groups)
+		if err != nil {
+			return nil, err
 		}
 
 		// Step 1: Try top-level metric (e.g., n.Network.RxBytes directly)
-		value, ok := e[metricKey]
-		if ok {
+		if value, ok := e[metricKey]; ok {
 			return value, nil
 		}
 
-		// Step 1.5: Check cache for previously resolved interface
-		if cache != nil {
-			if cachedInterface, found := cache.Get(entityID); found {
-				log.Debugf("Using cached interface '%s' for entity '%s' metric %s", cachedInterface, entityID, metricKey)
-				metric, err := getMetricFromInterface(cachedInterface, metricKey, e)
-				if err == nil {
-					return metric, nil
-				}
-				// If cached interface no longer exists in stats, fall through to resolve again
-				log.Debugf("Cached interface '%s' not found in current stats for entity '%s', re-resolving", cachedInterface, entityID)
-			}
+		// Step 2: Try cached interface
+		if metric, found := tryCache(cache, entityID, metricKey, e); found {
+			return metric, nil
 		}
 
-		// Step 2: Try default interface from routing table (only for nodes)
-		// For pods/containers, skip routing table check as they have their own network namespaces
-		// and the host's routing table doesn't apply to them.
-		var resolvedInterface string
-		if groupLabel == "node" {
-			defaultInterface, err := getDefaultInterface(groups)
-			if err == nil && defaultInterface != "" {
-				metric, err := getMetricFromInterface(defaultInterface, metricKey, e)
-				if err == nil {
-					resolvedInterface = defaultInterface
-					// Cache the resolved interface
-					if cache != nil {
-						cache.Put(entityID, resolvedInterface)
-					}
-					return metric, nil
-				}
-				// If we got a default interface name but couldn't find metrics for it,
-				// fall through to heuristic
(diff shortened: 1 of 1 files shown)


src/metric/definition.go
```diff
@@ -1760,7 +1754,7 @@ func filterCPUUsedCores(fetchedValue definition.FetchedValue, groupLabel, entity
 		return nil, errCPULimitTypeCheck
 	}
 
-	// check for impossibly high cpuUsedCoresVal
+	// check for impossibly high cpuUsedCoresVal - workaround for https://github.com/kubernetes/kubernetes/issues/114057 (resolved)
```
TmNguyen12: nice

PR #1433, opened 2026-03-09
fix: use container status resources over desired spec resources for cpu/memory resource metrics during in-place pod vertical scaling

kondracek-nr:
Kubernetes 1.33 introduced in-place pod vertical scaling (beta, on by default; GA in 1.35). With this feature, Pod.Spec.Containers[i].Resources becomes the desired state rather than the actual state. The actual applied resources live in Pod.Status.ContainerStatuses[i].Resources, which is only populated once the kubelet has successfully enacted the allocation.

This means that during an active resize, cpuRequestedCores, memoryRequestedBytes, cpuLimitCores, and memoryLimitBytes — and the utilization ratios derived from them — were reporting the target values before they were applied to the running container.

## Description

**Change**
In fetchContainersData, prefer Pod.Status.ContainerStatuses[i].Resources (actual applied state) over Pod.Spec.Containers[i].Resources (desired state), with a fallback to Spec when Status resources are nil. Sidecar init containers (RestartPolicy: Always) are handled the same way via Pod.Status.InitContainerStatuses.

**Backward compatibility**

- Pre-1.33 clusters: ContainerStatus.Resources is never populated, so the nil fallback always fires. No behavior change.
- 1.33+ clusters, no resize in progress: Spec and Status are identical. No behavior change.
- 1.33+ clusters, resize in progress: Values differ. We now report the currently-enforced allocation rather than the pending target.

**Why change existing metric semantics vs. adding new metrics**
Upstream kube-state-metrics is likely taking an additive approach (new kube_pod_container_actual_resource_* metrics) because changing existing metric semantics would silently break dashboards across their entire user base with no path to coordinate consumers.

Our situation is different: we control both the metrics and the dashboards that consume them, and cpuRequestedCores semantically means "what is currently being enforced on the container." Using the desired state for utilization calculations (cpuUsageCores / cpuRequestedCores) produces a ratio that doesn't reflect the container's actual resource envelope during a resize. The value change is also narrow in scope — it only occurs during the window between a resize being requested and the kubelet applying it.

Desired-state metrics (cpuRequestedCoresDesired etc.) and resize condition metrics (PodResizePending, PodResizeInProgress) are left as follow-up work, as is parity with the OTel collector chart (blocked on upstream KSM shipping the new actual-resource metrics).

**Related**

- Jira: NR-523430
- KSM tracking issue: kubernetes/kube-state-metrics#2665
- KSM PRs: #2702, #2773

## Type of change

- [ ] Security fix
- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] Add changelog entry following the [contributing guide](https://github.com/newrelic/nri-kubernetes/blob/main/CONTRIBUTING.md#pull-requests)
  - [x] unit tests
  - [ ] E2E tests


dbudziwojskiNR (dismissed):

src/kubelet/metric/pods.go
```diff
@@ -192,12 +194,13 @@ func getKubeServiceHost() string {
 
 func (podsFetcher *PodsFetcher) fetchContainersData(pod *v1.Pod) map[string]definition.RawMetrics {
 	statuses := make(map[string]definition.RawMetrics)
-	fillContainerStatuses(pod, statuses)
+	containerStatusByName := getContainerStatusesByName(pod)
+	fillContainerStatuses(pod, containerStatusByName, statuses)
```
jamescripter, line 198: should `fillContainerStatuses` take care of calling `getContainerStatusesByName` under the hood? seems a little redundant to pass both pod state and container statuses to it. not sure though

jamescripter (dismissed):

danielstokes (dismissed):

src/kubelet/metric/pods.go
dbudziwojskiNR: It was like that before, but now the calling method also needs the container statuses. Rather than calling `getContainerStatusesByName` twice, I just called it once and pass it down to avoid iterating twice.

dbudziwojskiNR (approved):

Philip-R-Beckwith (approved):

PR #1465, opened 2026-05-27
fix: bug with windows images

danielstokes:
## Type of change

- [ ] Security fix

## Checklist:

- [ ] Add changelog entry following the [contributing guide](https://github.com/newrelic/nri-kubernetes/blob/main/CONTRIBUTING.md#pull-requests)
  - [ ] unit tests
  - [ ] E2E tests


charts/newrelic-infrastructure/values.yaml
```diff
@@ -54,9 +54,12 @@ images:
   windowsAgent:
     registry: ""
     repository: newrelic/infrastructure-bundle
-    tag: 3.3.28
+    tag: v3.3.27
     pullPolicy: IfNotPresent
 
+# -- Enable Linux node monitoring.
+# @default -- `true`
+enableLinux: true
```
TmNguyen12: do we need this? it's only linked to the daemonset, and it may not be clear to the user that the other linux scrapers are going to deploy regardless.

danielstokes: yes.  i'm not sure how or why, but i guess we have at least 1 customer that had this issue

dbudziwojskiNR: Since K8s requires a least one linux node, we shouldn't support this.

charts/newrelic-infrastructure/tests/kubelet_image_test.yaml
```diff
@@ -216,3 +222,45 @@ tests:
...
+          value: "global-registry/newrelic/nri-kubernetes:1.0.0-ltsc2022"
+        template: templates/kubelet/daemonset-windows.yaml
+        documentIndex: 1
+
+  - it: global registry is used for Windows agent image with default repository
+    set:
+      licenseKey: test
+      cluster: test
+      enableWindows: true
+      images.windowsAgent.tag: v2.0.0
```
TmNguyen12, line 251: can you add a test case overwriting `images.windowsIntegration.registry` and confirm it overwrite the global? I think this was here initially because we were doing a lot of testing with our own images

PR #1493, opened 2026-06-26
support kubernetes 1.35

jamescripter:
## Description
- support kubernetes 1.35
- remove support for 1.30

## Type of change

- [ ] Security fix

## Checklist:

- [x] Add changelog entry following the [contributing guide](https://github.com/newrelic/nri-kubernetes/blob/main/CONTRIBUTING.md#pull-requests)
- [x] This change requires changes in testing:
  - [x] unit tests
  - [x] E2E tests


.github/workflows/build-windows-integration-test.yml
```diff
@@ -33,34 +29,9 @@ permissions:
   security-events: write
 
 jobs:
-  generate-matrix:
```
TmNguyen12, line 36: Can we keep this with just one version? That way when we are asked to support version 2025 we don't have to make as many changes?

TmNguyen12: Actually, we haven't officially dropped ltsc2019 yet so we still need to keep it in here for support.

.github/workflows/test-windows-e2e.yml
```diff
@@ -133,7 +131,6 @@ jobs:
           helm repo update
       - name: Add PullSecret to e2e-values.yml
         run: |
-          export IMAGE_2019="newrelic/nri-kubernetes-internal:windows-ltsc2019-e2e-testing-${{ needs.generate-key.outputs.run_key }}"
```
dbudziwojskiNR, line 136: Do we have product sign off on dropping 2019?

.github/workflows/test-windows-e2e.yml
```diff
@@ -147,7 +144,7 @@ jobs:
           FETCH_PODS_FROM_KUBE_SERVICE: "false"
         with:
           retry_seconds: 60
-          retry_attempts: 5
+          retry_attempts: 7
```
dbudziwojskiNR: We really shouldn't bump this unless absolutely necessary

src/ksm/ksm_test.go
```diff
@@ -138,8 +138,7 @@ func TestScraper(t *testing.T) {
 }
 
 func TestScraper_FilterNamespace(t *testing.T) {
-	// We test with a specific version to not modify number of entities
-	version := testutil.Version(testutil.Testdata134)
+	version := testutil.LatestVersion()
```
dbudziwojskiNR, line 141: Are we sure we want to do this? Seems like there was a reason it specified an exact version based on the comment?

TmNguyen12: I think this is fine because we are putting in the latest version on each k8s version upgrade anyway.

.github/workflows/test-windows-e2e.yml
jamescripter: added it back for k8s <1.35

danielstokes (approved):

TmNguyen12 (approved):
looks good. we should probably update [this](https://docs.newrelic.com/docs/kubernetes-pixie/kubernetes-integration/get-started/kubernetes-integration-compatibility-requirements/#req-windows) with specifics now on which version of windows is supported on which version of k8s (since we're not supporting ltsc2019 from v1.35

